Issue #14, opened 2025-06-17
microphone occasionally can get into a bad state


cjpais:
Not exactly sure what's happening with this one yet. Probably need to issue some debug builds to properly fix this but it seems like occasionally the microphone will just not be read at all and Result in empty transcriptions and since the VAD is permissive right now it will result in the thank you transcriptions

cjpais:
I believe this was a bug in the old audio backend and should be closed. Now we can reopen it if it comes back.

PR #29, opened 2025-06-30
activate the microphone on demand

### LINKED ISSUE #5: Microphone Always Active Prevents Sleep
While Handy is running the microphone is on all the time even when not actively transcribing audio.

The initial purpose of doing this was to minimize any latency from bringing the microphone online and active within the application. From some rough measurements I took, it can take about 100ms on MacOS to start the microphone. This is a latency penalty I am slightly concerned about, especially for users thinking they can just immediately start speaking after pressing a keyboard shortcut. 

The tradeoff of this is that on MacOS you see the microphone always active anytime that Handy is running. This is certainly not ideal. This also disables the computers ability to be able to sleep. This is also really not good. The overall energy impact on the system is quite small, but regardless we need the comptuer to sleep.

Currently I am a bit split what to do, for me, I don't care that the microphone is active all the time since I can read the source code and I can clearly see that nothing malicious is happening. However for an average user this behavior might be concering. The part which prevents sleep however, must be addressed. So either it means:

* not having the microphone on all the time
* having the microphone on, but detecting lid closure or similar to stop the microphone

Image
Image

cjpais:
This should resolve #5 and I suspect #14 as well.

We only initialize the microphone after the user has clicked the buttons to begin transcribing. The old low latency path still survives and can be enabled by editing the settings file. Specifically setting `"always_on_microphone": true`

On MacOS this is located: `/Users/<user>/Library/Application Support/com.pais.handy/settings_store.json`

### DIFF
--- a/src-tauri/src/actions.rs
+++ b/src-tauri/src/actions.rs
@@ -1,13 +1,16 @@
 use crate::managers::audio::AudioRecordingManager;
 use crate::managers::transcription::TranscriptionManager;
+use crate::settings::get_settings;
 use crate::utils;
 use crate::utils::change_tray_icon;
 use crate::utils::play_recording_start_sound;
 use crate::utils::play_recording_stop_sound;
 use crate::utils::TrayIconState;
+use log::debug;
 use once_cell::sync::Lazy;
 use std::collections::HashMap;
 use std::sync::Arc;
+use std::time::Instant;
 use tauri::AppHandle;
 use tauri::Manager;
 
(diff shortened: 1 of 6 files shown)


jacksongoode:
@cjpais Could you explain the reasoning for this change? I'm wondering if there is anyway to trigger the playback immediately without delaying the microphone? On macOS the sound trigger is super late and looking for ways to make it feel more responsive.

cjpais:
@jacksongoode what do you mean?

largely if we keep the microphone active all the time this can prevent App Nap on macOS and also keeps the microphone icon in the tray always. I think this is quite concerning to the average user.

there is a microphone initialization delay and this is generally the reason. let me know if you think there is some other way

jacksongoode:
Ah I see, mostly my curiosity was around the macOS behavior. The audio delay is quite long from when the hotkey is triggered.

PR #32, opened 2025-07-04
Add Translate to English Setting

segudev:
## Summary
- Added a new setting to enable automatic translation of speech to English
- Updated settings UI components to include the translation toggle  
- Integrated translation logic into the transcription process
- Adjusted accessibility permission calls for consistency

## Test plan
- [x] Verify the translation setting appears in the settings UI
- [x] Test that enabling translation works with multilingual Whisper models
- [x] Confirm the setting persists across app restarts
- [x] Validate that translation only works with compatible models (small, medium, large)

### DIFF
--- a/src-tauri/src/lib.rs
+++ b/src-tauri/src/lib.rs
@@ -174,6 +174,7 @@ pub fn run() {
             shortcut::reset_binding,
             shortcut::change_ptt_setting,
             shortcut::change_audio_feedback_setting,
+            shortcut::change_translate_to_english_setting,
             trigger_update_check,
             commands::models::get_available_models,
             commands::models::get_model_info,
--- a/src-tauri/src/managers/transcription.rs
+++ b/src-tauri/src/managers/transcription.rs
@@ -173,6 +173,9 @@ impl TranscriptionManager {
             )
         })?;
 
+        // Get current settings to check translation preference
+        let settings = get_settings(&self.app_handle);
+
         // Initialize parameters
         let mut params = FullParams::new(SamplingStrategy::default());
         params.set_language(Some("auto"));
@@ -182,6 +185,11 @@ impl TranscriptionManager {
         params.set_print_timestamps(false);
         params.set_suppress_blank(true);
         params.set_suppress_non_speech_tokens(true);
+        
+        // Enable translation to English if requested
+        if settings.translate_to_english {
+            params.set_translate(true);
+        }
 
         state
             .full(params, &audio)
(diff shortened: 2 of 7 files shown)


cjpais:
This PR breaks on MacOS. Some of the changes for the accessibility permissions make the UI not show up.

Also it seems to remove the ability to change the keyboard shortcut in the UI..

segudev:
Hi @cjpais,

Thanks for the feedback! I'm running macOS Sequoia 15.5 and the PR is working correctly on my machine - the UI displays properly and I can change keyboard shortcuts without any issues.

CleanShot 2025-07-07 at 12 49 31@2x

I had to remove and add again Handy from the accessibility permission list, but that's about it. Happy to help debug and fix these issues - just need a bit more guidance on what exactly needs to be addressed regarding the accessibility permissions.
Thanks!

cjpais:
The problem was just the typescript throwing errors with the accessibility permissions here, since these functions don't exist.

```
import {
   checkAccessibilityPermission,
   requestAccessibilityPermission,
 } from "tauri-plugin-macos-permissions-api";
 ```

I bet you fixed this if the UI is back. If you could, can you update the PR with the latest set of changes, I am happy to review. If you could also make the default window size of the application bigger so users can immediately see all the settings that would be great. If you could also change the order of the settings that would be great. I think the keyboard shortcut should be at the top, then push to talk, audio feedback, and finally translate.

A slight note (mostly to myself) is that I would really like to compress the UI a bit if possible. This is not something that is blocking this PR, but something I am actively thinking about in terms of usability, and may delay the merge of the PR in while I have a think

cjpais:
@segudev I am going to be pulling in #34 soon which is quite significant overall, and will likely have merge conflicts with this. Though, the suggestions I made will be much much easier to implement now. I will try to help you port the changes over after that code is in. I just want to thank you for your contribution so far and would be very happy to have you be the first outside contributor to the project!

segudev:
Ok no problem I will rebase this next week I guess.

Happy to help, I use it daily now, so glad I could replace superwhisper
with something fast and open source :)

Btw, maybe your release will fix it but I noticed sometimes the transcribed
text gets pasted 2x, like a double trigger. Have you seen this too?

cjpais:
So glad to hear! 

And I have not seen that. What platform are you on? Would you mind opening an issue? I only use MacOS regularly so the other platforms are not as well tested. I try to test before each release but it's just a few things

cjpais:
Closing in favor of #36. I went ahead and rebased, thanks for the new feature!

PR #41, opened 2025-07-23
Fix: Persist settings on change

dimikat:
This commit fixes an issue where the hotkey and other user settings were not being saved correctly, causing them to revert to their default values after a restart.

The root cause of the problem was that the `useSettings` hook in the frontend was not saving the settings to the Tauri store after an update. The `autoSave` option for the store was set to `false`, so an explicit save was required.

The following changes have been made to address this issue:

- A `saveSettings` function has been added to the `useSettings` hook to persist the settings to the Tauri store.
- The `updateSetting` and `updateBinding` functions in the `useSettings` hook now call the `saveSettings` function after each successful update.
- The `loadSettings` function has been simplified to be the single source of truth for loading settings.

### DIFF
--- a/src/hooks/useSettings.ts
+++ b/src/hooks/useSettings.ts
@@ -45,6 +45,19 @@ export const useSettings = (): UseSettingsReturn => {
     outputDevices: [],
   });
 
+  // Save settings to the store
+  const saveSettings = useCallback(async (settingsToSave: Settings) => {
+    try {
+      const { load } = await import("@tauri-apps/plugin-store");
+      const store = await load("settings_store.json", { autoSave: false });
+      await store.set("settings", settingsToSave);
+      await store.save();
+      console.log("Settings saved successfully.");
+    } catch (error) {
+      console.error("Failed to save settings:", error);
+    }
+  }, []);
+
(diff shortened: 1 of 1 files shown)


dimikat:
@coderabbitai review

cjpais:
@dimikat was there a certain platform you saw this on? I did try to replicate it on my Mac and I don't have the issue on the latest build

dimikat:
Windows!

cjpais:
@dimikat So far I've tried replicating the issue on Mac, Linux, and Windows, and I can't replicate it. 

I am:
1. Opening the app
2. Changing the binding from ctrl+space to ctrl+0
3. Closing the app
4. Reopen app and try ctrl+0

I also tried restarting my computer after the change and the settings persisted as expected.

What version of Handy did you run into the issue with? With the 0.3.0 version do you have this issue? Is there some other order of operation of events that you run into the issue with?

I'm slightly hesitant to merge a change that I can't replicate. Let me know the best way I can replicate this or can you send a screen share with the issue or something?

cjpais:
Closing this for now as I cannot replicate.

PR #54, opened 2025-07-28
Add language select to settings

### LINKED ISSUE #52: [feature request] Set language to improve accuracy
First of all, it looks really nice, love the UI.

But i have a problem that when i speak czech, it is mistaking it for other slavic languages. It would be cool to be able to set language/lanugages i am speaking to be more accurate. Thinking if just set like preffered language or have for each language separate push to talk shortcut, because i think i will use it for english and czech.

I might create PR, so what do you think about it?

vaclavskorpil:
Added language selection dropdown to settings panel

Fixes #52

### DIFF
--- a/src-tauri/resources/default_settings.json
+++ b/src-tauri/resources/default_settings.json
@@ -7,5 +7,6 @@
       "default_binding": "Platform-specific: ctrl+space (Windows/Linux), alt+space (macOS)"
     }
   },
-  "push_to_talk": true
+  "push_to_talk": true,
+  "selected_language": "auto"
 }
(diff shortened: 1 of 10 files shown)


cjpais:
Thank you so much for this PR it looks great to me. 

I made a minor tweak to move the language selector to the main settings panel instead of advanced, changed "Auto-detect" to just "Auto" for now, re-arranged the languages to be in the approximate commonality of them instead of alphabetical (though may change this back if it's a bad default), and tweaked the default height of the handy window so all settings show up by default.

PR #62, opened 2025-08-02
style: window height, logo width, group title size

vladstudio:
When I opened the settings window in the latest version, the height was just a little bit not enough and I saw the vertical scroll bar. So I tweaked some little things to fix it.


cjpais:
thanks :) I actually meant to fix this before release just forgot

PR #67, opened 2025-08-02
style(overlay): update recording overlay theme to pink palette

vladstudio:
The new overlay is great! 
Pink text on dark grey (Transcribing...) did not look good for me, so I took this chance to tweak some colors, and made the overlay more recognizably Handy.

image

image


cjpais:
I do agree the colors and things weren't amazing for the overlay, I am a bit hesitant to have pink as the background color even though it is definitely very Handy. 

I have been primarily using pink as the accent color, curious if there's any other redesign that you feel would be a bit nicer. I'm not really a designer so I kind of just threw something at the wall as a starting point and pretty open to other options. 

I was also at one point considering having the like hand logo be part of this but I don't know if it really makes sense just to show that it's more obviously handy but yeah I'm not really sure. I can also provide the SVG if needed for just the hand itself.

vladstudio:
"[dont use overlay height for calculation of y, it moves it too far up](https://github.com/cjpais/Handy/pull/67/commits/cc2c35f86d21c71546601e6dc4e7ef786702269c)"

hmm, on my mac (external display) the overlay is stuck to the bottom:

image

vladstudio:
Just push the commit. And I like it much better! And the overlay is now on top, near the "island". WDYT?


image

cjpais:
From the screenshot I like this version a lot. What do you think about a dark/light mode variant? I think theoretically this can all be done in CSS just so it follows the system theme better. Not sure how the light mode would work in terms of contrast over other apps and whatnot, the dark one seems to contrast generally well as far as I've tested.

For moving to the top, I think maybe we can make the position an 'advanced setting' keeping the bottom position as default for now. This doesn't need to be included in this PR necessarily, but if you have time that would be great too! 

I could see a future version where the overlay is always active as well (but minimizes/collapses) into the handy hand icon and animates/expands out when the voice starts

vladstudio:
New idea! Replace show overlay toggle with dropdown: "Do not show", "On bottom", "On top". See latest commit.

cjpais:
Yeah, this is great, much cleaner, thank you! I'm probably going to make a few more tweaks and just leave this open in case we want to make any more changes in the meantime. I'll probably pull it in later this week

cjpais:
Merged! Thanks @vladstudio!

Issue #65, opened 2025-08-02
Tray icon has a low contrast on Windows light theme


patrickiel:
<img width="26" height="30" alt="Image" src="https://github.com/user-attachments/assets/61ef1c50-0d16-430a-80d2-887f7e2ca2f5" />

I would prefer a tray icon that switches dynamically with the system theme. Thank you.

PR #69, opened 2025-08-02
Respect OS Theme with Tray Icon

cjpais:
should help to fix: #65


cjpais:
Tested on Windows 11 and it should work

PR #75, opened 2025-08-08
Custom words

vladstudio:
Here's a new feature: Custom words, because my company name is "Kadoa" and of course it never gets it right!

  Changes

  - Backend: Added phonetic (Soundex) and string similarity matching to automatically correct misspelled words during transcription
  - Frontend: Created CorrectWords component with streamlined single-line layout and clickable word badges
  - Settings: Integrated custom words into app settings with persistent storage


cjpais:
I tested this and it works great as far as I can tell based on my very minimal testing. The only main suggestion I have is to clean up the UI a bit. Right now the tooltip doesn't work and the text entry box is quite large. It could probably be smaller as well as for the deleting words, maybe you can take a look at the UI and make it more in theme if you feel it's appropriate. I'm okay with the red as is, but maybe there's a more consistent way for the UI. I'm not exactly sure. You probably have a good idea.

vladstudio:
Made some changes, please take a look

cjpais:
This is really great, thank you so much Vlad. I am gonna make a few minor tweaks later today and then merge it + release!

cjpais:
@vladstudio just a heads up I am familiar with the dev/localhost:1420 issue as well but I think we should handle it in another PR.

mainly, we probably just want to disable autostart from the dev environment itself, instead of not showing the window by default

PR #87, opened 2025-08-19
fix clipboard/paste issues on linux/i3wm

yapus:
When using `i3wm` X windows manager - the recording indicator window catches focus when the recording is started, causing the input to be pasted nowhere afterwards.

Also for Terminal window the standard paste shortcut is not Ctrl+V, so now there's a switch to use "typing" instead of pasting

### DIFF
--- a/src-tauri/src/actions.rs
+++ b/src-tauri/src/actions.rs
@@ -30,6 +30,15 @@ impl ShortcutAction for TranscribeAction {
         let start_time = Instant::now();
         debug!("TranscribeAction::start called for binding: {}", binding_id);
 
+        // Capture focused window ID BEFORE any UI changes
+        let focused_window_id = crate::utils::get_focused_window_id();
+
+        // Store it in managed state for later use during paste
+        let focused_window_state = app.state::<crate::ManagedFocusedWindow>();
+        if let Ok(mut state) = focused_window_state.try_lock() {
+            *state = focused_window_id;
+        }
+
         let binding_id = binding_id.to_string();
         change_tray_icon(app, TrayIconState::Recording);
         show_recording_overlay(app);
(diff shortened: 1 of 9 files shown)


cjpais:
Thanks for this I'll take a look! We may end up using different defaults on different systems depending on the performance. I'm happy to have type as the default on Linux once I've validated

cjpais:
Okay, so this needs a little bit of work still

I think we should be using the `enigo.text` function instead of writing our own: `send_text_directly`. Under the hood it tries to use a fast input method if possible, and if not it falls back to iterating through and pressing the keys.

The UI for it should also either be a toggle or a drop down. I would prefer a toggle. This way we don't need any new components. 

The default input method should be: 'paste' for Windows and MacOS and we can make it 'type' for Linux.

We should use an enum for the input_method, it is already one on the typescript side and we should unify this for the rust backend as well. 

I am okay with calling xdotool. But it might be worth writing a note about this somewhere as well as having it only happen on Linux. For the longer delay that was added, we should cut that back and feature flag that for different platforms as well.

nv2p:
Just wanted to mention that I have noticed weird things happening when using post-processing (ex. commands started in Emacs) which may be related to the way input is happening on Linux, at least when fast input can't be used and the "letter after letter" method is used. It may be worth considering this (more generally) when applying a fix for i3wm and other Xorg or Linux environnements.

yapus:
Since I posted this - I totally switched from `i3wm` (and X11) to `hyprland` and so I'm using `hyprwhspr`. Please do not expect me to work on this PR. Feel free to close it

cjpais:
Thanks @yapus, it inspired some other PR's which will likely be pulled in. Appreciate the contributions regardless

PR #89, opened 2025-08-19
Upgrade whisper-rs to 0.15

cjpais:
This just upgrades to the latest version of whisper-rs. However, this fails to build on macOS for me, and doesn't seem to support older versions of macOS. So this is a tentative upgrade, but it won't be pulled in until these issues are fixed.


cjpais:
This is a `transcribe-rs` problem now, may not be done for a while as it breaks some MacOS support.

PR #94, opened 2025-08-25
Add shortcut to Translate on the fly

alessandro308:
image

This PR, mostly AI generated, and tested on my MacOS, adds a shortcut setting to enable the transcribe+translation with push-to-talk, in the same way it already works when the Translate to English flag is activated, so that I can keep the spoken language as well as translate it, just switching the used shortcut

### DIFF
--- a/src-tauri/src/actions.rs
+++ b/src-tauri/src/actions.rs
@@ -22,6 +22,9 @@ pub trait ShortcutAction: Send + Sync {
 // Transcribe Action
 struct TranscribeAction;
 
+// Transcribe with Translation Action
+struct TranscribeTranslateAction;
+
 impl ShortcutAction for TranscribeAction {
     fn start(&self, app: &AppHandle, binding_id: &str, _shortcut_str: &str) {
         let start_time = Instant::now();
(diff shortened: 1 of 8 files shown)


cjpais:
Thank you so much for contributing this PR. I am so glad that you implemented this. This kind of capability is effectively one of the reasons why I wrote Handy and why I want people to fork it, so I'm super happy to see that you forked it and implemented this even if mostly AI written. That is the biggest win in my book. 

Before integrating this into the main project, I do want to rethink about how shortcuts can be added and be user defined, and how the UI can be cleaned up to support this. So, for the time being, this won't make it into the project, but at some point, I would absolutely love to have this. 

So, I'm going to keep this open, partially as a marker and a note, to figure out the best way to integrate this into the project moving forward. And also, for anyone who wants this feature, that they can go and use your fork. I may add a section on the readme for notable forks that implement certain features. Obviously distribution is a hard problem there, but at the very least people can attempt to get the features they want.

alessandro308:
No problem, that makes totally sense. I added the feature in the fastest way possible, and it works for my needs. I think that if we want to include it, a better UI should be offered at least to allow the shortcuts to scale better in the UI itself.

Let me know if I can help in any way on that.

cjpais:
Closing this for now, if this hasn't been mentioned in a discussion yet, let's move the conversation over there for now

PR #100, opened 2025-09-03
Pink Tray Icon on Linux

cjpais:
This fixes an issue on Linux where you can't see the tray icon because the theme prevents it from being shown.

Now we use a pink icon instead of a white or black one so that you can always see it in the tray regardless.


syabro:
I barely see the difference between idle/rec
Can it be tweaked locally?

PR #105, opened 2025-09-06
Added "Start Hidden"

Anaphylaxis:
[vibe coded but tested on Windows, not macOS or Linux yet)
Added "Start Hidden" feature. Needs more review but thought I'd get it out there so people can test it etc.

### DIFF
--- a/src-tauri/src/lib.rs
+++ b/src-tauri/src/lib.rs
@@ -91,6 +91,14 @@ pub fn run() {
         ))
         .manage(Mutex::new(ShortcutToggleStates::default()))
         .setup(move |app| {
+            // Apply macOS Accessory policy early if starting hidden
+            #[cfg(target_os = "macos")]
+            {
+                let settings = settings::get_settings(&app.handle());
+                if settings.start_hidden {
+                    let _ = app.set_activation_policy(tauri::ActivationPolicy::Accessory);
+                }
+            }
             // Get the current theme to set the appropriate initial icon
             let initial_theme = tray::get_current_theme(&app.handle());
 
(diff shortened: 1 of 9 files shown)


cjpais:
Thanks for this it's been requested a lot! I tested it on MacOS and works great

cjpais:
works on linux too, lgtm

not tested extensively, but the code looks good enough for me and it works so happy to have it! thanks! really appreciate your contribution

Anaphylaxis:
cjpais, after your latest commit to remove `visible: true` the window pops up as a white square before hiding itself. `visible: true` does hide the window by default, but "we" (ai) coded a check that if the option is disabled, it will force show itself. Therefore I believe this commit is a regression as we have broken the intended behavior and made it worse UX.

cjpais:
Maybe this is a windows only issue, it doesn't affect MacOS or Linux which are the two I tested

I'll test with it again and bring it back in. I just didn't like how "aggressive" it seemed to disable it in a config file versus programmatically. To me this sounds like a tauri bug personally and is a quirk of the config.

Anaphylaxis:
> Maybe this is a windows only issue, it doesn't affect MacOS or Linux which are the two I tested
> 
> I'll test with it again and bring it back in. I just didn't like how "aggressive" it seemed to disable it in a config file versus programmatically. To me this sounds like a tauri bug personally and is a quirk of the config.

I somewhat agree but I don't see a workaround other than this unless there's another method in Tauri. It seems like an elegant workaround, and from testing had no issues, but I understand that it might be too aggressive.

Issue #91, opened 2025-08-20
[BUG] Crash on|after model load


shockme:
## Bug Description
```
whisper_init_from_file_with_params_no_state: loading model from
 '/home/user/.local/share/com.pais.handy/models/ggml-small.bin'
whisper_init_with_params_no_state: use gpu    = 1
whisper_init_with_params_no_state: flash attn = 0
whisper_init_with_params_no_state: gpu_device = 0
whisper_init_with_params_no_state: dtw        = 0
[New Thread 0x7fff2f746600 (LWP 1388782)]
ggml_vulkan: Found 2 Vulkan devices:
Vulkan0: AMD Radeon RX 580 Series (RADV POLARIS10) (radv) | uma: 0 | fp16: 0 | warp size: 64
Vulkan1: Tesla P40 (NVIDIA) | uma: 0 | fp16: 0 | warp size: 32

Thread 1 "handy" received signal SIGILL, Illegal instruction.
0x00005555564067cf in ggml_backend_vk_reg_get_device(ggml_backend_reg*, unsigned long) ()
(gdb)
```
Might be because of lack of AVX2 instructions (see cpuinfo below).

## System Information
**App Version:**
0.3.8 AppImage

**Operating System:**
Ubuntu 22.04

**CPU:**
processor	: 31
vendor_id	: GenuineIntel
cpu family	: 6
model		: 45
model name	: Intel(R) Xeon(R) CPU E5-2670 0 @ 2.60GHz
stepping	: 7
microcode	: 0x71a
cpu MHz		: 2000.000
cache size	: 20480 KB
physical id	: 1
siblings	: 16
core id		: 7
cpu cores	: 8
apicid		: 47
initial apicid	: 47
fpu		: yes
fpu_exception	: yes
cpuid level	: 13
wp		: yes
[...]

Issue #77, opened 2025-08-14
[Feature Request] Slow Transcription - Whisper API support?


burntcouscous:
First off; I love the work you put into this app and tryign to make it as user friendly as possible. Thank you!

Unfortunately transcribing is suuuper slow for me (practically unuseable) as I need to wait +1 minute for transcribing even 1 second of audio recording with just even 'Whisper small'. 

I'm on a ~2015 laptop running windows 11 (8GB RAM Intel(R) Core(TM) m3-5Y30 CPU at 0.9GHz). 

Would you consider implementing support for Openai Whisper API powered transcription via asking api key? This should solve that issue.

cjpais:
Hey thank you for the kind words and I'm really sorry it's not running well on your computer. I did recently remove some BLAS support which may make the CPU inference run slower than before. Perhaps I can add it back, but I do need to test it. Let me see if I can find a machine with a lower power CPU and no GPU to see how it does. In addition I'll go through and see if there's any other optimizations on CPU that can be made where possible. 

Right now I am not planning to add support for remote Whisper API as the app is meant for local use only. And due to this there are effectively system requirements. 

I've had another request for API support in the past and I will think about how best to solve the issue of hardware. For now though the application will remain local only. I especially would like to not rely on specific API providers nor ones that make you pay. 

Perhaps there is a way that people who are running Handy can spare some of their idle compute to do transcriptions for others that don't have the hardware to do so. This is obviously a much more complex thing to do and would have to be carefully crafted and opt-in.  I'm not confident this is the way forward but something I'm pondering.

Another thought is maybe a computer donation program or something. Surely there are many people with older computers that are still quite powerful that are sitting around for one reason or another. Perhaps it's possible to get them moving and be an upgrade for someone else.

burntcouscous:
Do you mind explaining why so strict on the local only use?

I get not being eager to support non-open ai-companies; but what if you make it non-obvious or very optional looking to enable it? That way people with good enough requirements are unlikely to even look to enable it. 

It will be interesting to see the BLAS support back; I'm not familiar with it but I fear it isn't going to be nearly enough considering the delay. 

I think it adds massively to the value of this app as you could use tiny or old - cheap machines and use them as mostly voice controlled in combo with privacy secure open source ai chat ui solution like JAN ai interface. More power to move a step away from being caged into company privacy controlled AI UIs.

cjpais:
[...]

Issue #18, opened 2025-06-18
Add support/replace backend with Parakeet models


jacksongoode:
Currently near the top in quality and speed: https://huggingface.co/spaces/hf-audio/open_asr_leaderboard specifically the [parakeet-tdt-0.6b-v2](https://huggingface.co/nvidia/parakeet-tdt-0.6b-v2) model. This might be a big undertaking but the size and performance would make the tool lightweight and near instantaneous. It might be worth looking at [this repo](https://github.com/blakkd/faster-whisper-hotkey).

cjpais:
If someone wants to build support for this, I would willingly accept contributions, however it's not something that I will be specifically working on due to the general lack of support for cross-platform within Rust and the bindings for Parakeet in particular. If there are Rust bindings that come up for Parakeet, I would be interested. I think it is critical that it works cross-platform and cross-GPU architecture as well.

This was mentioned in #7

cjpais:
Closing this for now, if there's good cross-platform support for a parakeet as well as across accelerators, I'd be super happy to add it.

cjpais:
Just noting, there is an MLX implementation of Parakeet, which could serve as a good reference for building a library in Rust as well.

https://github.com/senstella/parakeet-mlx/tree/master

cjpais:
Opening this back up: https://github.com/jason-ni/parakeet-rs.git

Parakeet with ONNX

cjpais:
Also might be curious about Moonshine to support users with lower CPU requirements 

https://github.com/moonshine-software/moonshine

vladstudio:
here's what Claude has to say about the topic:

Executive Summary

  Feasibility: HIGH - Migration is technically feasible but requires significant architectural changes. Parakeet v3 offers substantial
  performance advantages (50-1000x faster inference) but comes with important trade-offs in language support and implementation complexity.

  Current Architecture Analysis

  Whisper Integration (Current State)

  The Handy app currently uses a well-integrated Whisper implementation:

  Core Components:
  - Transcription Manager (src-tauri/src/managers/transcription.rs): Manages WhisperContext and WhisperState with model loading and
  transcription logic
  - Model Manager (src-tauri/src/managers/model.rs): Handles model downloads, caching, and selection for Small/Medium/Turbo/Large Whisper
  variants
  - Audio Pipeline: AudioRecorder → VAD (Silero) → Whisper → Text output at 16kHz sample rate
[...]

Issue #11, opened 2025-06-16
Occasionally, 'thank you' will be added to the end of the transcription.


cjpais:
This is likely due to the VAD filter being too permissive. Some tuning and tweaking to this is definitely necessary and any help on this would be appreciated to get a good value that seems to stop this. And if it could be well replicated and tested, that would be great on audio samples that always have this happen to them.

kiranz:
I also noticed "Thank you for watching." text added to the end of the transcription sometime.
This is probably an issue with Whisper model itself.

cjpais:
This is definitely a whisper issue. When you pass silence to it, the 'thank you' comes out, It's possible that there's a sampling strategy that is better as well, but largely, hopefully, we can rely on the VAD filter to effectively filter out any silence.

really appreciate you noticing and running into this bug as well.

cjpais:
Generally this has been much improved with the on-demand mode and the VAD filter, especially with the improvements of the refactoring of the audio back end. If this becomes a significant problem again, we will reopen the issue.

cjpais:
Reopening it for now as occasionally I am still getting it. There's going to be a debug mode added soon which hopefully will help in addressing this issue permanently.

Anaphylaxis:
Sometimes I just get `!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!` instead of any real output. Can't see anyone else with this issue. Seems to be fixed if I swap from the model to another model (and even back again), so may be an issue with loading the model correctly? I can load medium, have one transcription, then future ones are the !!!!!s. Maybe it's failing to transcribe somehow? I have this issue on small, turbo, medium, and haven't tried large.

<img width="492" height="450" alt="Image" src="https://github.com/user-attachments/assets/a16b0199-c530-4f9b-9b2a-9bde7d1fc582" />

Logs are here: https://pastebin.com/mBgR8QYt

cjpais:
[...]

PR #111, opened 2025-09-10
Parakeet V3

cjpais:
This is a fairly substantial PR which adds support for NVIDIA Parakeet v3 model. From the basic testing I've done, it runs really well on a wide variety of hardware, including CPUs. Right now the inference for parakeet is done only on the CPU with no GPU acceleration. However, I think that this is fine for most computers at the moment.

Lower power computers this might not be okay for, but we will find out as more users report in.

One great thing about this is that we also rewrote the interface into the transcription engines, so we can support both Parakeet as well as Whisper very nicely. And potentially in the future, this would enable us to support API providers or something like that a little bit more friendly. This is done in the `transcription-rs` crate I published.

This also addresses #18. #11 does not apply with Parakeet. That is a whisper hallucination. Maybe #77 is partially addressed for those running on CPU, though API should be an option still. It may address #91 for users who do not need Whisper


alexonpeace:
I didn't know if I have to create an issue to voice my opinion but yes the Parakeet v3 is really fast on my PC not so much on my laptop, then again nothing is fast on that thing, is GPU acceleration on the timeline?

cjpais:
Nah thank you for your feedback. I may add support for GPU acceleration too it may just take some time. It's definitely possible and I left it out for now to just get the release out

PR #115, opened 2025-09-13
feat: Add history enable/disable setting

vladstudio:
- Add history enable/disable toggle in settings
- Conditionally save transcriptions
- show storage size
- delete history

image

### DIFF
--- a/src-tauri/src/actions.rs
+++ b/src-tauri/src/actions.rs
@@ -112,17 +112,20 @@ impl ShortcutAction for TranscribeAction {
                             transcription
                         );
                         if !transcription.is_empty() {
-                            // Save to history
-                            let hm_clone = Arc::clone(&hm);
-                            let transcription_for_history = transcription.clone();
-                            tauri::async_runtime::spawn(async move {
-                                if let Err(e) = hm_clone
-                                    .save_transcription(samples_clone, transcription_for_history)
-                                    .await
-                                {
-                                    error!("Failed to save transcription to history: {}", e);
-                                }
-                            });
+                            // Save to history if enabled
+                            let settings = get_settings(&ah);
+                            if settings.history_enabled {
+                                let hm_clone = Arc::clone(&hm);
+                                let transcription_for_history = transcription.clone();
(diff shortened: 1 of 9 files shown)


cjpais:
Overall, I'm happy to pull this in, however there's a few changes I would suggest. Or some things I have questions about.

Mainly, I think the size on disk and delete history button probably should be removed. Mainly because you're really never going to get more than five history messages ever in the current implementation. If we change the setting so that you can customize the history size, then it might make sense to have that button. However, I think we can just put a little X next to each individual history item itself. To me, this seems like a much cleaner UI and leaves it much less cluttered.

I also think that the enable history setting either should be in advanced. And perhaps the history tab disappears entirely if the history setting itself is disabled. 

It may be helpful to add a note in the history bit itself, probably italicized, that the history only saves the five most recent transcriptions. I am not sure how best to display this. I bet it could probably just be some light gray text that's italicized at the top describing how the history feature works. But again, I'm a little concerned about how it clutters the UI, so just keeping the UI fairly clean is most important here. Maybe it is on the same level of the history label itself and pushed off to the right side of the page. I don't know if that would work or not, but maybe it does.

And I guess my overall question is why do you want to disable the history? I personally would prefer not to have an additional setting because I think it just clutters the UI and gives too much option to the user. It's not clear to me why it would really need to be disabled, but perhaps that's just my opinion.

hanpham32:
Oh wow, just saw this feat PR. If the server only keeping 5 audio entries on system at all time and delete prior one then I don't think the delete all is needed?

My PR at #119 covers this and implements the single deletion for entries

vladstudio:
"never going to get more than five history messages ever" --- ooh, I did not catch that. I somehow assumed the history is "infinite" and my instinct was - I do not want that. Let me just close this PR then. 

Adding a note to the UI about 5 messages limit might be a good idea!

bobcate:
I was about to make a feature request for disabling history.
Looks like [it's implemented](https://github.com/cjpais/Handy/discussions/358#discussioncomment-15019102) already.
It doesn't work, opened https://github.com/cjpais/Handy/issues/460

Issue #108, opened 2025-09-08
Feature Request: Clear History


rb81:
**Great work with Handy! ❤️**

Nothing special to this request. The ability to delete an entry or clear all entries from the history would be great. (For now, simply deleting the recordings folder and history.db file does this.)

cjpais:
Thank you, and I forgot to respond to this when I originally saw it. 

I will probably add a small x next to each recording, so you can clear it from there. I probably won't add a clear all for now, because the mode is so simple and there's only like five or so things to click anyhow.

rb81:
I think the clear all feature is quite important if you want people to use this regularly. After a while, the number of recordings can add up quite quickly. I would also suggest having an option to disable history so that recordings are discarded as soon as they're processed.

cjpais:
The number of recordings is currently capped at 5. I don't expect people to be referring to the history for long-term things because that is not the application's purpose

This may change in the future, but for right now it's really meant to just be for inputting things into text fields

rb81:
That makes sense. But then maybe just getting rid of the history altogether would also make sense in that case :)

cjpais:
No, history was implemented because occasionally if you paste into the wrong text field or none at all and you wanna come back to the history. In addition, it's very helpful for me for debugging transcription errors

hanpham32:
I think it would be nice to be able to delete all entries at once when needed.

Karim-at-Powl:
+1

cjpais:
#119 closes this

PR #119, opened 2025-09-15
feat: delete a single audio entry

hanpham32:
This PR adds a new feature to Handy, allowing users to delete single audio entry from history.

Closes issue #108 

## Changes:
Backend (tauri): 
- added `delete_entry()` to `HistoryManager`
- added `get_entry_by_id()`
- implemented `delete_history_entry`
- event system for UI updates

Frontend:
- added confirmation dialog, loading states with spinner
- updated typescript interface for `HistoryEntryProps`

### DIFF
--- a/src/components/settings/HistorySettings.tsx
+++ b/src/components/settings/HistorySettings.tsx
@@ -152,7 +152,6 @@ const HistoryEntryComponent: React.FC<HistoryEntryProps> = ({
 }) => {
   const [audioUrl, setAudioUrl] = useState<string | null>(null);
   const [showCopied, setShowCopied] = useState(false);
-  const [isDeleting, setIsDeleting] = useState(false);
 
   useEffect(() => {
     const loadAudio = async () => {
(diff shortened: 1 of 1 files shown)


cjpais:
Thanks for this I'll take a look soon!

cjpais:
I took a look and it doesn't seem to work? When I click the button it does nothing

hanpham32:
Oh weird, it worked for me yesterday. Let me try it again. What OS are you on?

cjpais:
MacOS but this is primarily a UI change with backend hooks and shouldn't be OS specific as far as I can tell.

Probably the window 'are you sure you want to delete' is not cross platform enough. We can remove the confirmation dialog imo.

hanpham32:
I have removed the window.confirm modal. I hope that will fix that. 

I think this code is not working on your end:
```ts
const confirmed = window.confirm(
  "Are you sure you want to delete this entry? This action cannot be undone."
);
```

cjpais:
great thanks, if you could, could you also remove the code which has the state for `setIsDeleting`, I noticed it causes the cursor to flicker to a disabled state. Realistically this operation should almost always be instant (on the order of milliseconds) unless there is an error.

hanpham32:
Agree, the loading state is not necessary. I pushed the latest change and that should have instantaneous effect now!

cjpais:
Thank you so much for making the changes requested! Makes my life so much easier. I'll have some time to review tomorrow but it should be pulled in!

cjpais:
looks good works well, thanks!

PR #136, opened 2025-09-28
Support mac Fn shortcut handling

tekacs:
- special-case "fn" bindings via rdev listener so mac users can use Fn
- reuse shortcut dispatcher to drive existing push-to-talk/toggle flow
- expose Use Fn button on mac settings and keep Cmd glyph formatting

### DIFF
--- a/src-tauri/src/fn_monitor.rs
+++ b/src-tauri/src/fn_monitor.rs
@@ -1,0 +1,141 @@
+use once_cell::sync::Lazy;
+use rdev::{listen, Event, EventType, Key};
+use std::collections::HashMap;
+use std::sync::atomic::{AtomicBool, Ordering};
+use std::sync::{Arc, Mutex};
+use std::thread;
+use tauri::AppHandle;
+use tauri_plugin_global_shortcut::ShortcutState;
+
+use crate::settings::ShortcutBinding;
+use crate::shortcut::dispatch_binding_event;
+
+pub(crate) fn register_fn_binding(app: &AppHandle, binding: ShortcutBinding) -> Result<(), String> {
+    ensure_listener_started();
+
+    let mut state = MONITOR_STATE
+        .lock()
+        .map_err(|_| "Failed to lock Fn monitor state".to_string())?;
+
+    state.bindings.insert(
+        binding.id.clone(),
+        FnBindingEntry {
+            app_handle: app.clone(),
+            binding_id: binding.id,
+            shortcut_string: binding.current_binding,
+        },
+    );
+
+    Ok(())
+}
+
+pub(crate) fn unregister_fn_binding(_app: &AppHandle, binding_id: &str) -> Result<(), String> {
+    let mut state = MONITOR_STATE
+        .lock()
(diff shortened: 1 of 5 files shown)


akshar-dave:
🙏🏻

cjpais:
Thanks for this I'll give it a look and test. Do you know if you `tauri build` the release does it require extra permissions for rdev on MacOS? The initial versions of Handy used `rdev` and a custom global keyboard shortcut library, but required extra permissions

tekacs:
Hmmmm. I have it using input monitoring – is that something that you have in your main branch(es) as well?

cjpais:
Yeah that was actually the reason I moved away from it. I felt like it might feel too much like a key-logger or something which I was hesitant to add the permission for

tekacs:
I hear that – the thing is that I think it's... almost necessary? For push-to-talk, holding two keys is distinctly less than ideal.

Interestingly, both Wispr Flow and VoiceInk don't use that permission, so it's likely spurious? If I get a chance I'll investigate why.

cjpais:
Right, I've been trying to figure that out as well. I used MacWhisper for a while and don't recall the permission either. There must be some system native way of doing this

tekacs:
@cjpais this now works using only Accessibility permissions – tested by removing the Input Monitoring permission and using `fn` successfully. Should supersede #163 too.

cjpais:
Let's go, thank you! I cannot wait to test it. Hopefully I'll be able to on Tuesday and pull it in. Thanks

cjpais:
I tested this, however, I haven't looked at the code thoroughly yet. I do see that this only supports the function key. It would be really cool to move the entire code to use this general method of input detection on Mac OS, and move away from using the Tauri Global shortcut. This is so we have consistent behavior on Mac OS

Ideally can also detect `fn` in the frontend instead of adding a UI button. I don't really want that UI button to exist, the user should just be able to click the keyboard ideally. I would think javascript can get the keycode and we might just need to map it. I believe the other PR supports that however I have not tested it

dannysmith:
@cjpais Where's the best place to discuss this at the mo? Feels like there are a bunch of issues/PRs/discussions about handling keyboard shortcuts.

My tuppence...

- I'd very much like to use `fn` as my PTT trigger on macOS. 
- Other people seem very keen on all sorts of other key mappings for PTT/persistent or plain/translated/LLM-enriched etc etc etc.
- Supporting complex keymappings via a not-horrible UI is likely to take a little thought. I wonder if it'd make more sense to support more complex keymappings via user editing of `settings_store.json` or some other config file? 

**Related...**

I'd also like the ability to disable Handy's shortcuts when certain apps are focussed (ie those with conflicting app-local keybinds)

cjpais:
I'm very happy to support mappings through json right now and I think that's a good tradeoff before UI comes in. 

In regards to handling more keybinding generally I think we can do it here or in an issue if there is one. I really want this one to come in as well it just needs testing and validation that it won't require special perms on macOS, and it should be generalized beyond the fn or globe key I think to support any binding within reason

cjpais:

PR #137, opened 2025-09-28
Add a timer for microphone keep-alive, rather than simply 'forever' or 'not at all'

tekacs:
- replace always-on toggle with a keep-alive dropdown (including Forever)
- persist new MicrophoneKeepAlive setting and map legacy always_on values
- update audio manager to schedule mic shutdowns based on the timer


cjpais:
What is the reasoning behind adding this? I'm not sure the exact specific use case it's solving and I'm hesitant to add more bloat to the codebase especially the microphone path as it should be refactored.

tekacs:
You say in the message that this 'may prevent the Mac from sleeping'. Despite that, I wanted to turn it on for low latency n-th transcription. Setting it to a keepalive like a minute or minutes meant that in practice 'most of the time' the microphone is alive, but there's also no chance of it keeping the Mac awake, since it _will_ self-terminate.

I considered just proposing that the 'always on' is... not 'always' but just... a few minutes or something. The reason that I made it configurable is because for e.g. AirPods or bluetooth devices that change modes when the input is open, I thought that someone might care about the duration.

cjpais:
hmmm okay, let me think about this one a bit

curious, did you leave it off and run into issues? I thought I would run into lots of issues regularly without it, but for the most part it's not a problem for me

tekacs:
I did not, I just went straight for making it take on a timer.

cjpais:
I'm going to close this one for now. Let me know if this is an issue. I don't expect many people to be using this feature however if it comes up, we will have this code and bring it in then. Mainly just trying to keep the codebase as simple as possible for now. It's already too complex for my liking for what it does currently. Thanks for contributing!

PR #144, opened 2025-09-28
drop libsamplerate feature to fix macos build

alextes:
## Summary
- disable `vad-rs` default features so the helpers module (and libsamplerate-sys) stay out of our dependency tree
- refresh Cargo.lock; the git dependency now pulls only pure-Rust crates

## Why
Modern macOS installs ship CMake ≥3.30, which refuses to configure the vendored libsamplerate sources pulled in by the `samplerate` helper. The build blows up with:
```
error: failed to run custom build command for libsamplerate-sys v0.1.12

--- stderr
CMake Error at CMakeLists.txt:1 (cmake_minimum_required):
Compatibility with CMake < 3.5 has been removed from CMake.
```

Dropping the helper removes that dependency entirely, so the build succeeds again.

  ## Testing
  - cargo check
  - bun run tauri dev

### DIFF
--- a/src-tauri/Cargo.lock
+++ b/src-tauri/Cargo.lock
@@ -1075,15 +1075,6 @@ dependencies = [
 ]
 
 [[package]]
-name = "dasp_frame"
-version = "0.11.0"
-source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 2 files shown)


cjpais:
Cheers for this one, I'll give it a test and merge, thanks

PR #150, opened 2025-09-30
Add history limit setting

sbeckeriv:
Dearest Reviewer,

I have added a history limit setting. The limit change is applied instantly to the history. If the limit is 0 then no recordings are saved.

Becker

### DIFF
--- a/src-tauri/src/commands/history.rs
+++ b/src-tauri/src/commands/history.rs
@@ -44,8 +44,24 @@ pub async fn delete_history_entry(
     id: i64,
 ) -> Result<(), String> {
     history_manager
-        .as_ref()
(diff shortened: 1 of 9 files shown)


cjpais:
Thanks I'll take a look next week

cjpais:
Do you mind moving this into the debug menu? I want to pull in some of these features but I want to hide them from most users for the time being before I have a better idea how to represent many in the UI. I have an alternative interaction method I'm thinking about, so let's get features in, just mostly hidden.

sbeckeriv:
Moved

Screenshot 2025-10-01 at 9 50 31 PM

cjpais:
Thank you looks good!

PR #163, opened 2025-10-01
added support for macos function/globe key

jtviolet:
M4 MacBook Pro, fn/globe key is registering as unknown key 179. This key was not registering with Handy to use as a modifier key for Handy (nor the solo shortcut key).

This update adds the ability to use the fn key as (or as part of) your shortcut.


cjpais:
I appreciate the contribution. This is similar to #136. See the discussion there. 

In a nutshell. I am hesitant to pull these changes in because I believe it requires an extra permission on MacOS. I originally used `rdev` across the app instead of Tauri's Global Shortcut because Global Shortcut does not allow as many key combinations. But the problem with `rdev` is that it requires an extra permission which I am not super fond of that particular permission (I believe it is full keyboard access). There are other apps that are able to use the fn key without this permission so I think that should be our target for merge. I am not sure this is possible at the moment, but I think it's worth investigating to know for certain.

Also please note that that permission probably doesn't show up in the dev env (at least it didn't for me) so you may need to `tauri build` to see it. Let me know if it works without this permission I will gladly give it a try and pull if it does work nicely.

Ideally we also should have a single unified way of keyboard shortcut detection, and if that means moving away from Global Shortcut, I am happy to do that. But is something to keep in mind. We may pull something in earlier, but it really should come with the mindset as a replacement rather than bolting on another codepath.

dereklputnam:
This is great! I preferred using this button when I used WisprFlow, but I missed it with Handy.

alvinometric:
I definitely never allow the full keyboard permission, it's a pretty scary prompt

cjpais:
@jtviolet, would you mind running `bun tauri build` on your machine and installing the dmg and letting me know if the full keyboard permission shows up?

On my machine I did test it after building and I don't get the permission request! I may pull this in and will try to distribute a test build to anyone following this thread to validate we don't have that permission showing up. Ideally I'd like unified keyboard handling on MacOS to all use the same code, but this is a big enough feature request it will get pulled in if we can sanity check it

alvinometric:
Just my 2 cents but I looked at the code again, and I don't think it's worth the potential side-effects for such a niche use-case:
 
- Introduces another `useEffect`
- Bifurcates the logic completely which makes debugging much harder, I'd rather Chris's time was spent elsewhere

panckreous:
What about adding url-scheme support? Then we could use Alfred/Raycast/BTT/Keyboard Maestro/streamdeck/etc to detect the key combo and trigger the event in Handy. 

(You may even get more fun physical implementations for the Buttons page as a bonus)

A great implementation example (though way more complex than what this would require) is CleanShot X's:  
https://cleanshot.com/docs-api

cjpais:
I'm happy to review/pull in a PR which add's URL scheme support for start/stop for sure. Or generally making handy more programmatically controllable via a method like this

cjpais:

Issue #93, opened 2025-08-22
[BUG] Paste Not happening (Dvorak)


winkler1:
## Bug Description
Clipboard has the text but paste does not happen. Dvorak keyboard.

## System Information
**App Version:**
0.3.7

**Operating System:**
Mac 15.5

**CPU:**
Apple M1 Max

**GPU:**
Apple M1 Max

FWIW, this is what Aider suggested as a fix. Not a Rust dev, so not able to build/test.

src-tauri/src/clipboard.rs
> I'm a Dvorak user. The text is on the clipboard but does not get pasted. On a Mac.

It looks like you're having trouble with pasting on macOS with a Dvorak keyboard layout. This is happening because the application is simulating the keypress for 'V' based on its physical location on a QWERTY
keyboard, rather than the character itself.

To fix this, I'll change the macOS-specific code to use a layout-aware key, which is how it's already handled for Linux in send_paste(). This will ensure the paste command (Cmd+V) works correctly on any keyboard
layout.

Here are the changes for src-tauri/src/clipboard.rs:

src-tauri/src/clipboard.rs
<<<<<<< SEARCH
    #[cfg(target_os = "macos")]
    let (modifier_key, v_key_code) = (Key::Meta, Key::Other(9));
=======
    #[cfg(target_os = "macos")]
    let (modifier_key, v_key_code) = (Key::Meta, Key::Unicode('v'));
>>>>>>> REPLACE

cjpais:
It was changed to 9 to fix #50 

Surely there's a way to fix both but I will need to test or have help in testing

winkler1:
Tested again, clipboard not being set. Will try to get at the debug logs.

cjpais:
Sorry to be more clear this issue almost certainly is because of that other fix. Aider is suggesting to change it back to 'v' which will just have the other bug pop up. It's a bit of whack-a-mole which needs a better solution.

potts2020:
I am having this same issue using Dvorak. It seems when I speak, the software just pastes whatever is on my clipboard. What are the odds, having two Dvorak users?

App Version:
0.3.8

Operating System:
Ubuntu 24.04.2 LTS

CPU:
AMD Ryzen 9 5900X

GPU:
AMD Radeon RX 6800 XT

cjpais:
Ha pretty amazing to have two Dvorak users, thanks for trying it! I will try and get this fixed but it may take me some time

I'm also happy to accept PRs :)

alextes:
[...]

Issue #164, opened 2025-10-02
[BUG] Some non-native apps miss modifier key press and register only 'v' keystroke


justnero:
## Bug Description

Some non-native apps do not consume paste, leaving a text field with a lonely `v` in it. This happened to me in Minecraft chat, only `v` was sent to the text field. At first, I thought it was not supporting Cmd+V for some reason as it's not a Cocoa app, instead it uses LWJGL/GLFW stack and handles raw keystrokes. But in reality, it was just not fast enough to register fast modifier release.

## System Information
**App Version:**
0.5.1

**Operating System:**
macOS 15.6.1

**CPU:**
Apple M1 Max

**GPU:**
Apple M1 Max GPU

cjpais:
Thanks for reporting, I thought this was resolved. Also thanks for the PR

a-b:
`fn` key is not working either. I noticed that when I turn off `Push To Talk`, more combinations seem to work.

cjpais:
Next release should have a fix

cjpais:
Closing this as you fixed it @justnero in #165 and it's in 0.5.3

PR #165, opened 2025-10-02
fix: delay releasing modifier key on paste to let slow consumers handle the keystroke

justnero:
Addresses https://github.com/cjpais/Handy/issues/164.

It both adds a delay between pressing the `v` and replaces press+release `v` with just `click` which is effectively the same thing. 

It works for me in any app now, no matter what layout is set and which text field it is.

Unfortunately, I could not confirm if it's breaking anything for other platforms.

### DIFF
--- a/src-tauri/src/clipboard.rs
+++ b/src-tauri/src/clipboard.rs
@@ -24,13 +24,11 @@ fn send_paste() -> Result<(), String> {
         .key(modifier_key, enigo::Direction::Press)
         .map_err(|e| format!("Failed to press modifier key: {}", e))?;
     enigo
-        .key(v_key_code, enigo::Direction::Press)
+        .key(v_key_code, enigo::Direction::Click)
         .map_err(|e| format!("Failed to press V key: {}", e))?;
 
-    // Release V + modifier (reverse order)
-    enigo
-        .key(v_key_code, enigo::Direction::Release)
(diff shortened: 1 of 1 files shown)


justnero:
No way to confirm, but it might also resolve https://github.com/cjpais/Handy/issues/93 as symptoms are quite similar.

cjpais:
Interesting I used to run into this bug quite often but I haven't seen it on my machine in a while.

The code looks good, but I will wait to test it until next week, I'm quite busy. I ideally would like to avoid paste delays, but 'v' is much more annoying. Perhaps click breaks things on other platforms but it will need to be tested. I'll probably go read the Enigo implementation when I review properly

It most likely will not resolve #93, that is a different bug as far as I'm aware

justnero:
Well, the paste will happen as fast as it can be, the only thing delayed is modifier key release, but it does not imply paste delay. At least this is how it works for macOS. I hold the modifier key, then press V, and it's already inserted even though I didn't release the modifier key yet.

P.S. I'm now writing my PR comments with Handy. Thank you very much for making this great tool.

cjpais:
Sounds good thanks, I'll hopefully be able to test it this week on multiple platforms and pull in. It may not be able to be tested until next week

cjpais:
Works on my Mac! Alright I decided im just gonna pull it in and when I do my cross-platform testing if it breaks things on other platforms I'll revert. Thanks for the PR!

PR #167, opened 2025-10-02
Paste shortcut configuration for every layout

alextes:
i don't think this is mergable yet. its AI written and i don't quite understand it. i asked it to explain, explanation below. i did test the changes and resolved a bug around detected key capture. at first we still had the exact same issue of my dvorak layout resulting in cmd+v capturing as cmd+. instead, which is the qwerty equivalent. maybe a maintainer with more experience can review and see if all proposed changes make sense.

- Add a settings control that captures the paste combo using layout-aware key detection.
- Persist the binding through the store and enforce the same parser in the Tauri backend, including a reset endpoint and safer defaults.

rationale for large diff
The diff looks large because we had to bridge both ends of the feature instead of just adding a toggle. The existing paste logic was hard-coded to “Cmd/Ctrl+V” using literal keycodes, which breaks
on non-QWERTY layouts (Dvorak, AZERTY, etc.). To make a user-configurable binding actually work, we needed three changes:

- Frontend capture: The settings component must read actual key events, normalize them across platforms, show the recorded combo, and reset/cancel cleanly. That’s why PasteShortcut.tsx includes the
event handlers, refs, OS detection, and display formatting.
- Keyboard utilities: The prior helpers relied on event.code, which always reports the physical key (QWERTY assumption). We switched to layout-aware event.key, added normalization, and kept
formatting consistent, so recording “Cmd+,” on Dvorak doesn’t display as “Cmd+V”.
- Backend parsing: The Tauri side still needs to send the recorded combo to Enigo. That required parsing the stored string into modifiers + trigger key, validation, and a fallback to the default
combo if a custom one fails. Without reworking clipboard.rs, the UI change wouldn’t actually paste correctly.

So while the UI addition is just one row in Settings, making it dependable for any layout meant refactoring both capture and send paths.

resolves https://github.com/cjpais/Handy/issues/93

### DIFF
--- a/src-tauri/src/clipboard.rs
+++ b/src-tauri/src/clipboard.rs
@@ -1,45 +1,201 @@
-use enigo::Enigo;
-use enigo::Key;
-use enigo::Keyboard;
-use enigo::Settings;
+use crate::settings;
+use enigo::{Direction, Enigo, Key, Keyboard, Settings};
 use tauri::AppHandle;
 use tauri_plugin_clipboard_manager::ClipboardExt;
 
-/// Sends a paste command (Cmd+V or Ctrl+V) using platform-specific virtual key codes.
-/// This ensures the paste works regardless of keyboard layout (e.g., Russian, AZERTY, DVORAK).
-fn send_paste() -> Result<(), String> {
-    // Platform-specific key definitions
-    #[cfg(target_os = "macos")]
-    let (modifier_key, v_key_code) = (Key::Meta, Key::Other(9));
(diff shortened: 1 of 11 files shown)


cjpais:
I'll take a look next week and see. From the description you wrote here I'm a bit hesitant but won't know until I have time to test and review the code

cjpais:
I'm closing this for now I think #187 should solve, thank you for the PR and looking into it!

alextes:
@cjpais wonderful! much simpler in your version.

PR #177, opened 2025-10-06
feat(settings): add autostart toggle in advanced settings

Gr3mIin:
Enable users to configure autostart behavior through a new toggle in the advanced settings UI. The backend now checks user preferences on startup to enable or disable autostart accordingly, with the setting persisted and manageable via a new Tauri command. This provides better control over app initialization without affecting existing functionality.

### DIFF
--- a/src-tauri/src/lib.rs
+++ b/src-tauri/src/lib.rs
@@ -137,10 +137,17 @@ pub fn run() {
             // Initialize tray menu with idle state
             utils::update_tray_menu(&app.handle(), &utils::TrayIconState::Idle);
 
-            // Get the autostart manager
+            // Get the autostart manager and configure based on user setting
             let autostart_manager = app.autolaunch();
-            // Enable autostart
-            let _ = autostart_manager.enable();
+            let settings = settings::get_settings(&app.handle());
+
+            if settings.autostart_enabled {
+                // Enable autostart if user has opted in
+                let _ = autostart_manager.enable();
+            } else {
+                // Disable autostart if user has opted out
+                let _ = autostart_manager.disable();
+            }
 
             // Window is configured to start hidden to avoid flicker.
             // If user didn't choose Start Hidden, show it now.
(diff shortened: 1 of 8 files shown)


cjpais:
Thanks! I'll take a look tomorrow!

cjpais:
@Gr3mIin just leaving a note for myself, it doesn't seem to work properly on MacOS currently. Not sure if you have any ideas but I'll take more a look soonish. It's just not the most convenient to test since it requires restarting my machine

Gr3mIin:
> @Gr3mIin just leaving a note for myself, it doesn't seem to work properly on MacOS currently. Not sure if you have any ideas but I'll take more a look soonish. It's just not the most convenient to test since it requires restarting my machine

@cjpais Not off the top of my head. I don't have access to a Mac, so I can't really test it myself, I will have a look into the code a little bit more and see if I can come up with some theories as to what might be happening.

Quick question, when you're testing the auto launch, after toggling on the setting, do you restart Handy before restarting your computer?

I just wanted to check this because when you toggle on the setting in Handy you have to restart Handy to activate the new setting as the auto launch preference setting is only handled on startup of Handy. I did say in the tooltip "Takes effect on next restart" however I can see that that may be a little vague and people may think it means restart their computer.

cjpais:
I did just restart my computer. I actually had a change queued up to immediately have the setting reflect, but discarded it. I think it might be best to just immediately change the behavior without needing an app restart if possible

Also I probably have a lot of versions of Handy registered on my machine, so I may need to make sure all of them are cleaned up..

Gr3mIin:
@cjpais Sorry, I have been a tad busy this morning. I have only just gotten around to seeing the new updates to the PR; I've tested the changes that you made in your commit "immediate toggle effect?" and that does seem to work on both Windows 10 and Windows 11.
I don't currently have access to my Linux machine, but I can test that sometime later this week if need be.
As I stated before, I don't have access to a Mac at all, so if you could let me know how your testing goes once you get around to it.
I had a quick peruse of the tauri-plugin-autostart documentation and I couldn't see any reason why it wouldn't work on macOS, so please let me know if your changes have made a difference to the functionality on macOS.

I've also resolved the merge conflicts that had been introduced.

cjpais:
Thank you, appreciate everything you've done already. I'm guessing it's a side effect of my system. I'm going to try it on a clean machine tomorrow and see what happens. From the code I don't see any reason why it shouldn't work

cjpais:
Was able to test this on a fresh Mac and it seems to work great, going to pull it in, thanks!

PR #178, opened 2025-10-06
use Ctrl+Shift+V for paste on Linux terminals

### LINKED ISSUE #138: [BUG]  Handy doesn't paste into apps on Xubuntu
## Bug Description
I'm running Handy (thanks for making and sharing this BTW!  This is exactly what I had been looking for!) on Xubuntu.

I activate Handy's listening mode (either by Push To Talk, or else by toggling the listening mode) and it does listen (it displays the capsule, and the capsule shows activity as I talk).  If I tell it to stop listening (by releasing the push to talk button, or else by pressing the key combo to toggle the listening off) then whatever i just said will show up in Handy's history.

however, it doesn't get pasted into the thing that has the focus.

LibreOffice Writer didn't actually get the text
Nor did a Google Doc document, nor did the text input box for the Google Gemini web page.

If anyone has advice or ideas on how to troubleshoot / debug this I'd love to try them out.

## System Information
**App Version:**
0.5.0

**Operating System:**
Ubuntu 24.04.3 LTS, Xubuntu
Xfce version: 4.18
GTK version: 3.24.41

**CPU:**
Intel Core i7-10610U CPU @ 1.80 GHz x 8

**GPU:**
Mes Intel UHD Graphics (15.6 GiB)

vaknin:
## Issue
Paste functionality failed in Linux terminals because Ctrl+V is reserved for terminal control signals (SIGINT). It worked in GUI apps but not terminal emulators.

## Solution
Changed Linux to use Ctrl+Shift+V, the standard paste shortcut for Linux terminals.

## Why not `enigo.text()`?
The current architecture writes to clipboard then pastes, allowing clipboard restoration after paste (lines 59-75
in clipboard.rs). Using `enigo.text()` would bypass the clipboard entirely, preventing restoration of the user's
original clipboard content. The Ctrl+Shift+V approach preserves this feature with minimal changes and no new
dependencies.

## Testing
- Compiles successfully
- Works in terminal emulators
- Works in GUI applications
- No new dependencies added

Closes #138 and possibly #93

### DIFF
--- a/src-tauri/src/clipboard.rs
+++ b/src-tauri/src/clipboard.rs
@@ -19,18 +19,32 @@ fn send_paste() -> Result<(), String> {
     let mut enigo = Enigo::new(&Settings::default())
         .map_err(|e| format!("Failed to initialize Enigo: {}", e))?;
 
-    // Press modifier + V
+    // Press Ctrl (or Cmd on macOS)
     enigo
         .key(modifier_key, enigo::Direction::Press)
         .map_err(|e| format!("Failed to press modifier key: {}", e))?;
+
+    // Linux terminals need Ctrl+Shift+V (Ctrl+V is reserved for terminal control)
+    #[cfg(target_os = "linux")]
+    enigo
+        .key(Key::Shift, enigo::Direction::Press)
+        .map_err(|e| format!("Failed to press Shift: {}", e))?;
+
+    // Press V
     enigo
         .key(v_key_code, enigo::Direction::Press)
         .map_err(|e| format!("Failed to press V key: {}", e))?;
 
-    // Release V + modifier (reverse order)
+    // Release in reverse order
     enigo
         .key(v_key_code, enigo::Direction::Release)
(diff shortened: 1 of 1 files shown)


cjpais:
mmmm... im not confident this approach. i do think this will resolve the terminal and probably work properly everywhere, but I would prefer to have two options (one being enigo.text) like #87 in the advanced settings ui

cjpais:
I think #187 should solve this. Thanks for the PR. Happy to reopen if this is ultimately a better solution or used in tandem

vaknin:
Hey CJ!

#### tl;dr enigo.key > enigo.text

I've tested your changes, and it's indeed working.
In my personal opinion enigo.text() gives a worse experience due to it manually typing character by character which makes the user wait for it to finish typing.
Can be quite annoying on long recordings, but I wouldn't say it's a deal-breaker.
It's basically good-enough for Linux users right now, but if you ever want to make it smooth, I'd look into implementing enigo.key() instead.

thanks for taking the time to fix it
<3

PR #181, opened 2025-10-07
Load model in parallel to recording

jacksongoode:
I think this is quite simple change to load the model in parallel to recording. The effect of this is simply that recordings that are longer than the time it takes to load the model from scratch will feel instantaneous and all transcriptions from a cold model load will at least be somewhat faster.


cjpais:
Thank you! Will take a look later today or this week!

cjpais:
Just want to say I've done a once over this code and I want to take a slightly deeper look at it. I just want to make sure that there's not some other pattern existing in the code already that we could take advantage of. Or if there's a simple refactoring that could be done of the code to just generally improve it for the future as well and aid in maintainability. But I do need to take a deeper look myself. It's been a while since I've reviewed this part of the code base.

cjpais:
Alright I'm good with it, let's go. Thanks @jacksongoode!

PR #182, opened 2025-10-07
A few optimizations on startup

jacksongoode:
One more:
- Use a few common flags for a more optimal release profile
- Load managers & hotkeys first on init (to reach a "ready" state faster)
- Add a simple splash when we do want to load the gui first
- Simplify app handler (this seemed like an easy cleanup win)

I haven't let benchmarked the start up time yet but it at least feels somewhere around half a second faster to trigger the microphone on start.


cjpais:
Wow ty! Will look soon!

cjpais:
I did just take a quick glance over the changes. I think overall everything is good. I just want to do some cross-platform testing as well as making sure that the release still works on all platforms. I remember the app image having some trouble building in certain scenarios, and I wonder if any of the release config will break that. Just validating that all the startup options work properly is another thing.

cjpais:
I hope you don't mind I'm removing the splashscreen, it didn't display very nicely ootb on my Mac and im okay not having one for now.

Generally the cleanup looks good to me, and super thankful for it. Definitely is more sane to read and the ordering is better. I'm just gonna pull it in and before going for a release I'm going to test on all platforms anyhow. If there's a bug I'll fix it then

@jacksongoode if you're good with the changes im good to pull them in. Happy to wait if there's anything else you want to clean up in particular too. I'm just catching up with all the PR's and major bugs and trying to get things in. The next release will have a bunch of these minor bug fixes and small QOL improvements

jacksongoode:
Yea totally understand with the splash, more of a suggestion (so I made it a draft). I would like to actually see that the other changes are leading to a measurable average increase in startup. Maybe logging at the point the recorder is ready would be enough.

src-tauri/src/lib.rs
```diff
@@ -49,10 +49,94 @@ fn show_main_window(app: &AppHandle) {
...
+    shortcut::init_shortcuts(app_handle);
+
+    // Apply macOS Accessory policy early if starting hidden
+    #[cfg(target_os = "macos")]
+    {
+        let settings = settings::get_settings(app_handle);
+        if settings.start_hidden {
+            let _ = app_handle.set_activation_policy(tauri::ActivationPolicy::Accessory);
+        }
+    }
```
jacksongoode: @cjpais Just wanted to double check that we can move this later, I don't know much about this? Maybe it should still be first?

cjpais: I'll double check that it works properly on my Mac, I did mess around with some stuff earlier, but I haven't tested extensively yet

jacksongoode:
@cjpais I had one more optimization now that #181 is merged in, see the latest commit.

Now that the model loads on _demand_ (and it typically takes maybe only a second to load) when a recording is started, I think it makes sense to not load the model on start at all.

I did measure this change and on my M1 Pro the startup (from the earliest point I could start a timestamp) to when we are ready to transcribe, the time to start went from ~800-1000ms to 200ms.

It's hard to feel the delay, this is from a cold start where I trigger right after I press "enter".

https://github.com/user-attachments/assets/4d63e53d-d1de-4207-ac78-065ecfabd7b7

With that, feel free to merge if all the other changes look sound.

Edit: I just noticed one minor regression, in the settings now the model loading indicator will flash indefinitely when Handy is first loaded and before any transcription. Maybe there's a simple fix for this? If that's blocking let me know.

cjpais:
Yeah I'm definitely good with loading it on demand when the first transcription happens instead of at app launch. 

The regression does point to some concern over this, potentially, which is if there is a failure to load the model for whatever reason, that won't be very visible to the user. And to be fair, I guess this is largely the case already. I think it would probably make sense to have some kind of idle state, potentially. And quite frankly, the status of the model is not very well unified between the backend and front end, and maybe there's a bit of help there that could be done just so it's a bit more consistent and represents the state properly everywhere. I'm not really sure how well the code will support that. I think the idle state could be green potentially as well. Cause the app more or less should be ready. I also think generally the status right now now displayed in the UI is not super clear and maybe needs to be removed overall. Mainly because there is already an unload model function, and I don't believe the UI is paying attention to when the model is actually unloaded. So there's already a desync of the state already. 

So this is really all to say. I'm not super concerned about the regression right now. However, if we could mark it as green for the time being, that would be great. And if you have any other thoughts on this state stuff overall, I'd be really, really happy to hear what you think might be an appropriate way forwards. Even if it comes in another PR, that's totally fine. I also don't know if you have any thoughts on how the model selector generally could be modified or made more user friendly in any ways, or if it's good enough as is. It didn't receive a lot of time and focus when it was initially implemented, and the whole thing is kind of a bodge.

Edit: I'm sorry I've merged some more of the code in this area, it probably needs a rebase again. I may get on with the 0.5.2 release and then stage this for 0.5.3. I'm hoping 0.5.2 will resolve an input issue on Linux and fix a bunch of outstanding issues in the repo

jacksongoode:
@cjpais Just rebased, and I added a "unloaded" model status so we start with that and then will load if clicked or a transcription begins.

cjpais:
Awesome, thanks! This will make it in soon and I think it'll make a big difference to UX especially on Windows and Linux, sometimes I think the app has crashed or might crash. Hope this helps, I'm gone through the weekend but hopefully will get it in next week.

I've been fighting with the CI but may try and kick off another build with this change

jacksongoode:
Anything more I can do to make this ready to merge?

cjpais:
@jacksongoode nah, I just need to review once more and test haha, if you're able to test and confirm things are good too that helps but I want to get this in too! Maybe this afternoon, I've just had a week

cjpais:
@jacksongoode thanks so much for this! pulling it in

I mostly just did a quick review of the code, if there's any regressions I'll probably add you to the respective issue. But as far as I've tested on my Mac it works great

PR #189, opened 2025-10-08
Add auto submit setting with configurable key modes

swissyai:
Adds an Auto Submit setting to Advanced that automatically sends a return key after transcription completes.

This enables seamless integration with agentic coding tools like Cursor and Claude Code that use different submit shortcuts. The setting supports three key modes:

- Enter: Simple return key
- Ctrl+Enter: Control + return combination  
- Cmd+Enter: Command + return combination (macOS)

The dropdown is disabled when the toggle is off to match existing UI patterns.

Implementation:
- Backend: Added AutoSubmitKey enum and send_return_key() function in clipboard.rs
- Frontend: New AutoSubmit component in Advanced settings following existing patterns
- Settings: Persisted via existing settings store with default of false/enter

Tested all three key modes functionally with both clipboard and direct paste methods.

### DIFF
--- a/src-tauri/src/clipboard.rs
+++ b/src-tauri/src/clipboard.rs
@@ -1,5 +1,5 @@
-use crate::settings::{get_settings, PasteMethod};
-use enigo::Enigo;
+use crate::settings::{get_settings, AutoSubmitKey, PasteMethod};
+use enigo::{Direction, Enigo};
 use enigo::Key;
 use enigo::Keyboard;
 use enigo::Settings;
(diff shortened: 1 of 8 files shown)


cjpais:
Thank you for submitting the PR. On the whole the use case makes sense. 

I'm a bit hesitant to pull this in at the moment, mainly I'm trying to avoid having a lot of settings (and there are kind of too many for my taste already). Right now I am mostly focused on things which are fixing bugs or are highly requested. I might pull this in eventually. If you don't mind, please also go collect feedback and support in the Discussions. I will be following there!

sasha-computer:
I would love this feature btw, I'm getting even lazier and would love to hit push-to-talk shorcut, talk, release and it enter the transcription into the box and press enter @cjpais. Any chance we can revive this? I'd be happy to push it forward in terms of getting it compatible with latest release etc.

cjpais:
@sasha-computer yeah I will accept this now. Put it in advanced settings and I will test and pull the PR in

sasha-computer:
> @sasha-computer yeah I will accept this now. Put it in advanced settings and I will test and pull the PR in

Amazing thanks - PR opened here: https://github.com/cjpais/Handy/pull/765.

Issue #208, opened 2025-10-13
Select language based on the shortcut


byme8:
It would be nice if you can define multiple shortcuts to trigger recognition. Then, depending on selected shortcut, it will initiate speech recognition with specific language.

For example, "command+l" will mean English and "command+k" will mean Ukrainian

cjpais:
Please move this over to the discussions tab. This thread might be a good place to collect shortcuts

https://github.com/cjpais/Handy/discussions/170

PR #213, opened 2025-10-14
Support for multiple custom keyboard shortcuts with different languages

byme8:
This PR adds an ability to create multiple keyboard shortcuts and then set up different language per each shortcut. #208 

Notice, it is mostly done via cloud code. I did the best code-review  I could give it(never touched tauri before) and tested local build. I haven't seen any problems, works completely fine for me.

What is changed:
- Now we have a list of shortcuts. You can press add button to add new one. Then once it's added you can change the shortcut command and also adjust the language.
- You can remove shortcut.
- The global language selection has been removed, now it only exists per shortcut.
- Refactored the useModels hook. Now there is no need to subscribe on model change in every component where you need to figure out what model is there. All of it done inside the hook.

The UI looks like that:

Screenshot 2025-10-14 at 11 00 09


cjpais:
Thanks for the PR, this is not going to be accepted at this time. It will come in the future but not yet. 

Please use discussions for this, and suggest what you want there. We will collect ALL of the ideas people want and then bring in a nice solution when it makes sense

Issue #196, opened 2025-10-11
[Feature Request] Set volume of audio feedback


rodalpho:
I like audio feedback when I start and stop transcription, but find the volume to be too obtrusive. I'd like the option to reduce it by maybe 50-75%.

rodalpho:
For anyone else with this issue, this is a pretty quick and dirty fix to drop the volume by 90%. Unfortunately upgrades will probably overwrite these files, though.

in C:\Users\\$yourUser\AppData\Local\Handy\resources

```
ren rec_start.wav rec_start.ORIG.wav
ren rec_stop.wav rec_stop.ORIG.wav
ffmpeg -i rec_start.ORIG.wav -filter:a "volume=0.1" rec_start.wav
ffmpeg -i rec_stop.ORIG.wav -filter:a "volume=0.1" rec_stop.wav
```

LoopControl:
Agreed, I also wanted an audio 'ding' but its so loud and lasts so long that I turned it off.

cjpais:
Similar to #113 

I am happy to accept any PR's regarding this. It should be fairly easy for someone to implement. Plus if you do there's a bounty on #113 so you might as well try and claim it!

Issue #113, opened 2025-09-10
[Bounty] Better Audio Feedback options


odie5533:
The current Audio Feedback option is distracting and loud. I'd like to choose some better, softer options. I'm on Windows.

Offering a bounty via PayPal of
 
* $10 to whoever creates a Pull Request and releases a build for me to use
* $10 to cjpais if it gets merged

cjpais is welcome to claim both!

cjpais:
I'm very open to softer options. If you have some audio files as examples that you'd be interested in, that would make it even easier for this to happen! There probably won't be multiple options, but just one because I would like to keep the program fairly simple. However, the output volume definitely could be tweaked with a slider. And yeah, other sounds are welcome.

stem-owo:
Personally, I use these two copyright-free audios and lower the application's volume to 15% on Windows~

[rec_start.wav](https://github.com/user-attachments/files/22270770/rec_start.wav)
[rec_stop.wav](https://github.com/user-attachments/files/22270771/rec_stop.wav)

odie5533:
@stem-owo Can you please share the source of those?

stem-owo:
Sure!

So for 'rec_start' I use this one : https://pixabay.com/fr/sound-effects/punchy-taps-ui-4-183899/
And for 'rec_stop' I use this one : https://pixabay.com/fr/sound-effects/punchy-taps-ui-6-183902/

(And converted to .wav)

cjpais:
release for this should be coming today

jacksongoode:
Hey @odie5533,

If it looks like the bounty was filled for you with my PR in the latest release, I'd love to put the funds towards relief in Gaza [here](https://www.mecaforpeace.org/donate/).

Thanks!

odie5533:
Both bounties paid!

@cjpais supported you on Github as that seemed the easiest

@jacksongoode 

<img width="1905" height="1231" alt="Image" src="https://github.com/user-attachments/assets/17b6c11d-65ba-4dce-97aa-2e2b4999bb20" />

cjpais:
@odie5533 you're awesome

PR #214, opened 2025-10-14
Add custom sound and control notif volume

jacksongoode:
Resolves https://github.com/cjpais/Handy/issues/196
Resolves https://github.com/cjpais/Handy/issues/113

Still a WIP right now but will update the description once this looks a little better.

Adds:
- A new "pop" sounds (as suggested by one discussion - subject to change/removal)
- Ability to add custom wavs as notifications for start/stop
- Preview notif sounds
- Control the notification volume

Screenshot 2025-10-15 at 11 09 38
Screenshot 2025-10-15 at 11 09 35

### DIFF
--- a/src-tauri/Cargo.lock
+++ b/src-tauri/Cargo.lock
@@ -187,27 +187,6 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "7c02d123df017efcdfbd739ef81735b36c5ba83ec3c59c80a9d7ecc718f92e50"
 
 [[package]]
-name = "ashpd"
-version = "0.11.0"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "6cbdf310d77fd3aaee6ea2093db7011dc2d35d2eb3481e5607f1f8d942ed99df"
-dependencies = [
- "enumflags2",
- "futures-channel",
- "futures-util",
- "rand 0.9.2",
- "raw-window-handle",
- "serde",
- "serde_repr",
- "tokio",
- "url",
- "wayland-backend",
- "wayland-client",
- "wayland-protocols 0.32.9",
- "zbus",
-]
-
-[[package]]
 name = "async-broadcast"
 version = "0.7.2"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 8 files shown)


cjpais:
Thank you so much @jacksongoode!

If possible, could we have the volume control be in the main settings there and then which sound in debug for now? Just trying to keep the main settings panel as simple as possible. Also if possible can we just keep it to pairs of sounds for now, instead of individual start and stop?

In the future may move these around to consolidate sound in one place.

jacksongoode:
> If possible, could we have the volume control be in the main settings there and then which sound in debug for now?

So the enable/disable feedback would then only conditionally show the volume slider? And then the custom sounds would appear in the advanced section (also conditional or not). The tricky bit is that all of these items are tied to the feedback toggle, it doesn't make sense to show them without it?

cjpais:
Okay these are great questions, thank you for them. I probably was making things a bit confusing and overly complicated. Sorry about that. Let's do this if it makes sense:

Have the audio feedback toggle have the same two states (on/off).
When **on** it conditionally shows the following with the UI you have now
* Output Device
* Volume

When **off** it hides the output device/volume entirely.

In regards to the sound which is played. I really would like to have a great default which is not customizable eventually. With this sensibility in mind I would love to have this in the debug menu, so someone can change it if they really want, but it is a setting I would love to eventually get rid of entirely (and replace it with some instructions on how someone can change the sound if they really want to). This can be showed unconditionally in my opinion especially since it's in the debug menu

I know this is a bit complicated still, but hopefully it makes a bit more sense than my previous message. 

> The tricky bit is that all of these items are tied to the feedback toggle, it doesn't make sense to show them without it?

with relation to this, the settings store is a Zustand store so more or less the state is global and accessible from any of the components. It is a bit awkward from a user perspective for now

jacksongoode:
Sounds good, I've moved the UI around a bit now. I can't really think of a way of doing "pairs" of audio. That sounds a little complicated to enforce? Ultimately if there was some user accessible json config where the user could point to the soundfile they want that could also be an option, but this does feel like a small feature.

cjpais:
I've added pairs for now and moved the dropdown to debug (im pushing it there mainly to see how many people want a different sound, and how the defaults can be improved)

There's a chance I will remove custom sounds, I haven't quite decided on that yet

jacksongoode:
Ah, I just saw the "debug" mode, I was mistaken and thought you meant advanced. I can move the custom sounds to the debug mode if that works or I'll let you decide then.

cjpais:
It's all good! I'm super happy to make any other changes and pull it in

Really really appreciate you submitting the pr! It makes my life so easy!!

cjpais:
Just a small note, for the custom sounds I put it in the app resources directory

on my machine that is:

`/Users/cj/Library/Application Support/com.pais.handy/custom_start.wav` and `custom_stop.wav` there as well

PR #224, opened 2025-10-19
Add a cancel hotkey and configuration

jacksongoode:
The cancel action and trigger already exist, but there's no hotkey to trigger them.

It's often useful when you accidentally create a recording that you don't want to paste, and it's a lot easier to press escape than to click on the menu bar to cancel the active transcription.

The one quirk to note is that escape cannot be set as a key stroke. So pressing the reset button is the only way to revert back to escape. 

I think we can just blast through all active toggles and reset `states.active_toggles.values_mut().for_each(|v| *v = false);` simplifying a lot.

### DIFF
--- a/src/components/settings/HandyShortcut.tsx
+++ b/src/components/settings/HandyShortcut.tsx
@@ -15,14 +15,14 @@ import { toast } from "sonner";
 interface HandyShortcutProps {
   descriptionMode?: "inline" | "tooltip";
   grouped?: boolean;
-  shortcutIds: string[];
+  shortcutId: string;
   disabled?: boolean;
 }
 
 export const HandyShortcut: React.FC<HandyShortcutProps> = ({
   descriptionMode = "tooltip",
   grouped = false,
-  shortcutIds,
+  shortcutId,
   disabled = false,
 }) => {
   const { getSetting, updateBinding, resetBinding, isUpdating, isLoading } =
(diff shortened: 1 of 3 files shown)


cjpais:
Skimmed this and I think it looks good to me. I'll take a closer look later this week when I get a chance, thank you as always!

cjpais:
Hmmm.. so for me the UI looks weird and when I hit esc nothing happens. If we are adding another shortcut setting here, that will not be pulled in. It is okay if it's just universally handled as 'esc' but I don't want to add more shortcuts right now see  https://github.com/cjpais/Handy/discussions/211

Screenshot 2025-10-27 at 11 55 20 AM

jacksongoode:
Huh? Maybe the rebase might have changed something? I'll look into it, the state of the PR when I opened had both the transcribe and cancel shortcuts.

jacksongoode:
@cjpais I think all is good, I realized if the settings change, we'll need some migration logic (see the last commit before the rebase). 

Here's how it looks

image

cjpais:
I appreciate you adding this setting, though as stated above since it's adding a new shortcut in the UI it's probably not going to get pulled in the current state

There's a lot of requests for shortcuts at the moment and I don't really want to add any new ones without a UI which handles it all in a nice generic interface. I'm open to UI/UX proposals on that front. 

I will accept this feature if it is entirely hidden from the UI, and we can add the UI changes in later when there is a good UI for it. I'm very open to "iykyk" features as since there are no analytics it really helps me gauge popularity of features and which ones need to be mainlined/have UI around them

I'm not entirely convinced 'cancel' is really even needed since you can just press the shortcut again to stop the transcription and click out of a text field and it's effectively null anyway. But of course my machine is quite powerful so maybe it has more of an effect for those with less powerful machines.

jacksongoode:
I think for the users that don't want the overlay the method to cancel requires two clicks. Would you accept this as default to escape to cancel recording and allow that hotkey to be configured in the debug menu?

cjpais:
I don't think `escape` is a good default since so many applications rely on it, and I wouldn't want to override that.

I really would prefer no UI, but I guess debug is fine

jacksongoode:
The escape key is only handled when there is an active transcription running, but I understand the concern. Other common applications in this area, MacWhisper & superwhisper, use this default as the cancel hotkey.

Just to make the current behavior clear now
- The default to cancel an active transcription is "escape"
- The hotkey is only listened to when there is an active transcription
- We only allow the hotkey configuration when we have push to talk disabled
- The cancel hotkey can be configured from the debug menu

cjpais:
Okay this is good with me I'll test it later today and pull it in, thanks! Appreciate the feedback wrt other apps as well, gives me more confidence in this approach. I am a heavy vim user so I'm curious if this will end up impacting me, will be interesting to see

cjpais:
I know I made a change wrt the init... but I don't see where the binding is only listened to when the transcription is happening. This means it would have to be registered and unregistered, otherwise escape is captured. When I went to test this, it immediately breaks vim because the binding is globally registered. It really needs to be localized only to that function call. 

I did review some of the other changes, and specifically the `HandyShortcut.tsx`. I'm not entirely sure we need a list of id's there. I do see a prop was removed and also some SettingsContainer for the states. I think generally it would be better to leave those be (or use the new SettingsContainer)

```
        <SettingContainer
            key={id}
            title={binding.name}
            description={binding.description}
            descriptionMode="tooltip"
            grouped={grouped}
            disabled={disabled}
            layout="horizontal"
          >
```

jacksongoode:
Ah I see, the changes were made as a result of the HandyShortcut being a singular kind of setting to handle the shortcut config. I'll revert some of this.

cjpais:
No problem thank you!

jacksongoode:
@cjpais I believe this should do what you describe.

The binding is now dynamically registered and deregistered on its own thread. I think this is the cleanest way to handle this, but please let me know. The only UI change was just the HandyShortcut can handle an id so we can use the mapping for both the global and the cancel.

cjpais:
hmm... I tested this and I'm not sure it fully works. 

Right now the debug menu does not render at all for me. Also when hitting esc, I do see logs happening in the console, but the overlay stays up and the handy shortcut seems to get into a funky state as far as I can tell.

jacksongoode:
@cjpais Sorry for lagging on this, just made two fixes (latest commit) that ought to resolve this!

cjpais:
No problem, just appreciate the contribution! I probably won't get time to test this today and I'll be gone for almost a week, I will review and test when I'm back. 

I appreciate all the time you've put in. If you get a chance would you mind giving #322 a spin? I just want a bit more testing done on it, but I think it's getting fairly close. Ideally I would like to merge it before this comes in (and I'm happy to resolve the merge for this after). If you don't get a chance, no worries, just curious

cjpais:
@jacksongoode i merge, made a few changes and tested, i think its good. let me know if you can do a final review otherwise i will be merging it tomorrow

cjpais:
Thanks for all your work on this, I'm sorry it took so long to get merged

jacksongoode:
No worries! Thanks for carrying it forward!

PR #236, opened 2025-10-25
windows paste shift+ins instead of ctrl+v

mvsite:
Shift+Insert is older and more universal than Ctrl+V on windows. For example SSH terminal software often will not permit ctrl+v. This is the minimum POC to get paste working on Windows (critical for cli agents like claude code, codex, droid, auggie, opencode, amp, gemini, qwen, cursor-agent, warp, crush, etc.)

### DIFF
--- a/src-tauri/src/clipboard.rs
+++ b/src-tauri/src/clipboard.rs
@@ -6,14 +6,14 @@ use enigo::Settings;
 use tauri::AppHandle;
 use tauri_plugin_clipboard_manager::ClipboardExt;
 
-/// Sends a paste command (Cmd+V or Ctrl+V) using platform-specific virtual key codes.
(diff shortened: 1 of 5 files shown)


cjpais:
Right now I don't have access to a Windows machine so I'm a bit hesitant to pull this in. 

Besides just the terminal apps that you've described, have you tested this in a wide variety of apps? I'd rather not break any existing support where ctrl+v for some reason works where shift+insert does not. Perhaps one solution is in the debug menu to add an additional option there on Windows for shift insert. So we can keep the existing behavior working as it does, but until we have high levels of confidence, we don't need to ship it to every user.

Do you think it's possible to make the change to add 'shift+ins' as an option in the `paste method` part of the debug menu for now? @mvsite

PR #257, opened 2025-10-28
feat: add “mute while recording” setting

Maicon-Moreira:
This PR introduces a new “Mute while recording” feature that automatically mutes the system’s sound output when audio recording begins and restores the previous volume afterward.

### DIFF
--- a/src/components/settings/AdvancedSettings.tsx
+++ b/src/components/settings/AdvancedSettings.tsx
@@ -6,7 +6,6 @@ import { CustomWords } from "./CustomWords";
 import { SettingsGroup } from "../ui/SettingsGroup";
 import { StartHidden } from "./StartHidden";
 import { AutostartToggle } from "./AutostartToggle";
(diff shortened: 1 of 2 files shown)


cjpais:
I appreciate you going and implementing this! Let's just put the setting in the debug menu for now and I'll pull it in. Mainly doing this to see how many requests I get to have the feature in order to more widely support.

I really like that the changes are minimal, I expected a much larger lift, so thank you again for your time

If you want me to wait to pull the changes in so you can test I would also be happy to do that. Tested stuff is always great!

Maicon-Moreira:
Ah, I just overlooked the debug settings — moved “Mute while recording” there.

Maicon-Moreira:
I’m super low on time right now, so feel free to merge. If any issues pop up on Windows or Linux, we can fix them later. The changes are minimal, and I don’t see anything that could break.

cjpais:
Works great on MacOS, thanks. Merging

PR #259, opened 2025-10-29
fix: disable Translate to English setting for the whisper turbo model

bodyazinchenko:
Hello @cjpais , thank you for such wonderful app!

Based on [README](https://github.com/openai/whisper/blob/main/README.md) from the whisper repo, I found that whisper turbo is not available for translate.
> Note: The turbo model will return the original language even if --task translate is specified. Use medium or large for the best translation results.

So, I decided to improve disabling this settings.

Screenshot 2025-10-29 at 12 35 00

### DIFF
--- a/src/components/settings/LanguageSelector.tsx
+++ b/src/components/settings/LanguageSelector.tsx
@@ -11,6 +11,8 @@ interface LanguageSelectorProps {
   grouped?: boolean;
 }
 
+const unsupportedModels = ["parakeet-tdt-0.6b-v2", "parakeet-tdt-0.6b-v3"];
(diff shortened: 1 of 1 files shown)


cjpais:
Thank you! Would you mind also adding parakeet v2 to this list, I just added it last night and it's an English only model with no translation capability

bodyazinchenko:
> Thank you! Would you mind also adding parakeet v2 to this list, I just added it last night and it's an English only model with no translation capability

Yes, of course, I added it to the list.

cjpais:
thank you!

PR #269, opened 2025-10-30
fix: tooltip overflow

fampiyush:
Fixes the tooltip overflow by adding a prop to SettingContainer. So now we can pass if we want the tooltip to be displayed on top or bottom.

Before: 
Screenshot 2025-10-30 160858

After:
Screenshot 2025-10-30 200242

The Handy Shortcut's tooltip position has been set to bottom, other components default to top.


cjpais:
Thank you! Tested and it works

PR #222, opened 2025-10-18
feat: LLM based post-processing

eduardolat:
I've implemented an option in the debug panel to enable LLM-based post-processing.

This new feature uses OpenRouter as a backend to make it easier to use any model without having to maintain so many implementations in the code.

This is a working proof of concept. I'm not the best at design, but I've tried to make it as simple as possible. The implementation works, and you just need to finalize it wherever you like.

I think it is a good option to be activated in the advanced options or activate it in its own menu, either way you decide

Thanks for this great project.

image

image

PR #279, opened 2025-10-31
feat: add Gemini API integration for transcription

mmabas77:
Adds support for using Gemini API as an alternative to local Whisper/Parakeet models with automatic fallback.

Features:
- New API transcription module with OpenAI-compatible endpoint
- Configurable API key, endpoint, and model selection in Debug settings
- Automatic fallback to local models when API fails
- Full language support for API transcription
- Settings persistence for all API configuration

Backend changes:
- New api_transcription.rs module for Gemini API client
- Updated transcription manager with API routing logic
- Added TranscriptionSource enum (Local/API) to settings
- Four new Tauri commands for API settings persistence
- Skip model loading when using API mode

Frontend changes:
- New TranscriptionSource, ApiKey, ApiModel, ApiEndpoint components
- Updated DebugSettings with API Configuration section
- Enhanced settings store with API updaters
- Type definitions for API settings

image

### DIFF
--- a/src-tauri/src/api_transcription.rs
+++ b/src-tauri/src/api_transcription.rs
@@ -79,6 +79,7 @@ pub async fn transcribe_with_api(
     api_key: &str,
     api_endpoint: &str,
     api_model: &str,
+    api_prompt: &str,
     language: Option<String>,
 ) -> Result<String> {
     if audio.is_empty() {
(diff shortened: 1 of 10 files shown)


cjpais:
Im planning to close this as we are not supporting APIs for transcription currently. They are outside of the scope of the project for now. They may be considered in the future but not at the moment. This is a local transcription app primarily.

API support will only be accepted if it is a setting outside of the main UI and enabled via debug or experimental features for now. It also must support multiple API providers in a nice and generic enough way + transcribe-rs. See #222 for how an implantation may look in the UI

See #77 for more details and previous discussion

mmabas77:
Its currently added as a separate section in debug (with a swich to turn on local models or api) but can be modified to be in a separate view with a toggle btn such as in #222 
Currently it dose support any OpenAI API compatible model but i have added gemini as a drop down selection (can be changed easily too).
So if changes are required iam happy to do it as i like the project and its becoming a must have app on my system.

cjpais:
Please make the changes as requested above. #222 is a good reference for how this may be implemented. 

* transcription-rs has support for openai-api and that support should be used (if not already, I haven't looked deep in the code yet, im moving and have very minimal time)
* clear dropdowns for common providers, and custom, with storage for them as well
* the switch to enable this should be in a debug menu

I'm not sure how you intend to mix the local model UI and provider UI. This must be very clear and obvious. Screenshots will be necessary so I can review quickly and easily before I even sync the code. 

I would wait to make the changes so you can use #222 as a reference, it's not finished yet and will likely have code changes.

cjpais:
Closing this as I don't want to support API's for now and the changes requested were not made. I think this could also be done with much less code.

PR #280, opened 2025-10-31
fix: improve VAD sensitivity for better speech detection

mmabas77:
Improved Voice Activity Detection to handle background noise better:
- Lowered Silero VAD threshold from 0.3 to 0.2 (20% confidence)
- Reduced onset requirement from 2 to 1 frames for faster triggering
- Added debug logging to track speech vs noise frame classification

This fixes the issue where audio was sometimes detected as empty when speaking with background noise present. Testing shows improvement from 12.7% speech detection to much higher rates in noisy environments.

### DIFF
--- a/src-tauri/src/audio_toolkit/audio/recorder.rs
+++ b/src-tauri/src/audio_toolkit/audio/recorder.rs
@@ -237,6 +237,8 @@ fn run_consumer(
 
     let mut processed_samples = Vec::<f32>::new();
     let mut recording = false;
+    let mut speech_frames = 0usize;
+    let mut noise_frames = 0usize;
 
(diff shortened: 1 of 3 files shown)


cjpais:
@mmabas77 I'm okay with this overall, but how did you get the 12.7% number?

Do you have clips of audio I can use to test A, B on? I'm hesitant to pull in changes as I've not had any major issues with the VAD as it is currently and haven't heard of many reports or issues with it being a problem in it's current implementation

cjpais:
Closing this as there has been no response and I haven't received any particular feedback on the VAD. If it's an issue I would love feedback and audio samples where it's causing an issue. Happy to fix of course, but really need good data and I don't want to tweak something that doesn't seem broken

Issue #140, opened 2025-09-28
[BUG] Unable to trigger outside of Handy app


tmikaeld:
## Bug Description
Pressing the shortcut triggers record when handy app is active window, not otherwise.

## System Information
**App Version:**
AppImage - 0.5

**Operating System:**
Suse Leap 16

**CPU:**
i7-13700

**GPU:**
Nvidia 4090

cjpais:
Interesting. This is another Linux/Distro issue where I will need support to fix because I cannot test every combo of distro/display compositor/desktop environment. 

If you are able to compile from source and debug let me know, happy to help in any issues with that and add them to the readme

tmikaeld:
I'm quite certain this is a permissions issue, SUSE can be quite hard to get right with this. I'll debug and let you know!

cjpais:
Thank you! You are awesome!

ncoop57:
Same issue on Linux 6.16.8-arch3-1 with Hyprland

tmikaeld:
The is no rust crate that support wayland and especially not ”insert at cursor” globally, so I’ll close this issue. The official way to solve it are DBus extensions in gnome/KDE shell itself.

merrkry:
@cjpais This is a problem with Wayland protocols. Applications are not allowed to register global shortcuts themselves.

Tauri doesn't support global hotkey on Wayland. [tauri-apps/global-hotkey#28](https://github.com/tauri-apps/global-hotkey/issues/28). However, xdg-desktop-portal has `org.freedesktop.portal.GlobalShortcuts` for apps to register global shortcues. This is already used by chromium. It might be easier using crate `ashpd`.  But I am unsure whether this works outside of KDE/GNOME.

To avoid the Wayland hell, another solution is provide CLI options and let users bind hotkeys themselves in their setting center / compositor config.

merrkry:
@tmikaeld Handy implements input by simulating <C-v> keystroke to trigger clipboard paste... `Enigo` crate has experimental Wayland support but the feature flag isn't enabled yet.

tmikaeld:
@merrkry The problem with any of the clipboard managers is that wayland have no API for inserting at cursor position.

The available workaround is ydotool which needs access to /dev/uinput but that essentially gives all apps access to clipboard, which is not a good idea.

[...]

PR #265, opened 2025-10-30
feat: expose transcription over dbus for wayland support

tbezman:
# Add Wayland/Hyprland Support via D-Bus

## What This Does

Adds support for Wayland/Hyprland! I implemented most of this PR with Claude Sonnet 4.5 and it works consistently in my testing.

## The Problem

I'm on Hyprland (Wayland) and ran into two major issues:
1. **Global keyboard shortcuts don't work** - Wayland handles shortcuts differently than X11
2. **Chrome ignores the transcription** - Chrome/Chromium on Wayland blocks synthetic keyboard input for security, so even when transcription completed, nothing would paste into the browser

### D-Bus Interface

Added a D-Bus service so you can trigger transcription via system commands instead of global shortcuts:

```bash
# Start transcription
busctl --user call com.pais.Handy /com/pais/Handy com.pais.Handy.Transcription StartTranscription

# Stop transcription  
busctl --user call com.pais.Handy /com/pais/Handy com.pais.Handy.Transcription StopTranscription
```

You can bind these to keys in your Hyprland config:
```conf
bind = SUPER, R, exec, busctl --user call com.pais.Handy /com/pais/Handy com.pais.Handy.Transcription StartTranscription
bindrt = SUPER, R, exec, busctl --user call com.pais.Handy /com/pais/Handy com.pais.Handy.Transcription StopTranscription
```

### wtype Integration for Chrome

Chrome blocks synthetic input from libraries like `enigo`, so I added support for `wtype` (a Wayland-native text input tool). When you set the paste method to "Direct Input" in settings, it automatically uses `wtype` on Wayland, which Chrome trusts.

Install with: `sudo pacman -S wtype` (or your distro's equivalent)

### Focus Handling

The app now captures which window has focus before showing the overlay, then restores it before pasting. This ensures the text goes to the right place (using `hyprctl` for Hyprland).

## What Changed

- **New files**: 
  - `src-tauri/src/dbus.rs` - D-Bus interface implementation
  
- **Modified files**:
  - `src-tauri/src/clipboard.rs` - Added wtype support
  - `src-tauri/src/overlay.rs` - Added focus capture/restore
  - `src-tauri/src/actions.rs` - Calls focus restoration before pasting
  - `src-tauri/src/shortcut.rs` - Detects Wayland and disables shortcuts
  - `src-tauri/src/lib.rs` - Initializes D-Bus service
  - `src-tauri/Cargo.toml` - Added zbus dependencies

## Automatic Behavior

The app automatically:
[...]

PR #281, opened 2025-10-31
feat: SIGUSR2 handling for toggles, `wayland` enigo feature

johnpyp:
This PR adds:
- support for sending `SIGUSR2` to handy to toggle recording on and off,
  - This provides a robust way to handle keybinds centrally (in Wayland, for example) like discussed in https://github.com/cjpais/Handy/issues/140, rather than going through the app.
- adds the `wayland` feature to `enigo`
  - this allows enigo to use the [wayland input method protocol](https://wayland.app/protocols/input-method-unstable-v2#compositor-support), which is supported by most wayland compositors.
  - with this feature enabled, enigo will still attempt to write to all its possible input methods, so this won't conflict with X11.

### Example usage of SIGUSR2 (Sway config):

```
bindsym $mod+o exec pkill -USR2 -n handy
```

Now, whenever I press `alt+o`, it'll toggle on, and I press it again, it toggles off. Easy.

(`pkill` is used here but it doesn't actually "kill" anything. It's just sending that signal to the process matching the name "handy", which we handle!)

### Complications

Enabling the `wayland` feature on enigo seems to affect the overlay monitor detection (only inside of wayland). Just using X11 detection actually *works* (at least in Sway), but by using the wayland detection, enigo instead early-errors out with:

```
[2025-10-31T19:06:33Z ERROR enigo::platform::wayland] You tried to get the mouse location. I don't know how this is possible under Wayland. Let me know if there is a new protocol
```

This seems like an upstream bug which needs to be fixed to properly gracefully fallback.

### DIFF
--- a/src-tauri/Cargo.lock
+++ b/src-tauri/Cargo.lock
@@ -1425,6 +1425,10 @@ dependencies = [
  "objc2 0.6.3",
  "objc2-app-kit",
  "objc2-foundation 0.3.2",
+ "tempfile",
+ "wayland-client",
+ "wayland-protocols-misc",
+ "wayland-protocols-wlr",
  "windows 0.61.3",
  "x11rb",
  "xkbcommon",
@@ -2296,6 +2300,7 @@ dependencies = [
  "rustfft",
  "serde",
  "serde_json",
+ "signal-hook",
  "strsim",
  "tar",
  "tauri",
@@ -5562,6 +5567,16 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "0fda2ff0d084019ba4d7c6f371c95d8fd75ce3524c3cb8fb653a3023f6323e64"
 
 [[package]]
+name = "signal-hook"
+version = "0.3.18"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "d881a16cf4426aa584979d30bd82cb33429027e42122b169753d6ef1085ed6e2"
+dependencies = [
+ "libc",
+ "signal-hook-registry",
+]
+
+[[package]]
 name = "signal-hook-registry"
 version = "1.4.6"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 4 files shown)


cjpais:
Hey thank you so much! How much have you tested this? I don't have a Linux machine currently so I won't be able to test but would love to pull this in

johnpyp:
So far I've tested with tauri dev only, on Wayland.

I can try a prod build, and verify that it doesn't regress behavior on X11?

cjpais:
That would be great! And if you do get one could you post it here and also in the discord?

This thread had some stuff for Wayland debug: https://discord.com/channels/1391636091853078608/1432847242766716989

Also if time can you evaluate the impact relative to this PR as well? I want to merge it soon just haven't had time to review the code carefully

https://github.com/cjpais/Handy/pull/265

cjpais:
I'm going to merge for now, but I do suspect we may need to disable the `wayland` feature. I see an issue on the repo where there is double input if multiple methods are available.

cjpais:
Ahhhh. I was a bit optimistic. I reverted, mainly due to stipulations initially. Let's not break anything if possible. It seems possible to add just the SIGUSR2 without adding the Wayland feature. lets do that, I'm sorry you'll probably need to open another PR

johnpyp:
No problem, I'll do that.

OriLiMu:
I dont know how to use it yet.
I open handy and run the command `pkill -USR2 -n handy` from terminal
first time nothing happens, and second time run the command, the handy window is killed.
why?

I download the lastest app

johnpyp:
I believe the PR was reverted, so it's not in the latest release. I'm opening another PR with just these changes shortly

cjpais:
Yep it was reverted

OriLiMu:
> I believe the PR was reverted, so it's not in the latest release. I'm opening another PR with just these changes shortly

hi, have this feature be added or not?

cjpais:
SIGUSR2 has been added you can review how to use in the readme. Enigo feature has not

PR #293, opened 2025-11-03
Add files "patch to add french langage" via upload

aurlro:


cjpais:
uhhh, can you make this not a patch so it can be merged and tested without extra effort?

Issue #312, opened 2025-11-07
In the example prompt "impove transcriptions" add a rule too keep the language in the original form


lstep:
The current prompt for "improve transcriptions", when used while talking in a language other than english, will translate the result to english.

To correct this one can add a rule as follows:

> 5. Keep the language in the original version (if it was french, keep it in french for example)

cjpais:
Could you submit a PR for this please

cjpais:
Closed via #316

PR #316, opened 2025-11-07
Add rule for example prompt to keep the original language

lstep:
Corrects feature request #312 so that the language used is kept after AI corrections.


cjpais:
Thank you! Merged

PR #320, opened 2025-11-09
feat: Enable enigo wayland feature

lhotari:
## Motivation

Currently Handy cannot be used in Wayland compositors

## Changes

Enable enigo's [wayland feature](https://docs.rs/crate/enigo/latest/features).
- [wayland input method protocol is supported](https://wayland.app/protocols/input-method-unstable-v2#compositor-support) in some of the compositors such as Hyprland and Cosmic.

## Tested with PopOS 24.04 Cosmic

This makes Handy usable on at least PopOS 24.04 Cosmic with paste method "Clipboard (CTRL+V)".
image

To enable the global hot key in Cosmic it's necessary to change X11 compatibility settings and set the "Global shortcuts in X11 applications" value to "All keys".

image

### DIFF
--- a/src-tauri/Cargo.lock
+++ b/src-tauri/Cargo.lock
@@ -1451,6 +1451,10 @@ dependencies = [
  "objc2 0.6.3",
  "objc2-app-kit",
  "objc2-foundation 0.3.2",
+ "tempfile",
+ "wayland-client",
+ "wayland-protocols-misc",
+ "wayland-protocols-wlr",
  "windows 0.61.3",
  "x11rb",
  "xkbcommon",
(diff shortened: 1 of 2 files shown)


cjpais:
Another PR had this feature enabled and was ultimately reverted

> I do suspect we may need to disable the wayland feature. I see an issue on the repo where there is double input if multiple methods are available.

The repo I mention here is the enigo repo, and I think double input is not a good thing.

lhotari:
> The repo I mention here is the enigo repo, and I think double input is not a good thing.

thanks. Found the issue enigo-rs/enigo#464.

Looking forward to progress on #265 with the `wtype` integration.

cjpais:
@lhotari if you get a chance would you mind trying that PR? I don't have a Linux machine at the moment so it's gonna take me a bit to get this in unless I can get some confirmation from folks that this does help

I almost might extract the wtype bit out specifically if it does help. Then add the dbus support separately

lhotari:
> @lhotari if you get a chance would you mind trying that PR? I don't have a Linux machine at the moment so it's gonna take me a bit to get this in unless I can get some confirmation from folks that this does help
> 
> I almost might extract the wtype bit out specifically if it does help. Then add the dbus support separately

@cjpais The wtype parts work for me, as long as there's no overlay (since the original window loses focus because of the overlay). I made a PR #372 with the wtype parts extracted from #265.

PR #322, opened 2025-11-10
Add tauri-specta for generating TS bindings

cjpais:
This is part of an ongoing refactor to make the whole settings deal a bit easier to deal with. Making it easier and cleaner to add new settings and reducing the amount of boilerplate needed to do so.

### DIFF
--- a/src/components/model-selector/ModelSelector.tsx
+++ b/src/components/model-selector/ModelSelector.tsx
@@ -277,12 +277,11 @@ const ModelSelector: React.FC<ModelSelectorProps> = ({ onError }) => {
 
   const handleModelSelect = async (modelId: string) => {
     try {
+      setCurrentModelId(modelId); // Set optimistically so loading text shows correct model
       setModelError(null);
       setShowModelDropdown(false);
       const result = await commands.setActiveModel(modelId);
-      if (result.status === "ok") {
-        setCurrentModelId(modelId);
-      } else {
+      if (result.status === "error") {
         const errorMsg = result.error;
         setModelError(errorMsg);
         setModelStatus("error");


src-tauri/src/clipboard.rs
```diff
@@ -40,12 +40,11 @@ fn send_paste_ctrl_v() -> Result<(), String> {
...
 /// Sends a Shift+Insert paste command (Windows and Linux only).
 /// This is more universal for terminal applications and legacy software.
-#[cfg(not(target_os = "macos"))]
 fn send_paste_shift_insert() -> Result<(), String> {
     #[cfg(target_os = "windows")]
     let insert_key_code = Key::Other(0x2D); // VK_INSERT
-    #[cfg(target_os = "linux")]
-    let insert_key_code = Key::Other(0x76); // XK_Insert (keycode 118 / 0x76)
+    #[cfg(not(target_os = "windows"))]
+    let insert_key_code = Key::Other(0x76); // XK_Insert (keycode 118 / 0x76, also used as fallback)
```
jacksongoode, line 47: This is unrelated to the PR here but was this test to be fine on macOS too?

src-tauri/src/lib.rs
```diff
@@ -209,6 +214,90 @@ pub fn run() {
...
+    ]);
+
+    #[cfg(debug_assertions)] // <- Only export on non-release builds
+    builder
+        .export(
+            Typescript::default().bigint(BigIntExportBehavior::String),
+            "../src/bindings.ts",
+        )
+        .expect("Failed to export typescript bindings");
+
```
jacksongoode, line 302: What is this doing exactly? I guess stripping the debug for non-release? I think cargo might already be doing this, unless this is a recommended practice.

cjpais:
One of the main reasons is to avoid having to write boilerplate types which were typical in new features. It's generally an overall move towards less boilerplate if possible and really the first stage of a refactor. Though if you think there is more boilerplate too maybe a conscience wrapper or something else. Perhaps this isn't quite the solution.

Generally settings right now is quite painful I find as dev experience and not generic enough to move up the stack

jacksongoode:
Yeah, I'd say this looks good. I'd go ahead with it pending the changes that might not be related to the bindings.

src-tauri/src/lib.rs
cjpais: This is only generating the bindings themselves when you're in a development environment so that when you ship the application it has stable sets of bindings, I guess. I think it is not explicitly necessary, however it makes enough sense, I think, to have. So I'm keeping it for now, though we can always remove it. Though it'll add some code generation to the CI step, which I guess is what we're more or less trying to avoid.

cjpais:
okay going to yolo this in, everything i've tested so far works fine. for a while there was a regression in settings loading and it seems i resolved it by changing some things around

PR #328, opened 2025-11-11
feat: add fallback microphone for clamshell/desktop mode

jorgedanisc:
Implementation for #326 

Tested for macOS, and is working, would need feedback for Linux and Windows (in theory on these platforms, the Desktop Microphone option should not appear on the settings since this is a macOS only feature)


jorgedanisc:
closing to reopen with proper branch PR from my end

PR #329, opened 2025-11-11
feat: add fallback microphone for clamshell/desktop mode

jorgedanisc:
Implementation for #326

Tested for macOS, and is working, would need feedback for Linux and Windows (in theory on these platforms, the Desktop Microphone option should not appear on the settings since this is a macOS Laptop only feature)

Screenshot 2025-11-11 at 21 34 02

### DIFF
--- a/src-tauri/src/helpers/clamshell.rs
+++ b/src-tauri/src/helpers/clamshell.rs
@@ -26,11 +26,46 @@ pub fn is_clamshell() -> Result<bool, String> {
     Ok(stdout.contains("\"AppleClamshellState\" = Yes"))
 }
 
+/// Checks if the Mac has a built-in display (i.e., is a laptop)
+///
+/// This queries the macOS IORegistry for built-in displays.
+/// Returns true if a built-in display is found (MacBook), false otherwise (Mac Mini, Mac Studio, etc.)
+#[cfg(target_os = "macos")]
+#[tauri::command]
+pub fn has_builtin_display() -> Result<bool, String> {
+    let output = Command::new("ioreg")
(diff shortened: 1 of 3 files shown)


cjpais:
Going to test this when I get a chance (I don't have a monitor currently), but could you move the fallback microphone to debug settings. It should also be labeled more clearly. I don't really understand the UI and what microphone is which. I suspect this is a niche use case and while we can include the code, we will leave it in debug until it is requested a lot.

jorgedanisc:
Ok, moved to Debug and renamed the setting to "Closed Laptop Microphone".

From my experience, having this setting is useful for people like me that have an external monitor and use the laptop closed with it. This way it is possible to use a specific microphone when in clamshell mode

cjpais:
@jorgedanisc I made a few changes, can you verify the PR still works for you as expected. I believe it should I just simplified things a bit and cleaned up some code

jorgedanisc:
It looks good! To clarify, I had a method called "has_builtin_display", this was to end up using for conditional rendering on React as you did with "isMacOs" for example. The reason I chose to check for builtin display is that some apple devices like Mac Mini, Mac Studio and so on are natively desktop, so this setting would be redundant on those cases - it is only effective for laptops

cjpais:
I see, I think those functions were unused in the original PR and not for any condition rendering as far as I can read from the commits. Happy to add them back

jorgedanisc:
Yeah, I must have forgotten to use them in react for the conditional rendering on the PR. If you would like that behaviour (only show the setting actually for macOS laptops) then add it back, otherwise I think it works well the way it is!

cjpais:
Added it back because I do think it's good on the whole, merging

PR #330, opened 2025-11-11
feat: add more options for deleting recordings automatically and added button to open the folder that holds the recordings

jorgedanisc:
## Add recording retention period setting and open recordings folder button

- **Recording Retention**: Added "Delete Recordings" setting in Advanced Settings with time-based and count-based options. Defaults to preserving 5 recordings (the default since we are using `history_limit` here, therefore this value is dynamic to that setting). Backend cleanup logic now uses the selected option from the user preference instead of fixed count.

- **Open Recordings Folder**: Added button in History page header that opens the recordings directory in the system's file manager. Cross-platform implementation using Tauri commands.

These features improve user control over storage management and provide easy access to recorded files.

Here are some screenshots:
Screenshot 2025-11-11 at 19 57 11

Screenshot 2025-11-11 at 19 57 18

### DIFF
--- a/.github/workflows/build-test.yml
+++ b/.github/workflows/build-test.yml
@@ -25,6 +25,9 @@ jobs:
           - platform: "windows-latest"
             args: ""
             target: "x86_64-pc-windows-msvc"
+          - platform: "windows-11-arm" # for ARM64 Windows runner
+            args: "--target aarch64-pc-windows-msvc"
+            target: "aarch64-pc-windows-msvc"
 
     uses: ./.github/workflows/build.yml
     with:
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -77,7 +77,17 @@ jobs:
             echo "Build profile: release"
           fi
 
-      - uses: oven-sh/setup-bun@v2
+      - name: Setup Bun (standard)
+        if: ${{ !(contains(inputs.platform, 'windows') && contains(inputs.target, 'aarch64')) }}
+        uses: oven-sh/setup-bun@v2
+
+      # Bun does not fully support Windows ARM64 yet, so we pin the baseline build.
+      # See https://github.com/oven-sh/bun/issues/9824 for details.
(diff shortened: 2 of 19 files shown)


cjpais:
Going to review this more in some time, but would you mind moving the setting to the debug menu for now? Otherwise it looks well done from the first glance

cjpais:
Skimmed this and generally happy with it to merge with a few minor nitpicks

Wondering if: `update_recording_retention_period` can just take an `RecordingRetentionPeriod` instead of a string. I am planning to add specta bindings in the future which should make this fairly clean with the front end. I believe I can add type generation for the enum and command. This would slim up the code on the Rust side a bit.

There is also some duplicated code between the cleanup functions, we probably should unify that so it just cleans up by [id] or something. That way we can reduce some duplication and make it clear where exactly the boundary between `query to cleanup` and `actual cleanup` is

Would https://tauri.app/plugin/opener/ be useful? I actually kind of like the slimmer implementation you have rather than adding new deps, but mostly worried about compatibility on Linux. I think it would be good to at least have confirmation that the feature works as expected on Windows/MacOS. We can tune Linux as necessary from there. I do suspect opener might be slightly more robust. On my Mac the current implementation works great fwiw. To be clear I am happy to accept either, just curious to hear your viewpoint, and if it's a clear and easy win to use 'opener' we should just go ahead and use it. Generally curious what platforms you've tested this feature on

jorgedanisc:
Regarding Type generation for the enums between rust and ts, I am not too knowledgeable about this, however seems like a good idea to improve code quality if it works.

Whereas the opener plugin looks really nice, I think it should be used, it actually can be used either way, on rust or typescript, so it definitely is the right way to open folder.

I was testing on macos.
Would you implement these fixes or do you want me to do it? I could fix to use the opener plugin

jorgedanisc:
Refactored to use the opener plugin on "Open Recordings Folder"

cjpais:
merging

PR #340, opened 2025-11-15
Enable Windows ARM64 builds in CI

oddrationale:
## Summary
This PR adds native Windows ARM64 (aarch64) support by introducing a new build matrix entry and handling platform-specific tooling limitations in the GitHub Actions workflow.

### Build Matrix
- Added `windows-11-arm` runner configuration to both `build-test.yml` and `release.yml` workflows
- Targets `aarch64-pc-windows-msvc` for native ARM64 Windows builds

### Bun Setup (Windows ARM64)
- Split Bun installation into conditional steps
- **Windows ARM64 uses x64-baseline build**: Bun doesn't fully support Windows ARM64 yet (tracked in [oven-sh/bun#9824](https://github.com/oven-sh/bun/issues/9824)), so we download the baseline x64 build which runs via emulation on ARM64 runners
- Standard Bun setup used for all other platforms
- Frontend dependencies installed with `--cpu=arm64` flag on Windows ARM64 to ensure correct native module selection

### Vulkan SDK (Windows ARM64)
- **Manual Vulkan SDK setup required**: The `humbletim/install-vulkan-sdk@v1.2` action doesn't support Windows ARM64 targets yet (tracked in [humbletim/install-vulkan-sdk#22](https://github.com/humbletim/install-vulkan-sdk/pull/22))
- Custom 4-step process implemented:
  1. Set up `VULKAN_SDK` environment variable
  2. Download ARM64 binaries from LunarG (version 1.4.309.0)
  3. Extract binaries using 7-Zip and add to PATH
  4. Build headers and libraries from source using `humbletim/setup-vulkan-sdk@v1.2.1`
- Standard Vulkan SDK action used for x64 Windows builds

### Rust Target
- Updated target installation condition to exclude `x86_64-pc-windows-msvc` (default target, doesn't need explicit installation)
- ARM64 target (`aarch64-pc-windows-msvc`) installed explicitly via `rustup target add`

### Artifact Paths
- **Updated artifact upload paths for cross-compilation**: Windows x64 builds place artifacts in `target/release/bundle/`, but ARM64 cross-compilation nests them in `target/aarch64-pc-windows-msvc/release/bundle/`
- Added conditional path logic to handle both standard and cross-compiled output locations

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -341,8 +341,7 @@ jobs:
         with:
           name: ${{ inputs.asset-prefix }}-${{ inputs.target }}
           # Default Windows builds place bundles under release/, but cross-compiles (ARM64) nest under target/<triple>/release.
-          # This ternary keeps x86_64 uploads working while pointing cross-builds at their nested bundle directory.
           path: |
-            src-tauri/target/${{ inputs.target != '' && inputs.target != 'x86_64-pc-windows-msvc' ? format('{0}/{1}', inputs.target, steps.build-profile.outputs.profile) : steps.build-profile.outputs.profile }}/bundle/msi/*.msi
-            src-tauri/target/${{ inputs.target != '' && inputs.target != 'x86_64-pc-windows-msvc' ? format('{0}/{1}', inputs.target, steps.build-profile.outputs.profile) : steps.build-profile.outputs.profile }}/bundle/nsis/*.exe
+            src-tauri/target/${{ inputs.target != '' && inputs.target != 'x86_64-pc-windows-msvc' && format('{0}/{1}', inputs.target, steps.build-profile.outputs.profile) || steps.build-profile.outputs.profile }}/bundle/msi/*.msi
+            src-tauri/target/${{ inputs.target != '' && inputs.target != 'x86_64-pc-windows-msvc' && format('{0}/{1}', inputs.target, steps.build-profile.outputs.profile) || steps.build-profile.outputs.profile }}/bundle/nsis/*.exe
           retention-days: 30


oddrationale:
@cjpais, I tested the `build-test.yml` GitHub Action on my branch: https://github.com/oddrationale/Handy/actions/runs/19380145278/job/55456937759

I had to disable signing binaries, but the build succeeds. I installed the resulting `.msi` package on my local machine and Handy is working.

The Parakeet V3 model works great! The Whisper Small model is very slow but still works. Could probably get better performance by using NPU acceleration on the Snapdragon chip. But that is for another time!

cjpais:
Wow thank you for this! I'm just going to go ahead and merge for now and will follow up if a build from CI for testing!

cjpais:
@oddrationale if you have both a x86 and arm machine would you mind testing the builds here? I'm not sure these are accessible to everyone so I will upload copies to the handy blob storage as well

https://github.com/cjpais/Handy/actions/runs/19403353954

Should be available here directly:

ARM .msi: https://blob.handy.computer/win-arm-ci/Handy_0.6.0_arm64_en-US.msi
ARM .exe: https://blob.handy.computer/win-arm-ci/Handy_0.6.0_arm64-setup.exe
x86 .msi: https://blob.handy.computer/win-arm-ci/Handy_0.6.0_x64_en-US.msi
x86 .exe: https://blob.handy.computer/win-arm-ci/Handy_0.6.0_x64-setup.exe

I was able to use the arm exe on emulated windows 11 on my Mac, it worked flawlessly for parakeet. Whisper immediately crashed, though I'm sure Vulkan emulated is not the best. Curious what your results are on the actual hardware

oddrationale:
@cjpais, I only had the chance to test the ARM version.

Parakeet V3 ran smoothly and was quite responsive.

I did encounter a crash when switching to Whisper Small, but after restarting Handy and trying again, it loaded without issues. However, the performance of Whisper Small is very slow, taking 30 seconds to a minute to process even a very short transcription.

cjpais:
Sweet well thank you for all the work on this, I'm just happy we have a build!

talynone:
Bun is now available for Windows ARM, the GitHub step has also been recently updated:

https://github.com/oven-sh/bun/issues/9824#issuecomment-3969572326

Maybe time for another PR that grabs the native bun binary?

oddrationale:
I'm [working on it](https://github.com/oddrationale/Handy/tree/fix/simplify-bun-setup-windows-arm64)!

Waiting on one more missing piece before raising a PR: the [oven-sh/setup-bun](https://github.com/oven-sh/setup-bun) GitHub Action also needs to be updated because it was not aware of the new `aarch64` build. I've already submitted the PR ([#165](https://github.com/oven-sh/setup-bun/pull/165)) for that and it is merged into `main`. The maintainer just needs to update the `v2` tag to point to the `HEAD` of the `main` branch.

Technically, we could point the GitHub Action to the `main` branch. But I'd rather use a specific tagged release.

Issue #336, opened 2025-11-13
[BUG] Audio Feedback is also muted when starting


adofse:
Fist, thank you for creating such nice app! This is what I've been looking for as the alternative app to superwhisper!

## Bug Description
When Mute While Recording and Audio Feedback are both enabled, the feedback sound is also muted immediately when start recording. 

|Audio Feedback is ON|Mute While Recording is ON|
|-|-|
|<img width="300" height="570" alt="Image" src="https://github.com/user-attachments/assets/2395cc1e-6af6-4f27-a608-68605084d546" />|<img width="300" height="570" alt="Image" src="https://github.com/user-attachments/assets/614ae837-3b70-4413-a4cc-f1b0d13b3ca0" />|

## Expected Behavior
System audio should be muted after the feedback sound plays when start recording. 

## How to fix
To fix this, ideally app should get the duration of feedback sound and mute after that duration. Alternatively, app should mute after the fixed amount time, let's say ~2.0 sec.

## System Information

**App Version:**
- v0.5.5 ( which includes https://github.com/cjpais/Handy/pull/257 thanks to @Maicon-Moreira )
- v0.6.0

**Operating System:**
- macOS 15

**CPU:**
- Apple Silicon

**GPU:**
- Apple Silicon

cjpais:
Yeah I've actually experienced this as well and I agree it should be fixed

mihilmy:
This might be related: my audio volume drops to zero after the transcription completes.

cjpais:
@mihilmy can you detail this further? I'm not sure this is the same issue. Could you open another one and let's go from there

Maicon-Moreira:
I’m unsure about the exact sequencing: does the actual recording start only after the audio-feedback sound finishes?

Two important cases:

- If recording starts immediately when the user presses record and we delay muting until the feedback finishes, then system audio (music, notifications, etc.) will leak into the start of the recording during that short window. That’s what I’m worried about.

- If recording only begins after the feedback sound finishes, then the current mute logic is placed in the wrong spot — the mute should be triggered at the real recording-start event (i.e., where recording actually begins), not at feedback play time.

cjpais:
I believe #341 fixed this, @adofse let me know if the latest version of Handy works. I'm going to close it preemptively since it does work flawlessly on my machine. If it doesn't on yours please let me know

adofse:
Yes, thank you!

PR #341, opened 2025-11-16
Fix unstable mute implementation on Windows/Linux

Maicon-Moreira:
As discussed in PR #257 the mute-while-recording feature relied on the cpvc crate to mute/unmute and manage volume levels. While this worked reliably on macOS, it proved unstable on Windows and Linux due to poor integration between cpvc and Tauri.

This PR replaces that approach with a more explicit, platform-specific mute implementation. The new logic works consistently across all platforms tested:

- Ubuntu 24.04.2
- macOS Tahoe 26.0.1
- Windows 11 Pro 25H2

Additionally, cpvc could crash the entire application when mute/volume operations failed. The new set_mute implementation instead fails silently when the OS doesn’t support mute control, preventing full-app crashes.

In the future we may improve this by detecting unsupported systems, showing a user-facing warning, or disabling the “mute while recording” toggle when mute APIs aren’t available.

### DIFF
--- a/src-tauri/src/actions.rs
+++ b/src-tauri/src/actions.rs
@@ -181,13 +181,21 @@ impl ShortcutAction for TranscribeAction {
         debug!("Microphone mode - always_on: {}", is_always_on);
 
         if is_always_on {
-            // Always-on mode: Play audio feedback immediately
+            // Always-on mode: Play audio feedback immediately, then apply mute after sound finishes
             debug!("Always-on mode: Playing audio feedback immediately");
             play_feedback_sound(app, SoundType::Start);
+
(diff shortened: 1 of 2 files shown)


cjpais:
@Maicon-Moreira to be clear you've tested on all the platforms you specified? Just want to make sure.

Maicon-Moreira:
Yup, tested on all of them.
I mean, I tested set_mute on its own in a small Rust project and also inside Handy on all three platforms, just to make sure nothing conflicted — especially the Windows crate with Tauri.

cjpais:
Tested on my Mac, added a little code to fix #336 

Hopefully your fix will address #344 

Pulling it in

PR #348, opened 2025-11-17
fix: update laptop detection logic to be more reliable

jorgedanisc:
The previous implementation checked for a built-in display to detect clamshell mode, but this doesn't work reliably when the laptop lid is closed from my experience now using the released version of the app. This PR updates the logic to check for a battery instead, which provides a more reliable way to identify whether the device is a laptop.


cjpais:
works for me

PR #349, opened 2025-11-17
Improve audio feedback timing using real sound duration and trimmed WAV files

Maicon-Moreira:
This PR builds on the very recent updates from PR #341. That PR introduced the 500ms
mute delay, which worked fine, but since it just landed, this improvement could have
fit naturally there as well. The idea is similar, but this version switches to a
dynamic approach that reads the actual duration of each sound file before muting,
making the behavior more accurate and future-proof as sound themes evolve.

I also found that `pop_start.wav` and `pop_stop.wav` had almost a full second of silence
at the end, which threw off the mute timing. I trimmed that silence using:

```
ffmpeg -i pop_start.wav -af "silenceremove=stop_periods=1:stop_duration=0.1:stop_threshold=-50dB" temp.wav -y
mv temp.wav pop_start.wav

ffmpeg -i pop_stop.wav -af "silenceremove=stop_periods=1:stop_duration=0.1:stop_threshold=-50dB" temp.wav -y
mv temp.wav pop_stop.wav
```

Tested on macOS and everything works smoothly.

### DIFF
--- a/src-tauri/src/actions.rs
+++ b/src-tauri/src/actions.rs
@@ -1,4 +1,4 @@
-use crate::audio_feedback::{play_feedback_sound, SoundType};
+use crate::audio_feedback::{play_feedback_sound, play_feedback_sound_blocking, SoundType};
 use crate::managers::audio::AudioRecordingManager;
 use crate::managers::history::HistoryManager;
 use crate::managers::transcription::TranscriptionManager;
@@ -188,7 +188,7 @@ impl ShortcutAction for TranscribeAction {
             // The blocking helper exits immediately if audio feedback is disabled,
             // so we can always reuse this thread to ensure mute happens right after playback.
             std::thread::spawn(move || {
-                play_feedback_sound(&app_clone, SoundType::Start, true, false);
+                play_feedback_sound_blocking(&app_clone, SoundType::Start);
                 rm_clone.apply_mute();
             });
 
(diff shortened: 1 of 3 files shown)


cjpais:
So I largely rewrote the logic for this feature. Mainly I did it because I felt like there was a simpler path to just make sure we block until the sound is done playing and then we apply the mute. It seems like the simplest and most straightforward option rather than doing this thread spawning stuff and trying to measure the time. I'm curious what your thoughts are. Please go ahead and review all of the code that I have submitted and see what you think of it and if this is a reasonable way forward or if you have other ideas.

Maicon-Moreira:
Love your approach — much simpler and it works great in my tests too.
I added a bit of refactoring in the last three commits to simplify things even more.
Feel free to take a look!

cjpais:
I had a bit of my preferred style in there, I like the more verbose function names at the expense of some modularity. Mainly to improve readability. It might get refactored in the future but it's okay for now.

Maicon-Moreira:
@cjpais works great too!

just a tiny detail from my refactor: I had removed the  
```ts
// Wait before playing stop sound
await new Promise((resolve) => setTimeout(resolve, 800));
```

from **SoundPicker.tsx** because I made `playTestSound` blocking on the JS side (commit `49d77b5f458b478b1baa9272adc04836bdb7380d`).

on your latest commit, `playTestSound` became async again — which makes both sounds play at the same time when clicking **"Preview sound theme"**.

we can fix this by either:

* adding the timeout back in TSX (not ideal), or
* making `playTestSound` call `play_sound_blocking` instead of `play_sound_async`.

cjpais:
Ha thanks for this catch, I suspect it should be blocking and run in a thread, but would have to check the code to be sure. If you have a good idea for the fix I'll be happy to read the code and pull it in

PR #352, opened 2025-11-18
Allow users to deactivate save to history functionality

plafue:
This pull request allows users to deactivate saving transcriptions and recordings to history.
image

Looking forward to your feedback.

### DIFF
--- a/src-tauri/src/actions.rs
+++ b/src-tauri/src/actions.rs
@@ -292,22 +292,26 @@ impl ShortcutAction for TranscribeAction {
                                 }
                             }
 
-                            // Save to history with post-processed text and prompt
-                            let hm_clone = Arc::clone(&hm);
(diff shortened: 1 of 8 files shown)


cjpais:
This can already be done in the debug menu (cmd/ctrl+shift+d) by setting the limit to 0.

plafue:
I see! Would you consider making this option easier to find for people that don't know what debugging is, or people that don't know a debug menu exists?

cjpais:
It has been intentionally put there to generate requests to get an understanding of how many people are using the feature. Right now I don't like the UI/UX on this feature enough to move it.

plafue:
I understand. I'll keep an eye on it in case it changes. 
A couple of non-technical people I recommended Handy to were looking for this.

Thank you for your replies!

cjpais:
Thanks that's super helpful feedback

PR #354, opened 2025-11-18
feat: SIGUSR2 handling for toggles

johnpyp:
(Ported just signal logic from https://github.com/cjpais/Handy/pull/281)

This PR adds:
- support for sending `SIGUSR2` to handy to toggle recording on and off,
  - This provides a robust way to handle keybinds centrally (in Wayland, for example) like discussed in https://github.com/cjpais/Handy/issues/140, rather than going through the app.

### Example usage of SIGUSR2 (Sway config):

```
bindsym $mod+o exec pkill -USR2 -n handy
```

Now, whenever I press `alt+o`, it'll toggle on, and I press it again, it toggles off. Easy.

(`pkill` is used here but it doesn't actually "kill" anything. It's just sending that signal to the process matching the name "handy", which we handle!)

### DIFF
--- a/src-tauri/src/helpers/clamshell.rs
+++ b/src-tauri/src/helpers/clamshell.rs
@@ -38,9 +38,9 @@ pub fn is_laptop() -> Result<bool, String> {
         .arg("batt")
         .output()
         .map_err(|e| e.to_string())?;
-    
+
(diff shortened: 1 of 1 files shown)


cjpais:
Works great on my Mac when signaling via SIGUSR2, thank you for this!

PR #355, opened 2025-11-18
feat: add optional hotkey for post-processing request

sammcj:
This PR adds support for optionally setting a dedicated hotkey which triggers post-processing after a request.

This is useful if you don't usually want post processing, but still want to make use of it for specific request.

- Linting / formatter run before raising PR
- Builds locally and runs without issue

Note: I have no Rust experience, so if the code quality or implementation is not up to standard or does not align with your direction I totally understand if you wish to reject the PR.

image

image


sammcj:
@jacksongoode I note you added a confused reaction, is there something I should clarify?

jacksongoode:
Ah, I was just a little concerned around the complexity and use of a kind of hotkey? Is it like a modifier? Or a separate hotkey that is pressed *after* the generic hotkey is pressed. And also how does this interact with the push to talk setting?

sammcj:
Thanks @jacksongoode

As well as the existing standard hotkey, users can configure a second (optional) hotkey (e.g., Ctrl+Shift+Space) which applies post-processing regardless of global setting

The post-processing hotkey behaves identically to the standard transcribe hotkey with respect to push-to-talk

It's implemented as it's own hotkey as I figured this was simple and explicit and does not alter any existing behaviour while allowing users to configure a similar, or completely different combination.

I'll fix up those merge conflicts from the drift shortly and add a screenshot.

sammcj:
Rebased and updated PR with screenshots

cjpais:
wow i hate new hotkeys normally, but since this is already under the debug flag i am kind of okay with it. i may merge this

cjpais:
realizing not sure what the ux expection is (currently it is everything is post processed, this would mean you would only post process with this shortcut) im not really sure what the right behavior is. this might need to wait for a broader change to define shortcuts, where we can easily control the flow of what happens when pressing a shortcut

cjpais:
Hey @sammcj can you rebase this code and remove package-lock from the PR

sammcj:
@cjpais done :)

cjpais:
ok I promise I will merge this soon

sammcj:
Thanks @cjpais , I'm finding it very useful!

femaphore:
Hi! @cjpais and @sammcj -- This doesn't appear to actually be "optional". There is currently, from what I can tell anyway, no option to unset this hotkey, and the usage of this specific hotkey seems to be the only way that postprocessing is actually triggered. If I want to have postprocessing enabled for every usage, I have to set my main hotkey to a random binding (as it cannot be unset), and set the Post-processing hotkey to the binding I actually want to use. Can this be looked into? Thanks! I might be able to put out a PR for this when I have time myself as well.

PR #356, opened 2025-11-19
feat: add conversion between Simplified and Traditional Chinese.

figwood:
Chinese text is randomly outputted in either Simplified or Traditional characters after recognition, which creates difficulties for subsequent post-processing.

This fork includes the following modifications:

1. The language selection entries in the user interface have been modified: "Chinese" was removed, and "Simplified Chinese" and "Traditional Chinese" were added.

2. Post-processing for recognition has been added: If the user selects "Simplified Chinese" or "Traditional Chinese" as the language, the recognition result will be further processed to convert it to the selected character set.

3. The ferrous-opencc library is used for the Chinese Simplified/Traditional conversion. 

This fork has been successfully tested on Mac M1.

image

### DIFF
--- a/src-tauri/Cargo.lock
+++ b/src-tauri/Cargo.lock
@@ -110,6 +110,56 @@ dependencies = [
 ]
 
 [[package]]
+name = "anstream"
+version = "0.6.21"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "43d5b281e737544384e969a5ccad3f1cdd24b48086a0fc1b2a5262a26b8f4f4a"
+dependencies = [
+ "anstyle",
+ "anstyle-parse",
+ "anstyle-query",
+ "anstyle-wincon",
+ "colorchoice",
+ "is_terminal_polyfill",
+ "utf8parse",
+]
+
+[[package]]
+name = "anstyle"
+version = "1.0.13"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "5192cca8006f1fd4f7237516f40fa183bb07f8fbdfedaa0036de5ea9b0b45e78"
+
+[[package]]
+name = "anstyle-parse"
+version = "0.2.7"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "4e7644824f0aa2c7b9384579234ef10eb7efb6a0deb83f9630a49594dd9c15c2"
+dependencies = [
+ "utf8parse",
+]
+
+[[package]]
+name = "anstyle-query"
+version = "1.1.5"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "40c48f72fd53cd289104fc64099abca73db4166ad86ea0b4341abe65af83dadc"
+dependencies = [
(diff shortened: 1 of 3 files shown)


cjpais:
@figwood going to review the code shortly, but you've confirmed it works well for you?

I don't speak or read any Chinese (yet!) so I won't be able to test this

Edit: on brief review of the code it looks good to me

Is it correct you need OpenCC installed and the Rust lib links to it? Or OpenCC is compiled directly from the library and is part of the binary?

I think it's the former but just want to check as the docs are somewhat ambiguous to me. It would be better to have the app ship with it instead of dynamically link but this is fine for now

cjpais:
@figwood what was the reason for closing?

figwood:
It is hard for packaging with opencc, I've changed it to ferrous-opencc and reopen it.

figwood:
> @figwood going to review the code shortly, but you've confirmed it works well for you?
> 
> I don't speak or read any Chinese (yet!) so I won't be able to test this
> 
> Edit: on brief review of the code it looks good to me
> 
> Is it correct you need OpenCC installed and the Rust lib links to it? Or OpenCC is compiled directly from the library and is part of the binary?
> 
> I think it's the former but just want to check as the docs are somewhat ambiguous to me. It would be better to have the app ship with it instead of dynamically link but this is fine for now

I need to install OpenCC on my Mac for development, then I should copy the dynamic library files when packaging, or the app will be panic. I think it is difficult to deal with different plaforms, so I changed to ferrous-opencc, which is a pure rust lib.

figwood:
Also, this is my first time submitting a PR on GitHub. I initially closed it immediately after realizing there was a packaging issue. Seeing the 'Reopen' button, I decided to reuse the existing PR rather than opening a new one, thanks for your reply!

cjpais:
Wow thank you for finding that lib! Even better. Thank you, merging

xiaokyo:
how to use this version with simplified chinese option?

cjpais:
@xiaokyo use one of the whisper models and then you should be able to change the language on the main screen

xiaokyo:
> whisper

Thanks, worked for me

cjpais:
Awesome, I really would like the UI to document this better in the future. I don't know how to do that quite yet

figwood:
> Awesome, I really would like the UI to document this better in the future. I don't know how to do that quite yet

Do you mean adding notes about Simplified and Traditional Chinese items on UI?

cjpais:
Yes! Also what languages what models support. + Generally explaining somewhere why the language selector is disabled when you use Parakeet, I feel like it's not very obvious right now. Not sure exactly the best way to do this, but I think it requires rearranging the UI a bit.

figwood:
Linguistically, Simplified and Traditional Chinese are the same—they only differ in writing style (think of it like English uppercase/lowercase).

The current process handles this fine: Voice → Model → Random Chinese Output → Ferrous-OpenCC → Final S/T Chinese (based on UI setting). So, the only key requirement is that the model needs to support Chinese.

Since Parakeet seems to be English-only, we should definitely disable the language selector for Chinese. Hope this can help you to rearrange UI.

PR #364, opened 2025-11-20
feat: Add disabled option for pasting.

olejsc:
This adds a "Disabled" option to the "Paste method".  Ref https://github.com/cjpais/Handy/discussions/351#discussioncomment-15009989

The idea is that some users migth drift to other activities while the transcribing is happening and dont have the input field active/selected anymore - leading to frustration when the transcribed text is suddenly copy pasted into the wrong place. 

The natural option to handle this is to *only* copy it to the clipboard, and disable automatic pasting. 

Images for reference:
image

image

Disclaimer: AI was used to assist me with creating this PR and change, as I have little to no knowledge about rust. I have however ran it locally and tested, and the code changes are quite small and targeted. 

I couldnt find a test project, so there is unfortunately no tests provided.


cjpais:
This looks great, thank you. Don't mind AI PRs at all as long as they are reasonable to review! Big long slop is not acceptable but this is excellent!

I've reviewed it but away from a computer so won't pull it in quite yet as I think there will be a merge conflict with some upcoming changes. I'll merge it when I'm back at a computer

Thank you!!!

dannysmith:
I would recommend changing "Disabled" to something like "None (copy to clipboard only)". The expected behaviour for dictation apps like this is that the spoken text is inserted under the cursor. The fact that we use the clipboard to do this is an implementation detail which many users aren't gonna consider, and for them "Paste Method: Disabled" is likely to be confusing language?

kamilkrzyskow:
If there is any confusion then there is the "Show Help" description, and any short form value like `Disabled/None/Skip/No Paste/Without Pasting` could lead to confusion, as to why this is a "Paste Method".

However, +1 from me to simply use `None`, since the "Overlay Position" option already uses `None`, so these 2 options would be matching, and I feel like overall `None` is used more frequently in various apps to select an active "Does Nothing" state.

cjpais:
good enough for me, merging

PR #367, opened 2025-11-21
Use dotool to send paste action if on Wayland

alabhyajindal:
Discussed in #363

Adds a conditional branch in the two paste actions, using dotool to send the paste action if the user is on Wayland. Tested on my machine, KDE Plasma 6.5.2, and it works. Although I need to click back on the text input after triggering Handy since it steals the focus.

This change assumes that dotool is installed on the user's system. This needs to be communicated somehow, I believe this is the first external dependency on Handy, since everything else has been packaged nicely.

### DIFF
--- a/src-tauri/src/clipboard.rs
+++ b/src-tauri/src/clipboard.rs
@@ -1,15 +1,32 @@
 use crate::settings::{get_settings, ClipboardHandling, PasteMethod};
+use crate::utils::is_wayland;
 use enigo::Enigo;
 use enigo::Key;
 use enigo::Keyboard;
 use enigo::Settings;
 use log::info;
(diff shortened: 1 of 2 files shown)


lhotari:
It would be great to have support for `wtype` (https://github.com/atx/wtype) too. Ubuntu/PopOS has a package for it available. dotool isn't available directly. I think `wtype` should be preferred over `dotool`.
`wtype` is used in this draft PR: #265

lhotari:
I extracted wtype parts from #265 to #372.

lhotari:
I extracted some parts from this PR to add support for both `wtype` and `dotool` in #376

PR #369, opened 2025-11-21
feat: windows single left click on tray icon opens handy window

olejsc:
Adds double click support to open application from system tray. 

Tested locally on Windows
*NOT* tested on MacOS. Code for that is experimental.  Would be neat if someone on MacOS could verify that it works. 
Ref. discussion #368 

Simple GIF demo:
demo-double-click-handy-tray-icon

I dont know if this is desired behaviour, as the auther hasn't expressed any opinions on this. Feel free to reject if this is out of scope for the application 😊 
Disclaimer; the code changes was made partially with AI support.


ThioJoe:
Hm is the manual click tracking necessary considering Tauri has the built in double click event?

olejsc:
> Hm is the manual click tracking necessary considering Tauri has the built in double click event?

I tried with the double click event, but it fired for a single click then for me 🤔

dannysmith:
This doesn't seem to change any behaviour on macOS, which is probably correct because I wouldn't expect a menubar item to open an app's settings window. The current behaviour (opening a menu) is pretty normal for this kinda app.

ThioJoe:
Yea for macOS I don't think apps usually have a double click event in the top tray thing. It's mostly a windows thing but pretty common.

> I tried with the double click event, but it fired for a single click then for me 🤔

Do you mean the menu shows with a single left click, or it made it fire the double-click app just when left clicking once?

@olejsc If the former I bet it is because of this line: 
https://github.com/cjpais/Handy/pull/369/files#diff-eabcebd0ae5c9b77a5247e1bdbdb88b1869fc0c932d0ef0cf2ecf9ee5221be7cL157

```
.show_menu_on_left_click(true)
```

Even if it's the latter it might be interfering with the event somehow. Probably better to just set that to false in any case, if the double left click event is added. I assume right click will still show the menu.

cjpais:
@olejsc would be nice if you could try @ThioJoe suggestion, i think it would be preferable

olejsc:
> @olejsc would be nice if you could try @ThioJoe suggestion, i think it would be preferable

Will do when i have some time. Hopefully this weekend.

cjpais:
@olejsc any chance you could look at this? I would like to merge and clear this PR out

koloved:
It would be cool to see this PR in the main , because on Windows it is very strange to click with the left button and get a menu instead of seeing an open program.

VirenMohindra:
▎ hey, i tested this on macOS (arm64, macOS 26.2) with some automated tray interaction tests~

with current code (`show_menu_on_left_click(true)`)
- single click -> menu appears
- double click -> window does **NOT** open ❌
- right click -> menu appears
- rapid clicks -> no crash

with `show_menu_on_left_click(false)`
- single click -> nothing (no menu) ⚠️
- double click -> window opens
- right click -> menu appears
- rapid clicks -> no crash

so the core issue is exactly what @ThioJoe called out... `show_menu_on_left_click(true)` causes macOS to enter a modal menu tracking loop on the first click, which swallows the mouse-up event. the second click never reaches the tray handler, so double-click can't fire

also worth noting: `TrayIconEvent::DoubleClick` is documented as windows-only in the tray-icon crate. the macOS side never emits it. so the DoubleClick match arm on line 182 works on windows but is **dead code** on macOS. the manual `LAST_TRAY_CLICK` timing fallback is what's actually meant to cover macOS but it can't work while the menu is **stealing clicks**

couple of key issues to resolve before merge~
1. show_menu_on_left_click needs to be false — this is the blocker. with it set to true, double-click is impossible on macOS. setting it to false means left-click no longer shows the menu, but right-click still does. this is standard macOS menu bar convention (docker, bartender, rectangle all work this way)
2. the `LAST_TRAY_CLICK` / `DOUBLE_CLICK_THRESHOLD_MS` manual tracking might be unnecessary if we're OK with macOS not having double-click-to-open (since the built-in DoubleClick is windows-only anyway), we can remove the mutex and manual timing code entirely. simpler code, same behavior. alternatively, if we want double-click on macOS too, the manual tracking does work once `show_menu_on_left_click` is false, but then we can't see the tray, which isn't ideal
3. the PR is based on an old branch — it'll need a rebase onto current main

happy to open a follow-up PR with these fixes if @olejsc is unavailable — the changes are small once direction is decided

olejsc:
@VirenMohindra  @cjpais  apologies, I planned to do this but forgot completely about it.  Please @VirenMohindra feel free to tweak & adjust the PR if you have the time, I'm currently quite busy so I dont have the time to look at it. Excellent job testing ( I dont own a Mac).

cjpais:
Okay, I'm finally merging this. I've changed the behavior so it's a single left click opens the handy window and right click will open the menu

PR #372, opened 2025-11-22
feat: Add wayland paste support using wtype

lhotari:
I went ahead and picked the [wtype](https://github.com/atx/wtype) support changes from #265 by @tbezman. This works well with Cosmic DE in PopOS 24.04 when "Overlay Position" is set to "None" and "Paste method" is set to "Direct". I'm using `pkill -USR2 -n handy` in a keyboard shortcut to toggle recording.

### DIFF
--- a/src-tauri/src/clipboard.rs
+++ b/src-tauri/src/clipboard.rs
@@ -4,6 +4,7 @@ use enigo::Key;
 use enigo::Keyboard;
 use enigo::Settings;
 use log::info;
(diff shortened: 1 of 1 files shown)


lhotari:
I tested this and it's not very reliable. The test appearing in the application is sometimes different, although the Handy logs show that the text was transcribed properly. An approach similar as #367 would be better, but using wtype instead of dotool.

PR #375, opened 2025-11-23
Async play test sound

Maicon-Moreira:
@cjpais related to https://github.com/cjpais/Handy/pull/349#issuecomment-3563293866
I might be missing something, but the fix seems pretty straightforward.

These two small changes make playTestSound async on the frontend without causing any unexpected blocking on either the UI or the Tauri backend. Let me know if I’m overlooking anything.


Maicon-Moreira:
By the way, I tested that nothing is blocking on either side:
- Frontend: while the sound is playing, I can move the mouse over UI elements and their hover animations still update normally, so the UI thread isn’t being blocked.
- Backend: clicking the play button repeatedly triggers multiple sounds in parallel, which shows that playback is happening on separate threads and the command itself isn’t blocking the backend.

cjpais:
good with me, thanks

PR #376, opened 2025-11-23
feat: Add wayland paste support using wtype or dotool

lhotari:
This PR is an improvement over #372 which used `wtype` changes from #265. That solution wasn't reliable, presumably due to sending key events for every character in the content. This PR uses the clipboard instead. There's support for 2 command line tools to send the CTRL-V or Shift-Insert key events: `wtype` (https://github.com/atx/wtype) and `dotool` (https://sr.ht/~geb/dotool/). I picked some changes from the `dotool` related PR #367 and made some further improvements.

I've tested this with PopOS 24.04 Cosmic DE. It works well when "Overlay Position" is set to "None" and "Paste method" is set to "Clipboard (CTRL+V)". I'm using `pkill -USR2 -n handy` in a keyboard shortcut to toggle recording.

### DIFF
--- a/src-tauri/src/clipboard.rs
+++ b/src-tauri/src/clipboard.rs
@@ -1,15 +1,22 @@
 use crate::settings::{get_settings, ClipboardHandling, PasteMethod};
+use crate::utils::is_wayland;
 use enigo::Enigo;
 use enigo::Key;
 use enigo::Keyboard;
 use enigo::Settings;
 use log::info;
+use std::process::Command;
 use tauri::AppHandle;
 use tauri_plugin_clipboard_manager::ClipboardExt;
 
 /// Sends a Ctrl+V or Cmd+V paste command using platform-specific virtual key codes.
 /// This ensures the paste works regardless of keyboard layout (e.g., Russian, AZERTY, DVORAK).
 fn send_paste_ctrl_v() -> Result<(), String> {
+    #[cfg(target_os = "linux")]
+    if try_wayland_send_paste(&PasteMethod::CtrlV)? {
+        return Ok(());
+    }
+
     // Platform-specific key definitions
     #[cfg(target_os = "macos")]
     let (modifier_key, v_key_code) = (Key::Meta, Key::Other(9));
@@ -42,6 +49,11 @@ fn send_paste_ctrl_v() -> Result<(), String> {
 /// This is more universal for terminal applications and legacy software.
 #[cfg(not(target_os = "macos"))]
 fn send_paste_shift_insert() -> Result<(), String> {
+    #[cfg(target_os = "linux")]
+    if try_wayland_send_paste(&PasteMethod::ShiftInsert)? {
+        return Ok(());
+    }
+
     #[cfg(target_os = "windows")]
     let insert_key_code = Key::Other(0x2D); // VK_INSERT
     #[cfg(target_os = "linux")]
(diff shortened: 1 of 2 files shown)


alabhyajindal:
Thanks! Tested it on my machine, Fedora Linux, KDE Plasma 6.5.3, with both paste methods and it works.

I agree that using either of `wtype` or `dtool` is better for increasing the chance of not requiring a third party installation. However, from a release perspective, we should pick one and ask the users to install it if they're running Wayland. My preference would be `dotool` because it has recent commits.

I'm going to close my other PR so we can review this one.

lhotari:
> I agree that using either of `wtype` or `dtool` is better for increasing the chance of not requiring a third party installation. However, from a release perspective, we should pick one and ask the users to install it if they're running Wayland. My preference would be `dotool` because it has recent commits.

Onf of the differences between `wtype` and `dotool` is that they work in a different way. `wtype` works at the Wayland protocol level and doesn't require any device access. `dotool` uses the kernel `uinput` module, which requires special config. I think it would be useful to have support for multiple ways. However the way to achieve this could be different. Implementing the logic in a shell script could be more practical. For that purpose a "hook script" in Handy could be better so that anyone could tweak the logic to match their environment. For example, Wayland doesn't currently have a portable way to save the window which has focus and restore focus to this window later. PR #265 contains a way to handle this for Hyprland, but that's just one of the many compositors using Wayland.

cjpais:
Good with this, merging it. I won't be able to test, but will do anything to support the wayland folks!

@lhotari one question, wondering if you think for onboarding there could be a step added for wayland or something? where it checks some of the availability and guides through some install of wtype or dotool? maybe even similar for enabling the shortcuts with the SIGUSR2??? I think generally the onboarding flow could be worked on a bit and maybe this is one area to smooth things out a bit too? 

This is a bit of a tangent now, but just thinking out loud of this is a good enough place to discuss it, which is I'm generally wondering whether it makes sense for the user to kinda set up some of the more basic or general settings um along with recommendations like you know, for Mac OS using the built-in MacBook microphone, I think is a much better default, for example. Um and I would like to recommend that to users on Mac OS. Similarly for things like Wayland to be able to give some guidance or even help text on what to do on your system. And generally just showing people maybe how to change keyboard shortcuts or guiding them through their first transcription to see how it works in action. So they understand exactly what the application does more immediately rather than having to intuit uh what it is. I think most people somehow already know what the application is doing very precisely. Um but I wonder how many people read the documentation or read the website and don't quite have a perfect picture of what the application does yet. So I don't know. I'm wondering if there's any of these things maybe would smooth out the onboarding flow. I don't know. I'm sure uh most people have already gone through it and it's never come up again. So maybe it's not even a big deal.

lhotari:
> wondering if you think for onboarding there could be a step added for wayland or something? where it checks some of the availability and guides through some install of wtype or dotool? maybe even similar for enabling the shortcuts with the SIGUSR2??? I think generally the onboarding flow could be worked on a bit and maybe this is one area to smooth things out a bit too?

@cjpais Yes, it would be useful. Just wondering where to add this type of basic documentation:

When using Wayland:
* for toggling recording, you need to use a custom key shortcut of the Wayland compositor and make it call `pkill -USR2 -n handy`.
* You need to install either `wtype` https://github.com/atx/wtype  or `dotool` https://sr.ht/~geb/dotool/ so that handy can send `CTRL-V` or `Shift-Insert` to the application
* "Overlay Position" should be set to "None" to avoid losing focus

lhotari:
Just noticed #387 about the basic docs, that looks good to me.

PR #383, opened 2025-11-26
fix: re-assert always-on-top for overlay on Windows

Danielalnajjar:
## Summary

Fixed window z-order degradation causing recording overlay to periodically lose topmost flag on Windows, particularly after prolonged runtime.

## Solution

Uses Win32 `SetWindowPos` with `HWND_TOPMOST` to re-assert the overlay's Z-order immediately after showing it. This is more reliable than relying solely on Tauri's `alwaysOnTop` property.

- Called after `.show()` for correct timing
- Uses `run_on_main_thread` for thread safety
- `SWP_NOACTIVATE` prevents stealing focus from the active window

## Known Limitations

These are inherent to Windows topmost window behavior, not specific to this fix:

- Other apps with topmost windows can still compete for Z-order
- Fullscreen exclusive apps may not show the overlay (borderless windowed works fine)
- Elevated/admin windows can appear above the overlay

## Testing

Ran Handy continuously for multiple days on Windows 11 to reproduce the original issue (overlay losing topmost status after extended uptime). Confirmed the overlay now stays on top, even during periods of extended runtime.

### DIFF
--- a/src-tauri/src/overlay.rs
+++ b/src-tauri/src/overlay.rs
@@ -224,14 +224,6 @@ pub fn show_recording_overlay(app_handle: &AppHandle) {
             let _ = overlay_window.set_position(tauri::Position::Logical(tauri::LogicalPosition { x, y }));
         }
 
-        // Re-assert always-on-top for Linux
-        #[cfg(target_os = "linux")]
-        {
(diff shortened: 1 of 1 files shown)


cjpais:
Removed the linux related changes, I can't test for now so just going to merge and I will add you to threads if there are issues

Issue #390, opened 2025-11-28
[BUG] Version 0.6.3 broken on Linux Mint


ChrisB85:
## Bug Description

Transcription does not work - application freezes immediately after starting recording. The only option is to terminate the Handy process.

<img width="352" height="254" alt="Image" src="https://github.com/user-attachments/assets/3f60a34e-15bb-4a47-80cb-ec564a28d8a1" />

[handy.log](https://github.com/user-attachments/files/23815987/handy.log)

## System Information

**App Version:**

0.6.3

**Operating System:**

Linux Mint 22.2 Cinnamon

**CPU:**

AMD Ryzen 5 3600

**GPU:**

NVIDIA RTX 2070 SUPER

lstep:
Same here on Ubuntu 24.04.3 LTS
Logs don't show any error message at all.

cjpais:
Thanks for the reports. May roll back the version, as I don't have a system to test on. 

@lhotari are you able to replicate

@lstep do you get the same issue when building from source?

And to be clear this is a regression from older versions correct??

Can anyone confirm if this happens with the overlay disabled? or only with it enabled?

alabhyajindal:
I'm able to replicate this. I notice the program freezes when I stop the recording, though it may have froze at the start. Getting this issue on the latest release when overlay is both, enabled and disabled. 

I believe this regression is from the old version. Most recently, the application worked from me during development on the  #376 branch.

cjpais:
@alabhyajindal thank you that's helpful, would you mind trying commit `8136e6b`? 

I am suspecting maybe #224 since it is in the hotpath of the recording start/stop. Otherwise maybe its due to the specta overhaul i made?

alabhyajindal:
Just tried commit `8136e6b` and it's working! Of course, this is with disabling the overlay, since the overlay captures focus, but that's a different issue.

cjpais:
Thank you @alabhyajindal, can you try #392? I featured flagged off most of the cancel functionality in this PR. Wondering if it will generally fix things up here for the time being.

alabhyajindal:
Yes - works! Thank you!

cjpais:
Great pulling it in and cutting the next release, thank you for your quick testing! I really really appreciate it!

mg-dev25:
I noticed the audio recorder wasn't always picking the best settings. In fact, without this change, Handy was crashing in a loop on my machine.

[...]

PR #392, opened 2025-11-29
disable cancel on linux

cjpais:


cjpais:
@jacksongoode as heads up, #224 broke things on linux see: #390 

no big deal, but just disabling the feature on that platform for now. my guess is this is gonna be a bit annoying to fix on linux

jacksongoode:
Ah sorry about that! Yeah I didn't have a Linux machine to test it on but perhaps unsurprising. I'll try to debug why, maybe it would be good to make an issue for this?

PR #393, opened 2025-11-29
Improvement: Audio Recorder Quality & Crash Fix

mg-dev25:
### Improvement: Audio Recorder Quality & Crash Fix

While debugging audio issues on Linux/Wayland (related to #390), I noticed the audio recorder wasn't always picking the best settings. In fact, without this change, the application was crashing in a loop on my machine due to configuration mismatches.

I updated `recorder.rs` to explicitly check for `F32` support and prioritize that over `I16` or `I32` if available. This prevents the crash and ensures the highest possible input quality for transcription.

#### Technical Implementation
The `get_preferred_config` method now iterates through all supported configurations and assigns a score based on the sample format, rather than selecting the first match.

```rust
// Prioritize F32 > I16 > I32 > others
let score = |fmt: cpal::SampleFormat| match fmt {
    cpal::SampleFormat::F32 => 4,
    cpal::SampleFormat::I16 => 3,
    cpal::SampleFormat::I32 => 2,
    _ => 1,
};

if score(config_range.sample_format()) > score(current.sample_format()) {
    best_config = Some(config_range);
}
```

*Relates to #390*

### DIFF
--- a/src-tauri/src/audio_toolkit/audio/recorder.rs
+++ b/src-tauri/src/audio_toolkit/audio/recorder.rs
@@ -204,17 +204,34 @@ impl AudioRecorder {
         device: &cpal::Device,
     ) -> Result<cpal::SupportedStreamConfig, Box<dyn std::error::Error>> {
         let supported_configs = device.supported_input_configs()?;
+        let mut best_config: Option<cpal::SupportedStreamConfigRange> = None;
 
-        // Try to find a config that supports 16kHz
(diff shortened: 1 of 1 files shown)


cjpais:
@mg-dev25 im good to pull this in, but could you remove the change to bindings.ts, it's going to be autogenerated differently on different platforms so lets just keep it stable where it is for now

mg-dev25:
@cjpais Done.

Issue #193, opened 2025-10-09
[BUG] Handy steals focus on pressing hotkey to toggle recording ON


davidonabus:
## Bug Description
When pressing the assigned hotkey to toggle recording or transcription, the app steals input focus from the currently selected text field. This requires manually reselecting the text field before the transcribed text can be pasted.

On Windows, the hotkey works without changing focus, but on Pop!_OS (Linux), the active window loses focus when the hotkey is pressed.

## System Information
0.5.1

**Operating System:**
Pop!_OS (Linux)

**CPU:**
AMD Ryzen 9 7950X3D

**GPU:**
Nvidia 5090

cjpais:
Are you using the overlay? Try disabling it. This is a known issue, see #87

davidonabus:
Thanks that did it!

Issue #121, opened 2025-09-17
[BUG] Issue's working under Wayland


999sian:
Perfect, thanks for clarifying. So the AppImage does launch on Pop!_OS COSMIC, but the same clipboard/Wayland issues appear there too. The difference is that on CachyOS the AppImage won’t launch at all. Here’s a polished bug report that captures that clearly:

---

Bug Report

Description

The application behaves inconsistently across distributions:

CachyOS

AppImage: Fails to launch.

.deb (via Deb Tap): Launches, but clipboard paste after transcription does not work under Wayland.

COSMIC X11 compatibility: App launches and shortcuts work, but clipboard paste still fails.

GNOME with X11: Works correctly.

Pop!_OS 24.04 (COSMIC)

AppImage: Launches successfully, but clipboard paste after transcription fails under Wayland.

X11 session: Works correctly.

This points to two separate issues:

1. AppImage launch failure on CachyOS

2. Clipboard handling broken under Wayland (affects both CachyOS and Pop!_OS)

---

System Information

App Version: 0.51

Operating Systems Tested:

CachyOS (Wayland, COSMIC X11 compatibility, GNOME X11)

Pop!_OS 24.04 (COSMIC, X11)

CPU: Intel Ultra 7 155H

GPU: Intel Arc Graphics

---

Expected Behavior

The AppImage should launch on CachyOS.

Clipboard operations (copy/paste after transcription) should work under Wayland sessions on both distributions.

Actual Behavior

CachyOS AppImage: Does not launch.

Pop!_OS AppImage: Launches but clipboard paste fails under Wayland.

.deb on CachyOS: Launches, but clipboard paste fails under Wayland (works under X11).

Additional Notes

Likely related to Wayland clipboard APIs and AppImage runtime dependencies.

X11 sessions on both distros behave correctly, which suggests the issue is Wayland-specific.

---

Would you like me to expand this with a “Steps to Reproduce” for each distro/session (CachyOS AppImage fail, CachyOS .deb Wayland fail, Pop!_OS AppImage Wayland fail)? That would make it fully actionable for developers.

cjpais:
Thanks for the report across Distros. And especially Wayland vs X11. I will try to have a fix sometime.

If you could give more info about the failure on CatchyOS that would be helpful. 

I am unfortunately not going to be able to test all distros so I would recommend building from source.

999sian:
Under Cachy OS the App image was displaying s White Screen.
[...]

PR #394, opened 2025-11-29
Fix: Improve Paste Reliability with Overlay Timing Callback

### LINKED ISSUE #315: [BUG] Typing fails due to target window not focused at correct time
## Bug Description
Typing fails, apparently due to overlay stealing focus.

If I click on the target window really quickly after recording (before typing starts), then it works.

Otherwise, no text appears.
The target window regains focus only several seconds later, presumably after Handy has finished typing.
(So my guess would be that Handy is trying to type into the Handy overlay itself?)

Hiding the overlay solves the issue.
(Overlay Position: None)
In that case (overlay hidden), the target window regains focus immediately upon releasing the hotkey, and typing works normally.

## System Information
**App Version:** v0.6.0

**Operating System:**
Linux Mint 22.1

**CPU:**
Intel(R) Core(TM) i5-6300U CPU @ 2.40GHz

**GPU:**
Intel HD Graphics 520

mg-dev25:
### Fix: Improve Paste Reliability with Overlay Timing Callback

This PR fixes paste failures on Linux (especially Wayland) by ensuring the overlay is fully hidden and focus is restored to the target application before attempting to paste.

#### Problem

On Linux/Wayland, the current implementation attempts to paste text immediately after hiding the overlay. However, Wayland window managers handle focus changes asynchronously, meaning the previous application may not have regained focus by the time the paste operation is executed. This results in:

- Text not appearing in the target application
- Text being pasted into the overlay itself
- Users needing to manually click the target window before paste works

#### Solution

Introduces `hide_recording_overlay_with_callback()` which:
1. Triggers the fade-out animation (300ms)
2. Hides the overlay window
3. Waits an additional 150ms for the window manager to restore focus
4. Only then executes the paste operation via callback

This ensures reliable paste behavior across all Linux desktop environments, particularly on Wayland compositors.

#### Technical Changes

**`src-tauri/src/overlay.rs`:**
- Added `hide_recording_overlay_with_callback<F>()` function with callback support
- Comprehensive documentation explaining Wayland focus behavior

**`src-tauri/src/actions.rs`:**
- Updated `TranscribeAction::stop()` to use callback-based overlay hiding
- Paste now executes only after overlay is fully hidden and focus restored

#### Testing

Tested on:
- Linux Mint (Cinnamon/X11)
- Fedora (Wayland/GNOME)

#### Related Issues

- Fixes #315 - Typing fails due to target window not focused at correct time
- Relates to #121 - Clipboard paste fails under Wayland
- Relates to #193 - Handy steals focus on pressing hotkey

### DIFF
--- a/src-tauri/src/actions.rs
+++ b/src-tauri/src/actions.rs
@@ -5,6 +5,7 @@ use crate::managers::transcription::TranscriptionManager;
 use crate::settings::{get_settings, AppSettings};
 use crate::shortcut;
 use crate::tray::{change_tray_icon, TrayIconState};
+use crate::overlay::hide_recording_overlay_with_callback;
 use crate::utils::{self, show_recording_overlay, show_transcribing_overlay};
 use async_openai::types::{
     ChatCompletionRequestMessage, ChatCompletionRequestUserMessageArgs,
@@ -371,10 +372,14 @@ impl ShortcutAction for TranscribeAction {
                                 }
                             });
 
-                            // Paste the final text (either processed or original)
+                            // Hide overlay first, then paste when focus is restored.
+                            // This ensures reliable paste on Wayland/Linux where focus
+                            // changes are asynchronous.
                             let ah_clone = ah.clone();
-                            let paste_time = Instant::now();
-                            ah.run_on_main_thread(move || {
+                            hide_recording_overlay_with_callback(&ah, move || {
+                                let paste_time = Instant::now();
+                                change_tray_icon(&ah_clone, TrayIconState::Idle);
+
                                 match utils::paste(final_text, ah_clone.clone()) {
                                     Ok(()) => debug!(
                                         "Text pasted successfully in {:?}",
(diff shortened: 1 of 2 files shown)


cjpais:
So I've reviewed this briefly and also with the help of some large language models and I'm a bit skeptical to pull this in as is due to cross-platform concerns. Generally needing to review the history of the code, specifically in regards to the change away from running on the main thread. Quite frankly I'm not quite versed enough in the specifics of each platform to know if that was required anyhow, but the LLM's are concerned for what it's worth. Also there is some code duplication between `hide_recording_overlay` and `hide_recording_overlay_with_callback`. There's probably a much cleaner way to move forward. Maybe an optional callback.

Also, I'm not very happy with the delay that this introduces on MacOS. In testing, I find it quite awkward. It's fine for Linux because it fixes a real issue on many distros of Linux. But I don't think the delay is acceptable to have on Windows and MacOS and should be feature flagged off for just Linux.

I believe your fork prevents me from making any changes on this branch. I'm happy to both review and change files as necessary, if you're up to allow that.

mg-dev25:
@cjpais Because of this issues with github https://github.com/orgs/community/discussions/5634 
I don't have the option to allow edit of the Pr. So I give you access to write the fork create by my org. 
Tell me if this did not work.

cjpais:
Thanks, that worked, I will take a look when I can

PR #405, opened 2025-12-04
Add option to append trailing space after pasted text

dannysmith:
Adds an option to the debug menu to add a trailing space after the text is pasted (see [here](https://github.com/cjpais/Handy/discussions/223))

I added it to the debug menu because that seemed less obtrusive, but it could easily be moved to the General or Advanced pane, or be "JSON config only" and have no UI.


cjpais:
ty for adding to the debug menu, when I get a chance I will test and pull in.

cjpais:
works for me, pulling it in

PR #408, opened 2025-12-04
Prevent possible deadlocks when pressing Escape to cancel recordings

dannysmith:
So #224 allows the Escape key to be used to cancel recording. My #401 basically implemented this in a similar (though more generic way) before 224 was merged. Looking at [this commit in 401](https://github.com/cjpais/Handy/pull/401/commits/2915aea92f8b3000f263ef1beb64dba978d16e93), I think the only thing worth keeping from my implementation is this...

## Problem

In toggle mode (non-push-to-talk), the shortcut callback holds the ManagedToggleState
lock while calling action.start() or action.stop(). If the action needs to acquire the
same lock (e.g., cancel_current_operation resets toggle states), this causes a
deadlock.

## Solution

Release the lock before calling the action:

```typescript
let should_start: bool;
{
    let mut states = toggle_state_manager.lock()...;
    should_start = !*is_currently_active;
    *is_currently_active = should_start;
} // Lock released

if should_start { action.start(...) } else { action.stop(...) }
```

## Steps after this PR is merged

- [ ] Refactor #401 to remove escape (even though it'll [probably not get merged](https://github.com/cjpais/Handy/pull/401#issuecomment-3609975966), I'm using it in my fork until we have a better way of handling `fn` key on macOS)

@cjpais If you'd prefer the slightly more generic approach to handling dynamic keybindings in 2915aea92f8b3000f263ef1beb64dba978d16e93 I can also add that to this PR too.

### DIFF
--- a/src-tauri/src/signal_handle.rs
+++ b/src-tauri/src/signal_handle.rs
@@ -25,32 +25,40 @@ pub fn setup_signal_handler(app_handle: AppHandle, mut signals: Signals) {
                     let shortcut_string = "SIGUSR2";
 
                     if let Some(action) = ACTION_MAP.get(binding_id) {
-                        let toggle_state_manager =
-                            app_handle_for_signal.state::<ManagedToggleState>();
+                        // Determine action and update state while holding the lock,
+                        // but RELEASE the lock before calling the action to avoid deadlocks.
+                        // (Actions may need to acquire the lock themselves, e.g., cancel_current_operation)
+                        let should_start: bool;
+                        {
+                            let toggle_state_manager =
+                                app_handle_for_signal.state::<ManagedToggleState>();
 
-                        let mut states = match toggle_state_manager.lock() {
-                            Ok(s) => s,
-                            Err(e) => {
-                                warn!("Failed to lock toggle state manager: {e}");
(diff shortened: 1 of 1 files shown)


cjpais:
I really liked the dynamic keybindings bit when I reviewed from what I recall if you could bring that refactor in too it would be great. I think it starts to set us up a lot better for moving to letting users define all kinds of keybindings if they want

dannysmith:
#224 includes dynaming keybinding already, but not in an easily reusable way. I'll bring that stuff over into here from #401 because yeah - I agree. Will ping you when I ned a review

cjpais:
I had claude port over the dynamic bindings, I haven't reviewed honestly. It does work from my brief testing, but would be helpful to have your review as well. I will take a closer look at the code itself when I can. If I just messed things up, feel free to revert the changes

cjpais:
Im running into some issues while testing the dynamic part. I think maybe it would be a nice thing to bring in but I'd rather bring in the fix now rather than later so I'm going to revert to your merge, re-merge, test and pull in

jacksongoode:
Thanks for the PR! This implementation definitely catches a major flaw.

PR #411, opened 2025-12-05
feat: context-aware capitalization

rkpatel33:
## Summary

- Adds context-aware capitalization using macOS Accessibility APIs
- Capitalizes first letter after sentence-ending punctuation (. ! ?)
- Lowercases first letter after continuation punctuation (, ; : -)
- Adds trailing space after sentence-ending punctuation in output
- Falls back to capitalize + trailing space when context can't be read (terminal apps)

## Test plan

- [x] Test in supported apps (TextEdit, browser text fields)
- [x] Test in unsupported apps (terminal) - fallback mode works
- [x] Unit tests pass

### DIFF
--- a/.claude/docs/2025-12-07-01_streaming-output-architecture.md
+++ b/.claude/docs/2025-12-07-01_streaming-output-architecture.md
@@ -1,0 +1,387 @@
+---
+Title: Streaming Output Architecture Analysis
+Created: 2025-12-07
+Last Updated: 2025-12-07
+Tags: architecture, streaming, transcription, feature-design
+---
+
+# Streaming Output Architecture Analysis
+
+## Executive Summary
+
+This document analyzes the feasibility and approach for implementing "eager streaming mode" - incrementally outputting transcription results while the user holds down the hotkey, with intelligent updates at pause points.
+
+**Verdict: Feasible, but requires significant architectural changes**
+
+The current architecture processes audio as a batch after button release. Implementing streaming requires:
+1. A streaming-capable transcription backend (current `transcribe-rs` doesn't support this)
+2. New pause detection logic during recording
+3. Text replacement infrastructure with backspace fallback
+4. Settings integration for local-only mode
+
+---
+
+## Current Architecture Overview
+
+### Flow: Button Press → Text Output
+
+```
+┌──────────────┐     ┌─────────────────────┐     ┌─────────────────────┐
+│ Button Press │────▶│ AudioRecordingMgr   │────▶│ Vec<f32> (batch)    │
+│ (shortcut)   │     │ accumulates samples │     │ after button release│
+└──────────────┘     └─────────────────────┘     └─────────────────────┘
+                                                            │
+                     ┌─────────────────────┐                ▼
+                     │ TranscriptionMgr    │◀───────────────┘
+                     │ transcribe(Vec<f32>)│
+                     └─────────────────────┘
+                                │
+                                ▼
+┌──────────────────────────────────────────────────────────────────────┐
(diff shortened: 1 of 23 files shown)


cjpais:
I'm going to be a bit frank with this one. I'm okay to add this feature, but this code does a lot of other things than just this feature, and you gotta review it yourself before I spend time to review it.

Specifically, it also adds a shortcut. Let's keep it to one main change, not a bunch of changes. Makes it simpler for me to review and understand exactly what the code is doing. I don't need to parse through other bits of code that seem unrelated to the feature that you suggested.

I'm generally okay with adding this feature as long as the code for it is well thought out.

src-tauri/src/context.rs
```diff
@@ -0,0 +1,565 @@
...
+            debug!("Failed to get focused element: {:?}", e);
+            return None;
+        }
+    };
+
+    // Cast CFType to AXUIElement
+    let focused = unsafe { AXUIElement::wrap_under_get_rule(focused_cftype.as_CFTypeRef() as _) };
+
+    // Get the text value of the focused element using the predefined value() attribute
+    let value_cftype: CFType = match focused.value() {
```
dannysmith, line 77: This might break weirdly when the currently focussed text element contains a **very** long chunk of text? I guess if that happens it'll just failover to the fallback behaviour tho, right? How will this behave if `focused.value()` is a gagillion characters long?

src-tauri/src/context.rs
```diff
@@ -0,0 +1,565 @@
...
+
+    let should_add_trailing_space = ends_with_sentence_punctuation || !context.context_readable;
+
+    let result = if should_add_trailing_space {
+        format!("{} ", capitalized)
+    } else {
+        capitalized
+    };
+
+    info!(
```
dannysmith, line 379: Feel like this should be `debug!`

dannysmith:
This is cool. Never really seen the accessibility API before. Having literally just skimmed over the diff...

- `mod context;` in lib.rs is not gated for macos only so this module is compiled on all platforms, which I guess is fine because the macOS-specific stuff in `context.rs` is all properly gated. From what I can see on Win/Linux, everything "works" but always falls back to "capitalize + add trailing space". I guess the best approach here depends onhoweasy it is to implement `get_text_before_cursor()` for the other platforms. Likewise for showing the UI switch etc.
- I feel like this might behave weirdly with emojis that are composed of more than one character? Would need to test tho.
- `apply_capitalization()` and `find_relevant_punctuation()` feel like they might be worth extracting into some generic "text utils"? I can see other stuff wanting to do similar in the future. But also _premature optimisation_?
- Falling back to adding a trailing space might be confusing? Especially as there's [already](https://github.com/cjpais/Handy/pull/405) an option for that.
- If #391 is merged this could be achieved with a Post-processing rule on machines which support Apple Intelligence, though that would add more latency.

Over to @cjpais tho!

cjpais:
@rkpatel33 can you rebase to main, I also pretty much agree with @dannysmith feedback on the whole

Please refrain from adding any changes that are not related specifically to this PR (CLAUDE.md, and CHANGELOG.md) I will let @dannysmith drive some of this as we move forward in the future, but for now let's just keep the changes as minimal and focused as possible. More rules and guidelines on contribution will come in the future.

cjpais:
@rkpatel33 if you're going to do active development on this PR without respecting any comments it will be closed. You continue to diverge without response. You're welcome to continue development on your own branch before opening a PR. 

I actively monitor all code coming into Handy.

cjpais:
Closing. You're welcome to add scoped features in the future, but please keep them scoped!

PR #435, opened 2025-12-11
upgrade to transcribe-rs-0.1.5

cjpais:
this is upgrading the whisper bindings

drops support for older macOS versions


cjpais:
upgrading the version of whisper is not good right now. without spending a lot of time to investigate the bindings and potentially creating a new set I don't think a whisper.cpp upgrade is going to happen in the near term.

transcribe-rs now supports whisperfile, which will be included as an option in the future. see #457

PR #437, opened 2025-12-11
Add Internationalization Support

cjpais:
The application language can now be changed on the "About" page

Screenshot 2025-12-11 at 5 34 12 PM

### DIFF
--- a/CONTRIBUTING_TRANSLATIONS.md
+++ b/CONTRIBUTING_TRANSLATIONS.md
@@ -12,6 +12,7 @@ Thank you for helping translate Handy! This guide explains how to add or improve
 ## File Structure
 
 Translation files are located in:
+
 ```
 src/i18n/locales/
 ├── en/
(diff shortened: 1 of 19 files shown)


dannysmith:
@cjpais just seen this - great shout adding this. 

It feels a bit weird to me that this is in the About panel, since it's a setting (and the only one in this panel). I assume that you put it here because if affects only the Handy UI and not the actual functionality in use (ie transcription)?

I almost feel like the panes should be something more like:

- Transcription (incl some of the advanced stuff)
- Post-Processing
- Interface (incl overlay position, UI language)
- History
- Debug
- About

But this should probably be it's own discussion.

cjpais:
@dannysmith - yeah I initially wanted it somewhere a bit more obvious but disambiguating between what language is being transcribed vs the UI felt awkward. I really think the UI needs another overhaul generally. Maybe it makes sense to have a globe icon > English dropdown or similar somewhere. Potentially replacing the updates available bit? Or maybe it's in a consistent spot in the UI? I'm not really sure but it could use some help.

I started #449 as a larger discussion about the app and UI generally

PR #440, opened 2025-12-12
move to macos 26 ci

cjpais:


cjpais:
should fix the bug of Apple Intelligence builds

Issue #194, opened 2025-10-10
[FEATURE] Dashboard for Handy


deadcoder0904:
I feel Monologue by Every.to or VoiceInk both have a good-looking dashboard.

It is a good way to share too to promote Handy.computer since people love sharing their stats. Can make it like the recent X.com screenshot change that Nikita Bier added to X.

I'd love a Dashboard like this that tracks all my metrics. Or even Github-like streak.

This is from VoiceInk Dashboard:

<img width="1608" height="1644" alt="Image" src="https://github.com/user-attachments/assets/e4c353a2-01b7-46b2-992f-41d130b1c87f" />

cjpais:
I'm not interested in adding this at the moment

PR #442, opened 2025-12-12
feat: add analytics dashboard

VirenMohindra:
## Summary
- adds analytics tab to sidebar showing transcription statistics
- tracks total words, recording time, average wpm, and current streak
- supports filtering by today, this week, or all time
- includes proper error handling and singular / plural formatting
- this feature only tracks new transcriptions going forward and does not backfill historical data. existing transcriptions will not have analytics data
- this semi-addresses the feature request in #194, which was previously closed as not planned. happy to discuss if there are concerns about scope or maintenance

handy-analytics

## Testing
- [x] create a new transcription and verify analytics update
- [x] check all three time periods (today, this week, all time)
- [x] verify streak increments on consecutive days
- [x] confirm singular / plural formatting ("1 day" vs "2 days")

### DIFF
--- a/src-tauri/src/actions.rs
+++ b/src-tauri/src/actions.rs
@@ -1,6 +1,7 @@
 #[cfg(all(target_os = "macos", target_arch = "aarch64"))]
 use crate::apple_intelligence;
 use crate::audio_feedback::{play_feedback_sound, play_feedback_sound_blocking, SoundType};
+use crate::audio_toolkit::constants;
 use crate::managers::audio::AudioRecordingManager;
 use crate::managers::history::HistoryManager;
 use crate::managers::transcription::TranscriptionManager;
@@ -395,6 +396,13 @@ impl ShortcutAction for TranscribeAction {
                             // Save to history with post-processed text and prompt
                             let hm_clone = Arc::clone(&hm);
                             let transcription_for_history = transcription.clone();
+
+                            // Calculate analytics data
+                            let duration_seconds =
+                                samples_clone.len() as f32 / constants::WHISPER_SAMPLE_RATE as f32;
+                            let word_count =
+                                transcription_for_history.split_whitespace().count() as i32;
+
                             tauri::async_runtime::spawn(async move {
                                 if let Err(e) = hm_clone
                                     .save_transcription(
@@ -402,6 +410,8 @@ impl ShortcutAction for TranscribeAction {
                                         transcription_for_history,
                                         post_processed_text,
                                         post_process_prompt,
+                                        duration_seconds,
+                                        word_count,
                                     )
                                     .await
                                 {
--- a/src-tauri/src/commands/history.rs
+++ b/src-tauri/src/commands/history.rs
@@ -1,4 +1,4 @@
-use crate::managers::history::{HistoryEntry, HistoryManager};
+use crate::managers::history::{AnalyticsStats, HistoryEntry, HistoryManager};
 use std::sync::Arc;
 use tauri::{AppHandle, State};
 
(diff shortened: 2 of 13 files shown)


cjpais:
going to let @dannysmith chime in here with some opinions. @VirenMohindra if you dont mind, can you get some votes in discord for this and/or on the Github discussions?

I just want to see the support for it from the community. I don't mind adding this. I'm personally just not exactly sure I would ever open it haha

I also think the history feature might need to be refined a bit for this to really work great. Mainly it keeps only a few recordings right now by default. And I know people are also disabling it, which would need to be handled.

VirenMohindra:
added the relevant [thread in discord](https://discord.com/channels/1391636091853078608/1395929320303890514/1449016355964583987) to drum up some votes

i know this is a bit OOS to the project so happy to let folks weigh in ahead of time


> And I know people are also disabling it, which would need to be handled.

good catch, can take a look at this once we have a quorum

dannysmith:
So my take on this is basically:

1. The implementation looks sensible at first glance ❤️
2. I'm not sure this feature belongs in Handy.

To elaborate on that second point beyond "I wouldn't use it myself"... 

Handy is a utility app which to me means:

- It should do what users expect **really well**, reliably, and on all platforms.
- It should do nothing else.

Handy's primary UI is a keyboard shortcut and the `RecordingOverlay`. The settings window should be just that - a place to configure Handy which users never look at unless they're changing a setting. We should be focused on adding features which make voice transcription work **really well** in as many contexts as possible. Features like this feel like a (small) step in a different direction, and my main concern is that it's a slippery slope to code and feature bloat down the line. To quote @cjpais on the website (bold mine):

> Handy isn’t trying to be the best speech-to-text app. It’s trying to be the most forkable one.
> 
> Handy is a starting point for when you need accessibility features, want to experiment with voice computing, or just prefer tools you can actually own and modify. **The project is designed to be tiny and extensible**. 

Having said that, if it turns out a load of people really want this, I've no strong objection.

@VirenMohindra As someone who obviously would find this valuable, I'm legit interested in your use case. When do you look at these analytics and how do you use the information?

cjpais:
I think I am in general agreement @dannysmith. I'll wait a few days to see anything regarding community feedback, but I think it might be moving away from the purpose of handy a bit. There are some features like transcribing audio files which feel close enough to purpose to pull in even though it is feature creep, but metrics aren't really my jam personally.

VirenMohindra:
thanks for the thoughtful feedback @dannysmith @cjpais - totally understand the concern about scope creep and keeping handy minimal

to answer your q about my use case: i mainly wanted to track my usage patterns over time - seeing how much i'm actually using voice input vs typing, and the streak was a fun way to build the habit. kind of like how duolingo gamifies language learning

that said, i completely get the "slippery slope" and bloat concern. Handy's value is in being small and focused. if the consensus is this doesn't fit the project's direction, i'm happy to~
1. close this pr and keep it as a personal fork
2. or strip it down to something more minimal if there's a version that would fit better

no hard feelings either way - appreciate you both taking the time to explain the thinking behind it. will keep an eye on the discord thread for any community feedback, but where it currently stands it seems like there is minimal interest

dannysmith:
I guess for me, if this can be stripped down so something which is:

1. Not very much code.
2. Unobtrusive in the UI (perhaps something in the "About settings pane")

... then the whole "slippery slope" thing is much less of a concern for me. I guess we could also write this kinda data to disk in a way that folks like @VirenMohindra can read it from other tools?

Suggest leaving this PR open for now and seeing if there's any more interest in a bit? (or closing and re-opening if there is?) Over to @cjpais for a decision there.

Thanks for explaining your use case @VirenMohindra 🙏

> and the streak was a fun way to build the habit

That's interesting coz _streaks_ was the only bit of this where I thought "I might look at that occasionally".

cjpais:
I think I'm going to close this one for now. I think we can monitor interest in the future and see what happens. I'm not explicitly against it, but just don't know if there's a strong enough motivation at the moment. I do appreciate the contribution as well as thoughtful discussion (and non-attachment)!

I'm personally a bit against 'streaks' unless there's an obvious reason that a streak is helping you in a meaningful way. I don't feel it's very meaningful to know if you're typing with your voice or not personally. I think it's on the borderline of being a dark pattern/fueling some addictive tendency

PR #443, opened 2025-12-12
fix: prevent model auto-switch from interrupting active transcription

### LINKED ISSUE #418: [BUG] Model download completion auto-switches and interrupts active transcription
## Bug Description

When a model finishes downloading in the background while the user is actively transcribing (or recording), the application automatically switches to the newly downloaded model, interrupting the active transcription. The transcription is cut off and the result is lost.

## Steps to Reproduce

1. Have Model A already downloaded and active (e.g., Whisper Small)
2. Start downloading Model B (e.g., Whisper Medium) in the background
3. While Model B is downloading, start a transcription with Model A
4. Let Model B finish downloading during the active transcription
5. **Result**: Transcription is interrupted/cut off as the app auto-switches to Model B

## Expected Behavior

The active transcription should complete normally. The newly downloaded model should either:
- Not auto-switch at all (user manually selects when ready)
- Wait until the current transcription completes before switching
- At minimum, prompt the user before switching

## Actual Behavior

The model auto-switches immediately (after 500ms delay), regardless of whether a transcription is in progress, causing:
- Active transcription to be interrupted
- Potential loss of transcribed content
- Poor user experience

## Root Cause Analysis

**Location**: `src/components/model-selector/ModelSelector.tsx` lines 163-168 and 192-198

```typescript
// Listen for model download completion
const downloadCompleteUnlisten = listen<string>(
  "model-download-complete",
  (event) => {
    const modelId = event.payload;
    // ...

    // Auto-select the newly downloaded model  <-- BUG
    setTimeout(() => {
      loadCurrentModel();
      handleModelSelect(modelId);  // Switches immediately without checking state
    }, 500);
  },
);
```

Same issue exists for `model-extraction-completed` event handler.

**Why it causes interruption**:

1. `handleModelSelect()` calls `commands.setActiveModel(modelId)`
2. Backend `set_active_model` (`src-tauri/src/commands/models.rs:65-67`) calls `transcription_manager.load_model()`
3. `load_model` (`src-tauri/src/managers/transcription.rs:251-254`) acquires the engine mutex and replaces the active engine
4. If transcription is in progress, it gets interrupted when the engine changes

## Possible Fixes

**Option A - Remove auto-switch**:
Remove the auto-select logic entirely. User manually selects the model when ready.

**Option B - Check state before switching**:
Query backend for recording/transcription state before auto-switching. Only switch if idle.

**Option C - Queue the switch**:
Add the model ID to a pending queue, switch after current transcription completes.

## Environment

- **App Version**: v0.6.4
- **OS**: Windows (reported), likely affects all platforms (cross-platform code path)
- **Affected Files**:
  - `src/components/model-selector/ModelSelector.tsx`
  - `src-tauri/src/commands/models.rs`
  - `src-tauri/src/managers/transcription.rs`

## Additional Context

This is particularly problematic for users who:
- Download multiple models to try them out
- Have slow internet (long download times overlap with usage)
- Use the app continuously while models download in background

VirenMohindra:
## Summary
- prevents model auto-switch from interrupting active recordings/transcriptions
- adds `is_recording` tauri command to check recording state
- skips auto-switch if recording in progress, preserving transcription data

## Problem
when a model finishes downloading while user is actively transcribing, the app auto-switches to the new model after 500ms, interrupting the transcription and losing data.

## Solution
check recording state before auto-switching. if recording is in progress, skip the auto-switch. user can manually select the model after transcription completes.

## Testing
- [x] download model A, start transcription
- [x] start downloading model B during transcription
- [x] let B finish downloading
- [x] verify transcription completes normally without interruption
- [x] after transcription, manually select model B and verify it works

fixes #418

### DIFF
--- a/src/bindings.ts
+++ b/src/bindings.ts
@@ -517,13 +517,8 @@ async getClamshellMicrophone() : Promise<Result<string, string>> {
     else return { status: "error", error: e  as any };
 }
 },
-async isRecording() : Promise<Result<boolean, string>> {
-    try {
-    return { status: "ok", data: await TAURI_INVOKE("is_recording") };
-} catch (e) {
-    if(e instanceof Error) throw e;
-    else return { status: "error", error: e  as any };
-}
+async isRecording() : Promise<boolean> {
+    return await TAURI_INVOKE("is_recording");
 },
 async setModelUnloadTimeout(timeout: ModelUnloadTimeout) : Promise<void> {
     await TAURI_INVOKE("set_model_unload_timeout", { timeout });


cjpais:
thank you! will be testing this after i play around with something on my dev branch

cjpais:
I tested this and seems to work.

PR #444, opened 2025-12-12
feat: add german, japanese, and chinese translations

VirenMohindra:
## Summary
adds complete translations for three of the requested languages:
- german (de) - deutsch
- japanese (ja) - 日本語
- chinese simplified (zh) - 中文

## Context
builds on the internationalization foundation added in #437. this addresses requests from the community for additional language support.

## Changes
- added `src/i18n/locales/de/translation.json` - complete german translation
- added `src/i18n/locales/ja/translation.json` - complete japanese translation  
- added `src/i18n/locales/zh/translation.json` - complete chinese translation
- updated `src/i18n/languages.ts` to register new languages
- updated `CONTRIBUTING_TRANSLATIONS.md` to mark these languages as complete

## Translation Notes
- all ~400 strings translated for each language
- preserved all `{{variables}}` exactly as in english source
- kept technical terms (api, gpu, whisper, etc.) untranslated per guidelines
- maintained consistent tone with existing translations

## Testing
- [x] select german in settings, verify ui displays correctly
- [x] select japanese in settings, verify ui displays correctly
- [x] select chinese in settings, verify ui displays correctly
- [x] confirm no broken layouts or truncated text

## Screenshots

| Translations      | Screenshots |
| ----------- | ----------- |
| default - english, nothing broken, languages displayed      | Screenshot 2025-12-12 at 6 11 33 PM       |
| de (german)      | Screenshot 2025-12-12 at 6 11 41 PM       |
| jp (japanese)   | Screenshot 2025-12-12 at 6 11 47 PM        |
| zh (chinese)      | Screenshot 2025-12-12 at 6 11 36 PM       |

## Supported Languages After This PR
| language   | code | status  |
|------------|------|---------|
| english    | en   | source  |
| chinese    | zh   | new     |
| french     | fr   | existing|
| german     | de   | new     |
| japanese   | ja   | new     |
| spanish    | es   | existing|
| vietnamese | vi   | existing|

PR #446, opened 2025-12-12
feat: add tray menu localization using system locale

VirenMohindra:
## summary
- adds localization support for the system tray menu (settings, check for updates, quit, cancel)
- tray menu updates immediately when app language is changed in settings (no restart required)
- uses the same translation json files as the frontend (single source of truth)
- supports 9 languages: english, spanish, french, vietnamese, german, japanese, chinese, italian, polish

## technical details
- `build.rs` auto-generates tray translations at compile time from frontend locale files
- struct fields derived from english json keys (camelCase → snake_case)
- languages auto-discovered from `src/i18n/locales/` directory - no code changes needed for new locales
- locales without a "tray" section are skipped gracefully (falls back to english)
- uses existing `tauri-plugin-os::locale()` for os locale detection
- language priority: app setting → os locale → english fallback

| en      | fr |
| ----------- | ----------- |
|       |        |

## test plan
- [x] verify tray displays correct language based on app setting
- [x] verify tray updates immediately when changing language (no restart)
- [x] verify fallback to os locale when app language is "auto"
- [x] verify fallback to english for unsupported locales
- [x] test all 9 supported languages (en, es, fr, vi, de, ja, zh, it, pl)

### DIFF
--- a/src-tauri/src/tray_i18n.rs
+++ b/src-tauri/src/tray_i18n.rs
@@ -34,8 +34,3 @@ pub fn get_tray_translations(locale: Option<String>) -> TrayStrings {
         .cloned()
         .expect("English translations must exist")
 }
-
-/// Get the current system locale
-pub fn get_system_locale() -> Option<String> {
-    tauri_plugin_os::locale()
-}
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -219,7 +219,8 @@ Community feedback is essential to keeping Handy the best it can be for everyone
      - How you tested the changes
      - Screenshots/videos if applicable
      - Breaking changes (if any)
-       **Remember:** PRs with community support are prioritized. If you haven't already, start a [discussion](https://github.com/cjpais/Handy/discussions) to gather feedback before or alongside your PR. It is not explicitly required to gather feedback, but it certainly helps your PR get merged faster.
+
+   **Remember:** PRs with community support are prioritized. If you haven't already, start a [discussion](https://github.com/cjpais/Handy/discussions) to gather feedback before or alongside your PR. It is not explicitly required to gather feedback, but it certainly helps your PR get merged faster.
 
 ### Code Style Guidelines
 
--- a/src-tauri/src/settings.rs
+++ b/src-tauri/src/settings.rs
@@ -363,7 +363,9 @@ fn default_post_process_enabled() -> bool {
 }
 
 fn default_app_language() -> String {
-    "en".to_string()
+    tauri_plugin_os::locale()
+        .and_then(|l| l.split(['-', '_']).next().map(String::from))
+        .unwrap_or_else(|| "en".to_string())
 }
 
 fn default_post_process_provider_id() -> String {
--- a/src-tauri/src/tray.rs
+++ b/src-tauri/src/tray.rs
@@ -1,5 +1,5 @@
 use crate::settings;
-use crate::tray_i18n::{get_system_locale, get_tray_translations};
+use crate::tray_i18n::get_tray_translations;
 use tauri::image::Image;
 use tauri::menu::{Menu, MenuItem, PredefinedMenuItem};
 use tauri::tray::TrayIcon;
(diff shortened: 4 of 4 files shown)


cjpais:
ah nice i was going to do this in a second pass on localization so thank you

only request is don't add a new dep and instead use tauri-os built in

https://docs.rs/tauri-plugin-os/latest/tauri_plugin_os/fn.locale.html

it would be good if it didn't require an app restart, but its okay for now. there might be a function already to refresh the tray, so that might need to be run when the setting changes

VirenMohindra:
> only request is don't add a new dep and instead use tauri-os built in

done!  now using `tauri_plugin_os::locale()`

> it would be good if it didn't require an app restart

i looked into this and also managed to refresh the cart on lang change, the function helped

also added tray translations for german, japanese, and chinese (the new languages from #444)

src-tauri/src/tray_i18n.rs
```diff
@@ -0,0 +1,74 @@
...
+// Embed translation JSON files at compile time
+static TRANSLATIONS: Lazy<HashMap<&'static str, TrayStrings>> = Lazy::new(|| {
+    load_translations! {
+        "en" => "../../src/i18n/locales/en/translation.json",
+        "es" => "../../src/i18n/locales/es/translation.json",
+        "fr" => "../../src/i18n/locales/fr/translation.json",
+        "vi" => "../../src/i18n/locales/vi/translation.json",
+        "de" => "../../src/i18n/locales/de/translation.json",
+        "ja" => "../../src/i18n/locales/ja/translation.json",
+        "zh" => "../../src/i18n/locales/zh/translation.json",
```
dannysmith: This mapping feels like it should not be hardcoded and should be imported from a config file or inferred from file paths?

Ignore me if there's a good reason for not doing this tho.

dannysmith: As in... the need to say 

```
//! NOTE: When adding a new language to the frontend (src/i18n/locales/),
//! remember to also add it to the load_translations! macro below.
```

feels like a smell here?

dannysmith:
Note to self for when I'm on a machine where I can build/run this:

1. How will this code behave when the language is set to a language not supported by `tray_i18n.rs`.
2. How will this code behave if the relevant `translation.json` does not contain the required translations?

src-tauri/src/tray_i18n.rs
VirenMohindra: the constraint here is rust's `include_str!` macro - it requires compile-time string literals, so there's no way to dynamically discover files without either~
1. a build.rs script that scans the locales directory and generates the mapping
2. a procedural macro that reads the filesystem at compile time

both add complexity that felt like overkill for just seven languages. the explicit list has the benefit of compile-time safety (missing file equals build error) and the NOTE comment documents the manual step

cjpais: I think build.rs script might be most appropriate just to reduce complications for new contributors. It would probably be a write once and forget about it forever piece of code which is probably a good thing. I don't think a comment is sufficient for new contributors if they haven't read the code at all. This either needs to be documented in a README or similar, or some alternative approach. 

I'll have a think about the best way to approach this problem as well

VirenMohindra: i can get a start on `build.rs` and push it in in a sec

cjpais:
Another misc comment is we almost certainly need to have a setting for the UI language to be persisted in the settings file itself. I believe right now it's a local-storage value in the browser (which is definitely a hack imo). We probably should remove that in favor of a json stored string, so that the backend and frontend can both have a good source of truth for the language. 

In regards to the auto detection of the language, I'm not 100% the best way of going about that. Since it would either need an 'auto' setting itself, or something like `en-auto` to signify somewhere that this was an auto-picked language code. I forget if this code has it specifically, but since it's using a way to get the locale it probably is. Possibly we just drop support for auto-detection of locale and force an explicit pick?

cjpais:
@VirenMohindra when/if you get a chance can you merge with main, I took care of the using the settings file. There should be an appropriate setting to be able to pull from on the rust side. Would guess it will even clean up the implementation a bit

VirenMohindra:
> @VirenMohindra when/if you get a chance can you merge with main, I took care of the using the settings file

yeah thank you! will do right now. the build.rs implementation took a bit and wanted to land that to get some eyes on it before pinging ya

cjpais:
thank you! no worries, just excited to get this one in :)

src-tauri/src/tray_i18n.rs
VirenMohindra: added -- let me know if this works!

PR #453, opened 2025-12-14
Add Polish translation

Rejerf:
Initial pull with Polish translation added


cjpais:
I don't know polish but it show up in the UI and seems to work, merged!

Thanks for your contribution!

PR #459, opened 2025-12-15
Add Italian translation

giuliofoletto:
Italian translation based on the English version at the previous commit. I mostly translated manually, with some help from DeepL when needed. I kept some words in English to reflect current use (e.g. "Download" instead of literal "Scaricamento" which is not used) or to avoid confusion (e.g. "Log" instead of "Registro", which is too similar to "Registrazione" which actually means "Recording").


cjpais:
Awesome thank you so much!

PR #466, opened 2025-12-16
feat: add external providers for audio transcription and LLM-based post-processing [post processing is similer to PR #355]

avijitbhuin21:
## Summary
This PR adds support for external AI providers for both audio transcription and post-processing, enabling users to leverage cloud-based services for faster and more accurate speech-to-text conversion. This speeds up the processing and post processing significantly using provider like groq or cerebras even in my potato pc.

### 🎤 Online Audio Transcription Providers
- Added support for multiple external transcription providers:
  - **OpenAI** (Whisper)
  - **Groq** (Whisper Large V3 / V3 Turbo)
  - **Gemini** (2.0/2.5 Flash, Flash Lite, Pro models)
  - **SambaNova** (Whisper Large V3)
- Per-provider API key storage with secure masking
- Per-provider model selection with persistence
- Toggle to switch between local and online providers

### 📝 Post-Processing with LLM Providers  
- Extended post-processing to support multiple LLM providers:
  - OpenAI, OpenRouter, Gemini, Groq, Cerebrus, SambaNova, and Custom endpoints
- Configurable base URL for self-hosted/custom endpoints
- Model fetching and selection with refresh capability
- Custom prompt management (create, edit, delete prompts)

### 🔧 Backend Changes
- New Tauri commands for online provider settings management
- API key and model storage per provider
- Settings persistence for both transcription and post-processing providers

## UI Changes
- New "Online Providers" settings panel with provider/model/API key configuration
- Enhanced "Post Processing" settings with multi-provider support
- Consistent UI patterns with the existing settings design

## Related
Similar to PR #355 but focuses on settings-based configuration without hotkey integration.

## Screenshots
image
image
image
image


cjpais:
To be honest with you, I'm not exactly sure what to say. There's been many discussions around this topic and why it hasn't been approved before. Have you read them? What makes this PR different? I actually think you've done the best job of implementing the UI so far, but that doesn't really change my overall stance on API models. I am not particularly interested in having them in Handy for a variety of reasons at the moment and you can go ahead and read the existing lengthy conversations on this. You're welcome to continue to challenge my opinion and position as well as collect feedback from the community. Clear and obvious feedback, support, and discussion from the community on the best way forward would probably sway my opinion.

I may consider this PR if you remove the part with API models for transcription and let's just consider the post-processing part for now. 

Also if you are implementing API models you should use the support `transcribe-rs` has for it already.. and if it is not sufficient we should add more support there. I know this is almost certainly a vibe coded PR so this probably wasn't thought of. If it was vibe coded it would probably be helpful to know the prompts because at least it shows a clear intent and purpose, because a bunch of LLM generated text as the PR is something I will basically always skip reading because I want to know a humans intention not a machines intention.

Unfortunately I have to be a bit defensive of what makes it into the codebase and I really need to have someone who has a strong enough opinion or community support to have things make it into the repo. Not every PR can be accepted otherwise the entire app turns into an unmaintainable mess and it's already close to that in my opinion.

cjpais:
You gotta clean this PR up. It's not in any state to be merged. There's a bunch of breaking changes, things like changing tauri.conf.json and other stuff is not gonna fly.

Review every file for the changes, and only submit what is absolutely necessary. I skimmed the PR and don't have time to review a bunch of random changes to files that don't seem meaningful.

avijitbhuin21:
Hi @cjpais,

Thank you for the detailed and thoughtful feedback! I apologize for not reading through the previous discussions (#77, #279, #222, Discussion #168) before submitting. I've now gone through them and understand your position on keeping Handy primarily a local transcription app. 

## What I've done with this PR: 

- ✅ Removed the API transcription feature as you requested
- ✅ Keep only the LLM post-processing feature which seems to have more community interest

## Regarding transcribe-rs and future API support:

I checked transcribe-rs and noticed its `openai` feature currently uses a hardcoded enum that only supports 3 OpenAI models (`whisper-1`, `gpt-4o-mini-transcribe`, `gpt-4o-transcribe`). This means providers like Groq and SambaNova won't work out of the box since they use different model names (`whisper-large-v3`, `whisper-large-v3-turbo`).

I'll raise a PR to transcribe-rs to: 
- Allow passing model names as strings instead of the fixed enum
- Add helper configurations for providers like Groq, SambaNova, and others with OpenAI-compatible endpoints

Once that's merged, I can create a proper PR for Handy that uses transcribe-rs for API transcription, following your preferred approach.

## Why this matters to me (and others):

I have a low-end PC where local transcription is quite slow. Providers like Groq and SambaNova offer extremely fast transcription (often under 1 second) via their OpenAI-compatible APIs, which makes Handy usable for people like me who don't have powerful hardware. I believe there are others in the community with similar needs (like the user in #77).

## About AI assistance:

Yes, I used AI tools (Claude opus 4.5) to help implement this.  My intent was to: 
- Enable fast transcription for users with slow hardware
- Add flexible LLM post-processing with multiple providers

I should have been more thoughtful about how this fits with the project's architecture and your vision.

here is the updated ui for post processing
## post proessing disabled:
image

## post processing enabled:
image

## post processing tab:
image

## The history remains same as above

cjpais:
> I have a low-end PC where local transcription is quite slow. Providers like Groq and SambaNova offer extremely fast transcription (often under 1 second) via their OpenAI-compatible APIs, which makes Handy usable for people like me who don't have powerful hardware. I believe there are others in the community with similar needs (like the user in https://github.com/cjpais/Handy/issues/77).

This is fundamentally the same thing as the earlier discussion. My stance has not significantly changed. Please do not submit a PR for it without gathering significant support in Discussions. As well as coming up with a way forward which does not require someone to use an API key. Or hides this functionality in a nice way, which enables power users to use it.

**As for this PR. Please describe what you have done exactly in your own language. No AI descriptions. How does this improve Handy?** I can see some details in the screenshot but I want you to be explicit about it, so I can validate what the code is doing based on what you wanted to do. Post processing is not a 'general' feature right now. There's a reason it's in debug and I will be moving it to a new menu when I think the feature is good enough for prime time. I'm happy to accept other UI/UX improvements to the feature itself in the meantime

avijitbhuin21:
It's okay, since i need this feature desperately i'll modify and build it or my use only. i will close this pr as there is not much left without the online api support. post processing is already in debug state so there is not much point in implementing it. 
Thanks for the feedback man appreciate your time

cjpais:
Sounds good, and if you do want to get the API support in the mainline build please go collect support so we can have an open discussion there!

hdosys:
A useful option for local integration would be an OpenAI-compatible endpoint such as [Speaches](https://github.com/speaches-ai/speaches). Configurable endpoint support would allow users to choose local models and backends without requiring Handy to integrate each model family directly.

In my testing, Parakeet's accuracy on technical vocabulary is lower than that of the Whisper variants I use, while the current Whisper path in Handy has reliability issues. An endpoint abstraction would keep those trade-offs user-selectable.

hdosys:
Here are some other use cases:

- A small company with thin-client hardware could purchase one powerful machine equipped with a GPU, allowing all clients to use it for transcriptions.
- A family could run one GPU server that everyone can access through Tailscale, wherever they are.

hdosys:
A tailnet could be one deployment option for sharing transcription capacity between trusted devices. Although it simplifies private connectivity, security still depends on the tailnet's ACLs, authentication, and endpoint configuration.

Within Handy, the primary requirement would be a configurable API endpoint; users could then choose a local-only, LAN, tailnet, or other authenticated deployment.

cjpais:
Handy is not going to support STT API providers. People are welcome to fork if they want that support. It will eventually be a local provider itself.

avijitbhuin21:
Hi @User-3090 ,

I’ve created **Babbl**:  
https://github.com/avijitbhuin21/Babbl/releases/tag/v0.1.3

Babbl is a fork of **Handy**, extended with API integration and post-processing features.  
So far, my friends and I have been using it with **Groq** for super-fast processing, and it has been working very well.

Please check it out and let me know your thoughts.

Issue #462, opened 2025-12-16
[BUG] Race -> crash


joshribakoff:
## Before You Submit

**Please search [existing issues](https://github.com/cjpais/Handy/issues) to avoid duplicates.** Your bug may already be reported! Right now it's just me maintaining this project so many issues can be overwhelming! Help me out by checking first.

## Bug Description

When quickly toggling the push to talk accidentally, releasing it before it's done processing and pressing it again rapidly and triggering it to process small audio samples, it sometimes causes the entire program to hang with not responding and then you have to force quit. When this happens it's stuck on transcribing UI and the little icon shows the brain. I'm on Parakeet V two which runs nearly instantaneously, so this was definitely a permanent hang and not just slowness.

## System Information

**App Version:** 0.6.8

**Operating System:**

see attached

**CPU:**

**GPU:**

<img width="266" height="551" alt="Image" src="https://github.com/user-attachments/assets/6bfe0da3-1940-4916-8fe0-7a7eea9bf15f" />
<img width="978" height="746" alt="Image" src="https://github.com/user-attachments/assets/699d062d-815d-43f7-82b4-5bc2d4fb01eb" />
<img width="990" height="784" alt="Image" src="https://github.com/user-attachments/assets/52779ded-fdc8-4bee-9206-dc976f1c25e3" />
<img width="1016" height="771" alt="Image" src="https://github.com/user-attachments/assets/616c4be4-f880-4c37-831f-0c82c2bb6872" />
<img width="1000" height="791" alt="Image" src="https://github.com/user-attachments/assets/a3c49efb-cb71-4418-95e6-b022714b75ff" />
<img width="262" height="203" alt="Image" src="https://github.com/user-attachments/assets/e321537f-3f42-4ef7-9eaa-876706cd25a1" />
<img width="393" height="241" alt="Image" src="https://github.com/user-attachments/assets/658730e6-5ffc-48d0-8767-817e17722f15" />
<img width="55" height="32" alt="Image" src="https://github.com/user-attachments/assets/47f9a6fb-d348-453d-8597-948e05a6aa84" />
<img width="157" height="49" alt="Image" src="https://github.com/user-attachments/assets/49b633df-6dc5-485e-b4f0-032d41a5259e" />

cjpais:
I would guess this is a possible duplicate or same bug fix that is needed for #406 

Curious if the same bug shows up for you there too @joshribakoff 

I believe it was pressing and holding the shortcut when not in push to talk mode

[...]

PR #467, opened 2025-12-17
refactoring to avoid deadlocks/crashing

cjpais:
mainly for #462 

needs review and testing.

### DIFF
--- a/src-tauri/src/actions.rs
+++ b/src-tauri/src/actions.rs
@@ -3,6 +3,7 @@ use crate::apple_intelligence;
 use crate::audio_feedback::{play_feedback_sound, play_feedback_sound_blocking, SoundType};
 use crate::managers::audio::AudioRecordingManager;
 use crate::managers::history::HistoryManager;
+use crate::managers::operation::OperationCoordinator;
 use crate::managers::transcription::TranscriptionManager;
 use crate::settings::{get_settings, AppSettings, APPLE_INTELLIGENCE_PROVIDER_ID};
 use crate::shortcut;
@@ -13,7 +14,7 @@ use async_openai::types::{
     CreateChatCompletionRequestArgs,
 };
 use ferrous_opencc::{config::BuiltinConfig, OpenCC};
-use log::{debug, error};
+use log::{debug, error, info, warn};
 use once_cell::sync::Lazy;
 use std::collections::HashMap;
 use std::sync::Arc;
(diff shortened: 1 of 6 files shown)


joshribakoff:
nit: not a refactor, bug fix. not a crash, just a deadlock/hang

Does this throw away the previous in flight transcription if I accidentally queue another? That could potentially trade one class of bug for a more annoying one. I am looking into adding unit tests in a separate PR.

joshribakoff (changes requested):
## Review

I had some time to test this. Here are my notes! Thnx so much for your time, and this great starting point!

- I can replicate the original ~crash~ hang, on demand. Make a long recording (switch to a slow model first, if needed), then deliberately interrupt it to record again, while its transcribing the previous one.
- On `main`, it spins forever -> force quit
- On this branch, the issue is solved
- It adds new issues
- It does not have tests

## Remaining Issues

Now, when I interrupt, I see a "recording" UI flicker for 0.1s, and then vanish.

At this point, there is no UI indicating that it is recording, but I suspect it still was, so I continue speaking out loud.

I pressed the button to start a recording but it actually was already recording, so it stops the recording that had been going on in the background, and spews text unexpectedly from audio I was unaware it was recording.

## Opus 4.5 Explanation:

  actions.rs lines 391-396 - When Operation 1's async task detects it's stale, it calls cleanup which overwrites the UI that Operation 2 just set

## Open Questions:

 I am slightly concerned about landing tests during other active contributions, not sure how we plan to prioritize testing vs adding new features.

I wonder if it needs to account for n-2, in addition to the n-1 recording, etc.

Are we okay with extracting pure functions for the state machine (to facilitate testing). I would advocate for this, once there is a single source of truth and well tested "model", the UI is just a projection of that (and becomes very easy). Essentially loosely following MVC or React like model (one source of state, view layer declaratively renders the state if its different, no imperative updates to UI)

cjpais:
Hey this issue is low priority generally for the moment, and the pr is just scaffolding to remind me to get to it at some point. It's effectively placeholder. It's called a refactor because that's the overall goal of the PR and it happens to fix an issue as well. It's putting a bookmark that I think this code generally needs to be worked on more broadly

Right now just contribute what you can and I'll do my best to pull things in when I get a chance. Scoped changes that don't involve core parts of the app make it in much faster. This touches too much for me to consider it in the short term

Adding tests is going to be best effort and incremental most likely. You can propose whatever changes you would like to see and if they have good logic behind them they will likely be accepted. The app generally runs on community contributions, I don't have a top down or strict plan for anything except stabilizing core features and fixing the biggest outstanding issues. I am much more focused on the inference and distribution side at the moment as well as trying to sort out keybindings across platforms.

If you want to help take on any maintenance and opinions in certain parts of the codebase I am happy to delegate and build trust through quality accepted PRs

joshribakoff:
Thanks, I see related bug reports, so while it may seem low priority its a risk to keep an eye on. Sounds good, I would definitely agree on keeping it incremental and building mutual trust. I do subscribe to different meanings to these words though, a refactor by definition (as defined by Martin Fowler) changes code without affecting behavior, by definition it cannot also fix a bug. If you are changing behavior it should not be called a refactor, and if code has no tests, changing it should not be called refactoring -- according to Martin Fowler.

Splitting hairs aside, sounds like we're aligned on small incremental changes to add testability and go from there, thanks for the reply and sorry for the delay on my end!

cjpais:
Yeah I don't care about semantics of words this much as long as the general point is understood. Language can be flexible

cjpais:
im closing this in favor of #672

PR #473, opened 2025-12-19
fix: prevent crash on macos 26.x beta during startup

VirenMohindra:
## Before Submitting This PR

## Description

defers the apple intelligence availability check from app initialization to when the user actually tries to use the feature.

on macos 26.x (tahoe) beta, accessing `SystemLanguageModel.default` during early app startup (before the cocoa run loop is fully established) causes `SIGABRT`. this only affects fresh installs or users without saved settings.

**changes:**
- remove availability check from `default_post_process_providers()` in settings.rs - always include apple intelligence provider on macos arm64
- add `checkAppleIntelligenceAvailable` tauri command for lazy availability checking
- update frontend to check availability when user selects apple intelligence provider
- show error message if apple intelligence is not available on the device
- move apple intelligence provider before "custom" in the provider list
- add "unavailable" translation key for all 9 supported languages

## Related Issues / Discussions
- fixes startup crash reported by user on discord running macos 26.1 beta (25B78)
- https://discord.com/channels/1391636091853078608/1395928437511684227/1451529733161619587

| Menu Bar      | Post Processing | Apple Intelligence unavailable |
| ----------- | ----------- | ----------- |
| menu      | post-processing       |   demo     |

## Testing

- [x] deleted `~/Library/Application Support/com.pais.handy/settings_store.json` to simulate fresh install
- [x] built and launched app on macos - no crash
- [x] verified apple intelligence provider appears in post-processing settings (before custom)
- [x] tested unavailable error message displays correctly when apple intelligence is not available
- [x] cargo check passes
- [x] lint and format checks pass

### DIFF
--- a/src/components/settings/post-processing/PostProcessingSettings.tsx
+++ b/src/components/settings/post-processing/PostProcessingSettings.tsx
@@ -56,6 +56,14 @@ const PostProcessingSettingsApiComponent: React.FC = () => {
         </div>
       </SettingContainer>
 
+      {state.appleIntelligenceUnavailable ? (
+        <div className="p-3 bg-red-500/10 border border-red-500/50">
+          <p className="text-sm text-red-500">
+            {t("settings.postProcessing.api.appleIntelligence.unavailable")}
+          </p>
+        </div>
+      ) : null}
+
       {state.isAppleProvider ? (
         <SettingContainer
           title={t("settings.postProcessing.api.appleIntelligence.title")}
--- a/.github/workflows/pr-test-build.yml
+++ b/.github/workflows/pr-test-build.yml
@@ -4,7 +4,7 @@ on:
   workflow_dispatch:
     inputs:
       pr_number:
-        description: 'PR number to build'
+        description: "PR number to build"
         required: true
         type: string
 
--- a/src-tauri/src/commands/mod.rs
+++ b/src-tauri/src/commands/mod.rs
@@ -115,3 +115,18 @@ pub fn open_app_data_dir(app: AppHandle) -> Result<(), String> {
 
     Ok(())
 }
+
+/// Check if Apple Intelligence is available on this device.
+/// Called by the frontend when the user selects Apple Intelligence provider.
+#[specta::specta]
+#[tauri::command]
+pub fn check_apple_intelligence_available() -> bool {
+    #[cfg(all(target_os = "macos", target_arch = "aarch64"))]
+    {
+        crate::apple_intelligence::check_apple_intelligence_availability()
+    }
+    #[cfg(not(all(target_os = "macos", target_arch = "aarch64")))]
+    {
+        false
+    }
+}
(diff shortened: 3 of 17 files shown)


src/components/settings/post-processing/PostProcessingSettings.tsx
```diff
@@ -56,6 +56,14 @@ const PostProcessingSettingsApiComponent: React.FC = () => {
         </div>
       </SettingContainer>
 
+      {state.appleIntelligenceUnavailable ? (
```
VirenMohindra, line 59: so alert() doesn't work in tauri-land. resorting to show an inline alert https://mui.com/material-ui/react-alert/

will eventually make this into a component once we create a project from https://github.com/cjpais/Handy/discussions/449

cjpais:
Thanks for this. I'm going to take a look soon. Uh hopefully I'll get a chance tomorrow and hopefully pull this in as well.

cjpais:
Hey @VirenMohindra this is great. I did play around with it a little, and in the dropdown for the UI what is the expectation? I notice when I click "Apple Intelligence" with the feature off it sets Custom instead, but Apple Intelligence warning comes comes up

Screenshot 2025-12-22 at 10 05 18 AM

VirenMohindra:
> I notice when I click "Apple Intelligence" with the feature off it sets Custom instead, but Apple Intelligence warning comes comes up

the issue was that clicking the currently selected provider (e.g. custom) after the error appeared wouldn't clear it -- the early return happened before the state reset

fixed it by moving the clear to before the early return. now clicking any option in the dropdown will dismiss the error

cjpais:
The behavior is still weird. When I click the Apple Intelligence with it disabled it still changes the option to something else. I feel like it should just show the banner?

To replicate:
* Set model provider to anything other than Apple Intelligence
* Disable Apple Intelligence
* Click "Apple Intelligence" in the dropdown of Handy

The provider will show something other than "Apple Intelligence" in the dropdown, and the API key and other settings for that model provider.

My expectation is as a user I clicked the "Apple Intelligence" in the dropdown, so it should still be selected as "Apple Intelligence" and then display the error or warning only for that dropdown item

VirenMohindra:
> The behavior is still weird. When I click the Apple Intelligence with it disabled it still changes the option to something else. I feel like it should just show the banner?
> The provider will show something other than "Apple Intelligence" in the dropdown, and the API key and other settings for that model provider.
> My expectation is as a user I clicked the "Apple Intelligence" in the dropdown, so it should still be selected as "Apple Intelligence" and then display the error or warning only for that dropdown item

fixed! now when clicking "Apple Intelligence" with the feature unavailable the dropdown shows "apple intelligence" as selected (reflects what user clicked and the error banner appears below explaining it's not available

cjpais:
I'm merging this for now, but the UI still needs work. Let's follow up with another PR if possible. 

Let's hide the dropdown for the model for Apple Intelligence as well as remove the UI Infobox for "Apple Intelligence" since it's largely N/A now, or move the error state directly into that.

Issue #475, opened 2025-12-19
[BUG] Handy crashes when no mic is connected


bauerpl:
## Bug Description

Handy crashes when the `push to talk` shortcut is used on a Mac Mini without a mic connected.

## System Information

**App Version:**
`v0.6.8`

**Operating System:**

`MacOS Sequoia 15.7.2 (24G325)`

**CPU:**

`Apple M2 Pro`

**GPU:**

`Apple M2 Pro GPU`

cjpais:
Thanks for reporting this, this is a good bug and probably some bad error handling in the code.

I will try to ship a patch (#477) in the next version, and I've noted down a bigger issue to propagate errors to the UI better. I think 0.6.9 will come out soon. There's one more PR I want to land before shipping it

cjpais:
@bauerpl @noelboss @mane81 does the latest version 0.7.10 fix this for you?

mane81:
> [@bauerpl](https://github.com/bauerpl) [@noelboss](https://github.com/noelboss) [@mane81](https://github.com/mane81) does the latest version 0.7.10 fix this for you?
Works like a charm, thank you!

bauerpl:
Yes, thank you!

PR #477, opened 2025-12-20
dont crash if there's no mic

cjpais:
title

### DIFF
--- a/src-tauri/src/audio_toolkit/audio/recorder.rs
+++ b/src-tauri/src/audio_toolkit/audio/recorder.rs
@@ -76,8 +76,13 @@ impl AudioRecorder {
         let level_cb = self.level_cb.clone();
 
         let worker = std::thread::spawn(move || {
-            let config = AudioRecorder::get_preferred_config(&thread_device)
-                .expect("failed to fetch preferred config");
+            let config = match AudioRecorder::get_preferred_config(&thread_device) {
+                Ok(c) => c,
+                Err(e) => {
+                    log::error!("Failed to get audio config: {}", e);
+                    return;
+                }
+            };
 
             let sample_rate = config.sample_rate().0;
             let channels = config.channels() as usize;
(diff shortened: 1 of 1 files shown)


VirenMohindra:
this prevents the crash from #475, which is good. however it introduces silent failures -- if there's no mic, the app won't crash but also _won't_ tell the user anything. they'll press push-to-talk and nothing happens

additionally, if the worker exits early during init, calling `stop()` later will deadlock because it waits on a response channel from a dead thread

i think we should validate the audio config in `open()` before spawning the worker thread, so failures are caught and returned to the caller. and this also lets the UI show an appropriate error message

src-tauri/src/audio_toolkit/audio/recorder.rs
```diff
@@ -76,8 +76,13 @@ impl AudioRecorder {
         let level_cb = self.level_cb.clone();
 
         let worker = std::thread::spawn(move || {
-            let config = AudioRecorder::get_preferred_config(&thread_device)
-                .expect("failed to fetch preferred config");
+            let config = match AudioRecorder::get_preferred_config(&thread_device) {
+                Ok(c) => c,
+                Err(e) => {
+                    log::error!("Failed to get audio config: {}", e);
+                    return;
```
VirenMohindra, line 83: this prevents the crash but now `open()` returns `Ok(())` even when the worker immediately exits. the caller (audio manager) sets `is_open = true` and _thinks_ everything is fine

maybe we do config validation in `open()` before spawning the worker, or use a oneshot channel to communicate init success / failure back

src-tauri/src/audio_toolkit/audio/recorder.rs
```diff
@@ -93,28 +98,37 @@ impl AudioRecorder {
...
-            stream.play().expect("failed to start stream");
+            let stream = match stream {
+                Ok(s) => s,
+                Err(e) => {
+                    log::error!("Failed to build audio stream: {}", e);
+                    return;
+                }
+            };
+
+            if let Err(e) = stream.play() {
```
VirenMohindra, line 128: if the worker exits here ^, `stop()` will deadlock, and it sends `Cmd::Stop(resp_tx)` to a dead worker and `resp_rx.recv()` blocks forever waiting for a response that never comes

VirenMohindra: think we should not let `open()` succeed if init fails

cjpais:
Thanks for the review @VirenMohindra, I definitely want to consider a toasting ui of some kind for error handling generally. I think probably having some unified way to throw up to the UI would be helpful for big errors

Tbh I just let Claude write whatever and hadn't even reviewed this, if you would like to submit changes and improvements I would be overjoyed!! I already have a huge backlog and would love to focus on keyboard input stuff instead

VirenMohindra:
>  if you would like to submit changes and improvements

happy to! def want to clear out my in flight PRs before i jump on this one though, before they get too stale 😅

cjpais:
@VirenMohindra i will try and get them in soon! I'm not planning to pull much else in for a week or two. I'm trying to wrap up a rewrite of the keyboard input library generally

cjpais:
@VirenMohindra i will try and get them in soon! I'm not planning to pull much else in for a week or two. I'm trying to wrap up a rewrite of the keyboard input library generally

PR #478, opened 2025-12-20
feat: add models settings page with filtering and management

VirenMohindra:
## Before Submitting This PR

## Description

adds a dedicated models settings page accessible from the sidebar, allowing users to view, download, switch, and delete transcription models without going through onboarding.

**changes:**
- add models settings page in sidebar with filter buttons (all/multi-language/translation)
- reuse `ModelCard` component from onboarding for consistent UI
- add model deletion with native confirmation dialog using `tauri-plugin-dialog`
- consolidate all model event listeners into `useModels` hook (removes duplicates from ModelSelector, LanguageSelector, TranslateToEnglish)
- gate `AccessibilityPermissions` component to macOS only (was rendering on all platforms)
- add `is_recommended` field to model registry for featured models
- remove unused `get_recommended_first_model` rust command (dead code)
- add translations for all 9 supported languages

## Related Issues/Discussions

- related to model management improvements
- model registry discussion from https://github.com/cjpais/Handy/discussions/449

## Testing

- [x] verified model download/switch/delete works from new settings page
- [x] verified filter buttons correctly filter models by capability
- [x] verified delete confirmation dialog appears (native macOS dialog)
- [x] verified no duplicate event listeners (checked via code review)
- [x] verified lint passes
- [x] verified all translations present in 9 languages

n/a - UI reuses existing ModelCard component

| state | demo      | 
| ----------- | ----------- | 
| Model Section | models-section      | 
| Model dropdown | models-dropdown   |
| Dropdown downloading | models-dropdown-downloading   |
| Model Section "Extracting..." | model-section-extracting   |
| Model auto switch after downloading | model-auto-switch-after-downloading   |
| Delete confirmation | model-delete-confirm   |

| whisper      | parakeet |
| ----------- | ----------- |
| parakeet      | whisper       |

### DIFF
--- a/src/stores/modelStore.ts
+++ b/src/stores/modelStore.ts
@@ -355,6 +355,17 @@ export const useModelStore = create<ModelsStore>()(
         },
       );
 
+      listen<string>("model-download-cancelled", (event) => {
+        const modelId = event.payload;
+        set(
+          produce((state) => {
+            delete state.downloadingModels[modelId];
+            delete state.downloadProgress[modelId];
+            delete state.downloadStats[modelId];
+          }),
+        );
+      });
+
       listen<string>("model-deleted", () => {
         get().loadModels();
         get().loadCurrentModel();
--- a/src/components/model-selector/ModelDropdown.tsx
+++ b/src/components/model-selector/ModelDropdown.tsx
@@ -25,7 +25,7 @@ const ModelDropdown: React.FC<ModelDropdownProps> = ({
   };
 
   return (
-    <div className="absolute bottom-full left-0 mb-2 w-64 max-h-[60vh] overflow-y-auto bg-background border border-mid-gray/20 rounded-lg shadow-lg py-2 z-50">
+    <div className="absolute bottom-full start-0 mb-2 w-64 max-h-[60vh] overflow-y-auto bg-background border border-mid-gray/20 rounded-lg shadow-lg py-2 z-50">
       {downloadedModels.length > 0 ? (
         <div>
           {downloadedModels.map((model) => (
--- a/src/components/model-selector/ModelSelector.tsx
+++ b/src/components/model-selector/ModelSelector.tsx
@@ -1,9 +1,9 @@
 import React, { useState, useRef, useEffect } from "react";
 import { useTranslation } from "react-i18next";
 import { listen } from "@tauri-apps/api/event";
-import { produce } from "immer";
-import { commands, type ModelInfo } from "@/bindings";
+import { commands } from "@/bindings";
 import { getTranslatedModelName } from "../../lib/utils/modelTranslation";
+import { useModelStore } from "../../stores/modelStore";
 import ModelStatusButton from "./ModelStatusButton";
 import ModelDropdown from "./ModelDropdown";
 import DownloadProgressDisplay from "./DownloadProgressDisplay";
@@ -15,13 +15,6 @@ interface ModelStateEvent {
   error?: string;
 }
 
-interface DownloadProgress {
-  model_id: string;
-  downloaded: number;
-  total: number;
-  percentage: number;
-}
-
 type ModelStatus =
   | "ready"
   | "loading"
@@ -31,46 +24,59 @@ type ModelStatus =
   | "unloaded"
   | "none";
 
-interface DownloadStats {
-  startTime: number;
-  lastUpdate: number;
-  totalDownloaded: number;
-  speed: number;
-}
-
 interface ModelSelectorProps {
   onError?: (error: string) => void;
 }
 
 const ModelSelector: React.FC<ModelSelectorProps> = ({ onError }) => {
   const { t } = useTranslation();
-  const [models, setModels] = useState<ModelInfo[]>([]);
-  const [currentModelId, setCurrentModelId] = useState<string>("");
+  const {
+    models,
+    currentModel,
+    downloadProgress,
+    downloadStats,
+    extractingModels,
+    selectModel,
+  } = useModelStore();
+
   const [modelStatus, setModelStatus] = useState<ModelStatus>("unloaded");
   const [modelError, setModelError] = useState<string | null>(null);
-  const [modelDownloadProgress, setModelDownloadProgress] = useState<
-    Record<string, DownloadProgress>
-  >({});
   const [showModelDropdown, setShowModelDropdown] = useState(false);
-  const [downloadStats, setDownloadStats] = useState<
-    Record<string, DownloadStats>
-  >({});
-  const [extractingModels, setExtractingModels] = useState<
-    Record<string, true>
-  >({});
+  // Track pending model switch for optimistic display
+  const [pendingModelId, setPendingModelId] = useState<string | null>(null);
 
   const dropdownRef = useRef<HTMLDivElement>(null);
 
-  useEffect(() => {
-    loadModels();
-    loadCurrentModel();
+  const displayModelId = pendingModelId || currentModel;
 
-    // Listen for model state changes
+  // Check model status when currentModel changes
+  useEffect(() => {
+    const checkStatus = async () => {
+      if (currentModel) {
+        try {
+          const statusResult = await commands.getTranscriptionModelStatus();
+          if (statusResult.status === "ok") {
+            setModelStatus(
+              statusResult.data === currentModel ? "ready" : "unloaded",
+            );
+          }
+        } catch {
+          setModelStatus("error");
+          setModelError("Failed to check model status");
+        }
+      } else {
+        setModelStatus("none");
+      }
+    };
+    checkStatus();
+  }, [currentModel]);
+
+  useEffect(() => {
+    // Listen for model loading lifecycle events
     const modelStateUnlisten = listen<ModelStateEvent>(
       "model-state-changed",
       (event) => {
-        const { event_type, model_id, model_name, error } = event.payload;
-
+        const { event_type, error } = event.payload;
         switch (event_type) {
           case "loading_started":
             setModelStatus("loading");
@@ -79,11 +85,12 @@ const ModelSelector: React.FC<ModelSelectorProps> = ({ onError }) => {
           case "loading_completed":
             setModelStatus("ready");
             setModelError(null);
-            if (model_id) setCurrentModelId(model_id);
+            setPendingModelId(null);
             break;
           case "loading_failed":
             setModelStatus("error");
             setModelError(error || "Failed to load model");
+            setPendingModelId(null);
             break;
           case "unloaded":
             setModelStatus("unloaded");
@@ -93,170 +100,29 @@ const ModelSelector: React.FC<ModelSelectorProps> = ({ onError }) => {
       },
     );
 
-    // Listen for model download progress
-    const downloadProgressUnlisten = listen<DownloadProgress>(
-      "model-download-progress",
-      (event) => {
-        const progress = event.payload;
-        setModelDownloadProgress(
-          produce((downloadProgress) => {
-            downloadProgress[progress.model_id] = progress;
-          }),
-        );
-        setModelStatus("downloading");
-
-        // Update download stats for speed calculation
-        const now = Date.now();
-        setDownloadStats(
-          produce((stats) => {
-            const current = stats[progress.model_id];
-
-            if (!current) {
-              // First progress update - initialize
-              stats[progress.model_id] = {
-                startTime: now,
-                lastUpdate: now,
-                totalDownloaded: progress.downloaded,
-                speed: 0,
-              };
-            } else {
-              // Calculate speed over last few seconds
-              const timeDiff = (now - current.lastUpdate) / 1000; // seconds
-              const bytesDiff = progress.downloaded - current.totalDownloaded;
-
-              if (timeDiff > 0.5) {
-                // Update speed every 500ms
-                const currentSpeed = bytesDiff / (1024 * 1024) / timeDiff; // MB/s
-                // Smooth the speed with exponential moving average, but ensure positive values
-                const validCurrentSpeed = Math.max(0, currentSpeed);
-                const smoothedSpeed =
-                  current.speed > 0
-                    ? current.speed * 0.8 + validCurrentSpeed * 0.2
-                    : validCurrentSpeed;
-
-                stats[progress.model_id] = {
-                  startTime: current.startTime,
-                  lastUpdate: now,
-                  totalDownloaded: progress.downloaded,
-                  speed: Math.max(0, smoothedSpeed),
-                };
-              }
-            }
-          }),
-        );
-      },
-    );
-
-    // Listen for model download completion
+    // Auto-select model when download completes (fires after extraction too)
     const downloadCompleteUnlisten = listen<string>(
       "model-download-complete",
       (event) => {
         const modelId = event.payload;
-        setModelDownloadProgress(
-          produce((progress) => {
-            delete progress[modelId];
-          }),
-        );
-        setDownloadStats(
-          produce((stats) => {
-            delete stats[modelId];
-          }),
-        );
-        loadModels(); // Refresh models list
-
-        // Auto-select the newly downloaded model (skip if recording in progress)
         setTimeout(async () => {
-          const isRecording = await commands.isRecording();
-          if (isRecording) {
-            return; // Skip auto-switch if recording in progress
+          try {
+            const isRecording = await commands.isRecording();
+            if (!isRecording) {
+              setPendingModelId(modelId);
+              setModelError(null);
+              setShowModelDropdown(false);
+              const success = await selectModel(modelId);
+              if (!success) {
+                setPendingModelId(null);
+              }
+            }
+          } catch {
+            // Ignore errors in auto-select
           }
-          loadCurrentModel();
-          handleModelSelect(modelId);
         }, 500);
       },
     );
-
-    // Listen for model download cancellation
-    const downloadCancelledUnlisten = listen<string>(
-      "model-download-cancelled",
-      (event) => {
-        const modelId = event.payload;
-        setModelDownloadProgress(
-          produce((progress) => {
-            delete progress[modelId];
-          }),
-        );
-        setDownloadStats(
-          produce((stats) => {
-            delete stats[modelId];
-          }),
-        );
-        // Reset status if no other downloads in progress
-        setModelDownloadProgress((currentProgress) => {
-          if (Object.keys(currentProgress).length === 0) {
-            loadCurrentModel(); // Restore normal status display
-          }
-          return currentProgress;
-        });
-      },
-    );
-
-    // Listen for extraction events
-    const extractionStartedUnlisten = listen<string>(
-      "model-extraction-started",
-      (event) => {
-        const modelId = event.payload;
-        setExtractingModels(
-          produce((extracting) => {
-            extracting[modelId] = true;
-          }),
-        );
-        setModelStatus("extracting");
-      },
-    );
-
-    const extractionCompletedUnlisten = listen<string>(
-      "model-extraction-completed",
-      (event) => {
-        const modelId = event.payload;
-        setExtractingModels(
-          produce((extracting) => {
-            delete extracting[modelId];
-          }),
-        );
-        loadModels(); // Refresh models list
-
-        // Auto-select the newly extracted model (skip if recording in progress)
-        setTimeout(async () => {
-          const isRecording = await commands.isRecording();
-          if (isRecording) {
-            return; // Skip auto-switch if recording in progress
-          }
-          loadCurrentModel();
-          handleModelSelect(modelId);
-        }, 500);
-      },
-    );
-
-    const extractionFailedUnlisten = listen<{
-      model_id: string;
-      error: string;
-    }>("model-extraction-failed", (event) => {
-      const modelId = event.payload.model_id;
-      setExtractingModels(
-        produce((extracting) => {
-          delete extracting[modelId];
-        }),
-      );
-      setModelError(`Failed to extract model: ${event.payload.error}`);
-      setModelStatus("error");
-    });
-
-    // Listen for model deletion
-    const modelDeletedUnlisten = listen<string>("model-deleted", () => {
-      loadModels(); // Refresh models list
-      loadCurrentModel(); // Update current model in case deleted model was active
-    });
 
     // Click outside to close dropdown
     const handleClickOutside = (event: MouseEvent) => {
@@ -273,78 +139,21 @@ const ModelSelector: React.FC<ModelSelectorProps> = ({ onError }) => {
     return () => {
       document.removeEventListener("mousedown", handleClickOutside);
       modelStateUnlisten.then((fn) => fn());
-      downloadProgressUnlisten.then((fn) => fn());
       downloadCompleteUnlisten.then((fn) => fn());
-      downloadCancelledUnlisten.then((fn) => fn());
-      extractionStartedUnlisten.then((fn) => fn());
-      extractionCompletedUnlisten.then((fn) => fn());
-      extractionFailedUnlisten.then((fn) => fn());
-      modelDeletedUnlisten.then((fn) => fn());
     };
-  }, []);
-
-  const loadModels = async () => {
-    try {
-      const result = await commands.getAvailableModels();
-      if (result.status === "ok") {
-        setModels(result.data);
-      }
-    } catch (err) {
-      console.error("Failed to load models:", err);
-    }
-  };
-
-  const loadCurrentModel = async () => {
-    try {
-      const result = await commands.getCurrentModel();
-      if (result.status === "ok") {
-        const current = result.data;
-        setCurrentModelId(current);
-
-        if (current) {
-          // Check if model is actually loaded
-          const statusResult = await commands.getTranscriptionModelStatus();
-          if (statusResult.status === "ok") {
-            const transcriptionStatus = statusResult.data;
-            if (transcriptionStatus === current) {
-              setModelStatus("ready");
-            } else {
-              setModelStatus("unloaded");
-            }
-          }
-        } else {
-          setModelStatus("none");
-        }
-      }
-    } catch (err) {
-      console.error("Failed to load current model:", err);
-      setModelStatus("error");
-      setModelError("Failed to check model status");
-    }
-  };
+  }, [selectModel]);
 
   const handleModelSelect = async (modelId: string) => {
-    try {
-      setCurrentModelId(modelId); // Set optimistically so loading text shows correct model
-      setModelError(null);
-      setShowModelDropdown(false);
-      const result = await commands.setActiveModel(modelId);
-      if (result.status === "error") {
-        const errorMsg = result.error;
-        setModelError(errorMsg);
-        setModelStatus("error");
-        onError?.(errorMsg);
-      }
-    } catch (err) {
-      const errorMsg = `${err}`;
-      setModelError(errorMsg);
+    setPendingModelId(modelId);
+    setModelError(null);
+    setShowModelDropdown(false);
+    const success = await selectModel(modelId);
+    if (!success) {
+      setPendingModelId(null);
       setModelStatus("error");
-      onError?.(errorMsg);
+      setModelError("Failed to switch model");
+      onError?.("Failed to switch model");
     }
-  };
-
-  const getCurrentModel = () => {
-    return models.find((m) => m.id === currentModelId);
   };
 
   const getModelDisplayText = (): string => {
@@ -364,7 +173,7 @@ const ModelSelector: React.FC<ModelSelectorProps> = ({ onError }) => {
       }
     }
 
-    const progressValues = Object.values(modelDownloadProgress);
+    const progressValues = Object.values(downloadProgress);
     if (progressValues.length > 0) {
       if (progressValues.length === 1) {
         const progress = progressValues[0];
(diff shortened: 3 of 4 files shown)


cjpais:
From first glance I'm digging this, I'd like to review around how model specific settings potentially overlap with this as well

dannysmith:
Love this at first glance.

(Observation on looking screenshots: It might look nicer if the download/delete buttons on the ModelCards were to the left of the accuracy/speed bars. The bars are always present so should be consistently positioned?)

VirenMohindra:
> I'd like to review around how model specific settings potentially overlap with this as well

can you expand on that a bit? is there anything on the roadmap that could build on this? my goal was to put a solid foundation in place for future models -- things like tagging, filtering for now, and eventually search once the list grows

if there’s something that makes sense to support now while i’m already working in this area, happy to add it

> It might look nicer if the download/delete buttons on the ModelCards were to the left of the accuracy/speed bars

yeah, i experimented with that briefly, and i _think_ you’re pointing out that downloading a model causes the accuracy / speed bars to shift, which isn’t _ideal_

at the same time, i was trying to keep things consistent between the model dropdown and the model selection tab, and we _generally_ place destructive actions on the far right

open to alternative layouts if you have something specific in mind

cjpais:
@VirenMohindra regarding

> can you expand on that a bit? is there anything on the roadmap that could build on this?

Yeah definitely. Right now there's a bit of awkwardness in that whisper supports translation built in as well as language selection. But parakeet the language is always "auto" and there's no translation built in. So it's just trying to disambiguate this in the UI since I think it could be confusing for some users. I think we just disable those settings for parakeet, which is okay but it's not super obvious why they are disabled

I was considering making them invisible to the UI for parakeet, but just curious about feedback or other potential UX ideas around this

VirenMohindra:
@cjpais that makes sense. current state: settings are disabled with explanatory text, but that can feel clunky

i went ahead and implemented a contextual approach in this PR — there's now a "model settings" card in general settings that adapts based on the current model:
- whisper models show language selector + translation toggle (when supported)
- parakeet models show "this model works automatically with no configuration needed"

this way users only see settings that apply to their model, no more confusing disabled / grayed out options. let me know if you'd prefer a different approach or want any adjustments


| whisper      | parakeet |
| ----------- | ----------- |
| parakeet      | whisper       |

cjpais:
Wow this is much nicer. Thank you @VirenMohindra 

I don't think I'll be able to review this for a bit but I think we should try and have this be part of a v0.7.0 release

I don't know how reasonable it would be, but it would be very cool to have some filtering on language or something like that in the model page too. So you could pick your specific language and make sure the model you're using supports that language. Mainly I know some people have had problems in the past creating issues for example that Parakeet doesn't support their language well

VirenMohindra:
> Wow this is much nicer. Thank you @VirenMohindra 

❤️❤️

> I don't think I'll be able to review this for a bit but I think we should try and have this be part of a v0.7.0 release

no worries. def would love a couple more eyes on this since it's a thicc one 

> I don't know how reasonable it would be, but it would be very cool to have some filtering on language or something like that in the model page too. So you could pick your specific language and make sure the model you're using supports that language. Mainly I know some people have had problems in the past creating issues for example that Parakeet doesn't support their language well

Great idea -- I actually played around with this but didn't want to blow out scope since we're already crossed a thousand lines of code. happy to take a peek and add it back in, or have a quick follow up PR

what do you think?

cjpais:
Since it's already large, I'm happy to just do a full review of the feature as a whole here I think. I pretty much know I'm going to have to sit down with it already so a couple hundred lines more of code will probably be best now rather than later. I'll probably have some more detailed feedback at some point as well

VirenMohindra:
done! added the ability to filter by language. opted to keep it on the frontend since we already had access to `LANGUAGES`. still relies on the BE via `supports_language_selection` flag

| all languages      | simplified chinese filtered |
| ----------- | ----------- |
| Screenshot 2026-01-01 at 10 59 10 PM      | Screenshot 2026-01-01 at 10 59 13 PM       |

cc @cjpais for vis

cjpais:
Thank you for the continued work on this. I'm quite excited to bring this in. I really need to play with it myself. It's been a long time and I don't want to let this get too stale. I'm really busy, but maybe within the next week or two I can make some more concrete decisions. It is quite a a big change, so it probably will take a bit longer for me to ensure this, but I'm liking generally where it's at and I can probably make some tweaks myself to clean anything else up.

VirenMohindra:
@cjpais thanks for the update! no rush at all - take the time you need to review thoroughly. the feature is working well and we've addressed the cancel functionality issues mentioned [here](https://github.com/cjpais/Handy/discussions/519). happy to make any adjustments based on your feedback when you get a chance to test it out

scripts/check-translations.cjs
VirenMohindra, line 1: realized we're adding support for new languages every couple days and keeping 'em at parity is important, so maybe a CI / CD action to block PRs until we check translations are consistent with the english one might be a useful addition

can move this to a separate PR tho if needed

VirenMohindra:
## Code Review - PR #478: Models Settings Page

Hi @cjpais! thanks for the continued work on this feature. i've done a thorough review to help make your evaluation easier before merging to v0.7.0

---

### 🎯 **High-Level Summary**

This PR delivers a dedicated Models Settings page with filtering, download management, and cancellation support. Beyond the core feature, it includes **3 critical bug fixes** discovered during implementation, while iterating, plus tooling improvements for translation consistency

**Bottom line:** Based on AI assistance and skimming the code on GitHub myself, the architecture is solid, follows existing patterns, and the bug fixes address real user pain points (UI freezes, ghost downloads). code is maintainable and well-structured. There are a few areas worth discussion regarding complexity vs. simplicity trade-offs

---

### 📦 **What's Inside (by commit)** (6 Commits)

**Feature Evolution:**
1. 45f3315 - core Models Settings page with capability filtering (all / multi-language / translation)
2. b790f55 - language filtering with searchable dropdown (100+ languages)
3. 2edd153 - download cancellation support (`Arc<AtomicBool>` mechanism, **partial file preservation**)

**Critical Bug Fixes:**
4. fd87206 - translation consistency checker (prevents incomplete i18n at scale)
5. 92a022b - double-click cancel race condition fix (uses `downloadingModels` set as source of truth)
6. 30fdbc9 - download event throttling + intelligent state sync (fixes UI freeze, prevents auto-resume bugs)

### 🏗️ **Architecture Overview**

**Code Quality & Maintainability:**
- **Follows existing patterns:** Manager -> Command -> Store -> Component flow (consistent with codebase)
- **Component reuse:** `ModelCard` from onboarding reused in settings (DRY principle)
- **Consolidated event handling:** `useModels` hook eliminates duplicate listeners from 3+ components
- **Clean separation:** UI logic in components, business logic in store, system calls in Rust backend

**New Components:**
- `ModelsSettings` - main settings page with dual filtering system
- `modelsStore` - zustand store managing download lifecycle (**critical file**)
- `ModelSettingsCard` - contextual settings adapting to model capabilities

**Backend Changes (Rust):**
- Cancellation via `Arc<AtomicBool>` flags (thread-safe, lock-free, minimal overhead)
- Event throttling (100ms intervals) prevents UI saturation
- Thread-safe state management with Mutex-protected updates

### 🚀 **Performance Improvements**

**UI Freeze Fix (Commit 30fdbc9):**
- **Problem:** large model downloads (>1GB) emitted hundreds of progress events per second, freezing the UI completely
- **Solution:** throttle to 100ms intervals (max 10 events / sec)
- **Impact:** 100x reduction in event traffic, smooth UI during downloads
- **Trade-off:** progress updates slightly chunkier (10/sec vs real-time) but imperceptible to users

**State Management Optimization:**
- intelligent state merging prevents unnecessary re-renders
- `downloadingModels` set provides O(1) lookups vs O(n) array scans
- zustand `subscribeWithSelector` prevents cascading re-renders

### 🎨 **User-Facing Changes (hopefully for `v0.7.0`)**

**New Features:**
1. **Dedicated Models Settings** - accessible from sidebar, no longer hidden in dropdown
2. **Advanced Filtering** - filter by capabilities (multi-language support, translation) + language support
3. **Download Management** - cancel in-progress downloads, resume from partial files
4. **Model-Specific Settings** - UI adapts based on model capabilities (addresses your Dec 22 feedback)

**Bug Fixes Users Will Notice**
- no more UI freezes during model downloads - existing issue iirc
- download cancellation actually works (not just visual feedback)
- progress bars update smoothly without stuttering

**What Won't Change:**
- model switching still works the same way
- existing downloads are preserved (partial files compatible)
- all existing functionality remains intact

### 🔍 **Areas for Discussion**

#### 1. **State Merging Complexity** (`modelsStore.ts` lines 82-105)
the store uses intelligent merging to prevent race conditions between frontend and backend state. this adds a bit of complexity but solves 2 specific bugs (ghost downloads, UI flicker)

**question:** is this complexity justified, or should we accept a simpler approach with ~200ms delay for backend confirmation?

**my take:** the complexity is localized to one function and prevents user-visible bugs. worth it for the UX benefit

#### 2. **Event Throttling Interval** (100ms)
testing showed `50ms` = occasional lag, `200ms` = visibly choppy. `100ms` balanced responsiveness and performance

**question:** should this be configurable, or is 100ms a good default for all scenarios?

**my take:** `100ms` works well across devices (tested on MacOS and Windows). making it configurable adds complexity for unclear benefit

#### 3. **Translation Coverage** (12 languages, 36 new keys)
all translations AI assisted per project convention. new CI tooling catches missing keys automatically and blocks PRs

**question:** are you comfortable with the machine translations for initial release? native speakers can improve later

**my take:** structural consistency is more important than perfect translations initially. tooling prevents incomplete coverage

#### 4. **Removed Dead Code**
- Deleted `get_recommended_first_model` command (unused)
- Consolidated duplicate event listeners

**question:** any concerns about removed code? all deletions were verified unused

**my take:** cleanup improves maintainability

### **What I Verified**

**Code Quality:**
- ci / cd passing, rebased
- no new external dependencies (except std library imports)
- follows project conventions (`i18n` enforcement, component patterns)

**Functionality:**
- download / cancel / delete flows work correctly
- filtering (capability + language) works as expected
- model-specific settings adapt to capabilities
- no duplicate event listeners (code inspection)
- **partial file resume works after cancel** - pretty cool feature imo, might have some overhead but i believe its worth it

**Edge Cases:**
- cancel during extraction phase handled gracefully
- rapid cancel-download-cancel clicks debounced properly
- backend / frontend state drift prevented via intelligent merging
- network timeout + cancel don't leave zombie processes

### 📝 **Recommendation**

**ready to merge with optional discussion on the complexity trade-offs above.** rhe core implementation is production-quality, the bug fixes are valuable independent of the feature, and the architecture is maintainable

**Suggested Review Path**
1. **Quick wins**: skim fd87206 (tooling), b790f55 (language filter) - straightforward additions
2. **Core feature**: review 45f3315 + screenshots in PR description - main functionality
3. **Critical sections**: deep dive into~
   - `modelsStore` L82-105 (state merging - **most complex** logic)
   - `model` `Arc<AtomicBool>` cancellation mechanism implementation
4. **Bug fixes**: 92a022b and 30fdbc9 with context from this summary

i've also posted 6 inline comments on specific code sections highlighting architectural decisions, trade-offs, and areas that might warrant tweaking. these are meant to be educational rather than blocking - they provide context if you want to make adjustments

lmk if you'd like me to clarify any architectural decisions or provide more detail on specific areas!

src/stores/modelStore.ts
```diff
@@ -0,0 +1,383 @@
...
+        const result = await commands.getAvailableModels();
+        if (result.status === "ok") {
+          set({ models: result.data, error: null });
+
+          // Sync downloading state from backend, but merge with current state
+          // to avoid race conditions during download start
+          set((state) => {
+            const backendDownloading = new Set(
+              result.data.filter((m) => m.is_downloading).map((m) => m.id),
+            );
```
VirenMohindra: **Intelligent State Merging Strategy**

this merging logic prevents race conditions between frontend and backend during download operations. it's the most complex code in the PR, so worth explaining the rationale

**the problem**
when a user clicks "Download", FE optimistically adds the model to `downloadingModels` Set for immediate UI feedback. however, the backend takes around 50-100ms to start the download and set `is_downloading: true`. If `loadModels()` is called during this window (triggered by other events), naive state sync would~
1. see backend doesn't show `is_downloading: true` yet
2. replace `downloadingModels` Set with empty Set
3. UI flickers (download button reappears)
4. user might click again, causing a double-download

**the solution (L82-105)**

```tsx
  // Merge intelligently instead of replacing
  const backendDownloading = new Set(...); // What backend says
  const merged = new Set(state.downloadingModels); // Start with frontend state

  // Add models backend confirms are downloading
  backendDownloading.forEach((id) => merged.add(id));

  // Only remove if **BOTH** conditions met:
  state.downloadingModels.forEach((id) => {
    if (!backendDownloading.has(id) && !state.downloadProgress.has(id)) {
      merged.delete(id); // Backend says no + no progress data = truly done
    }
  });
```

**why `downloadProgress` Map matters**
the `Map` serves as a secondary source of truth during transitions. If we have progress data for a model, it's definitely downloading, even if backend state hasn't synced yet

alternative considered~
- delay frontend state update until backend confirms (adds 50-100ms perceived lag). Rejected for UX reasons

maintainability~
- this complexity is localized to one function `loadModels`. the rest of the codebase uses the clean `downloadingModels` Set API. if we remove this, the race condition bugs returns (tested)

performance impact~
- negligible - Set operations are O(1), runs only on backend sync events and not per-render

qq - would you prefer a simpler approach that accepts the 50-100ms lag, or is this complexity justified for the snappier UX?

src-tauri/src/managers/model.rs
```diff
@@ -376,6 +401,13 @@ impl ModelManager {
             }
         }
 
+        // Create cancellation flag for this download
+        let cancel_flag = Arc::new(AtomicBool::new(false));
+        {
+            let mut flags = self.cancel_flags.lock().unwrap();
+            flags.insert(model_id.to_string(), cancel_flag.clone());
+        }
```
VirenMohindra, line 445: **Download Cancellation Architecture**

this implements thread-safe download cancellation using `Arc<AtomicBool>` - a _common_ Rust pattern for cooperative cancellation

why `Arc`?
- `Arc` (Atomic Reference Counted): allows cloning the flag for use across threads without copying
- `AtomicBool`: lock-free boolean that can be read/written from multiple threads safely
- `Ordering::Relaxed`: no memory barriers needed - we only care about the boolean itself, not coordinating access to other data

**how it works**
```rs
  // Create flag at download start
  let cancel_flag = Arc::new(AtomicBool::new(false));
  self.cancel_flags.insert(model_id, cancel_flag.clone());

  // Check in download loop (called for each chunk)
  if cancel_flag.load(Ordering::Relaxed) {
      drop(file); // Close file cleanly
      // Cleanup state, return early
      return Ok(());
  }

  // Set flag in cancel_download() command
  flag.store(true, Ordering::Relaxed);
```

**why keep partial files?**
- Users expect cancel ≠ delete progress. partial files enable HTTP range requests for resume. iirc only explicit deletion should removes models

alternatives considered~
- `tokio::sync::CancellationToken` - heavier abstraction, overkill for simple boolean flag
- channel-based signaling - more complex setup, no clear benefit
- aborting the task - leaves files in inconsistent state, prevents clean shutdown

edge cases handled~
- cancel during extraction (tar.gz): stops extraction loop, keeps partial `.tar.gz`
- cancel right before completion: race condition handled gracefully, file either completes or stops
- multiple rapid cancels: Flag already set, no-op on subsequent calls

performance~
- `AtomicBool::load()` is extremely fast
- check happens once per chunk (~every 8KB), not a hot path
- no locks taken in download loop (locks only in setup / teardown)

code quality~
- standard Rust idiom for cooperative cancellation. follows patterns from tokio/async-std ecosystems. easy to reason about and test

qq: any concerns about the approach or edge cases I might have missed?

src-tauri/src/managers/model.rs
```diff
@@ -456,8 +488,36 @@ impl ModelManager {
             .app_handle
             .emit("model-download-progress", &initial_progress);
 
+        // Track last emission time for throttling
+        let mut last_emit = Instant::now();
+        let emit_interval = Duration::from_millis(100);
```
VirenMohindra: **Download Progress Event Throttling**

this fixes a critical performance bug where large model downloads would freeze the UI by emitting hundreds of progress events per second, and i believe this occurred before this PR also, but was surfaced when we added support for resume / cancellation in this PR

the problem~
before throttling, progress events were emitted on every chunk received from the HTTP stream:
- typical chunk size: 8-64 KB
- fast connection (50 MB/s): ~800-6000 chunks / sec
- each event triggers zustand state update + react re-render
- result: UI thread saturated, app freezes during downloads

the solution (100ms Throttling)~
```rs
  let mut last_emit = Instant::now();
  let emit_interval = Duration::from_millis(100);

  while let Some(chunk) = stream.next().await {
      // ... download chunk ...

      let now = Instant::now();
      if now.duration_since(last_emit) >= emit_interval {
          app_handle.emit("model-download-progress", &progress)?;
          last_emit = now;
      }
  }

  // Final 100% event after loop completes
  app_handle.emit("model-download-progress", &final_progress)?;
```

performance impact~
- before: 500-1000 events / sec on fast connections (tested with 2GB model @ 50MB/s)
- after: max 10 events / sec (`100ms` interval)
- result: 100x reduction in event traffic, smooth UI throughout download

why 100ms Specifically?
- `50ms` still occasional UI lag on slower devices (Chromebooks tested)
- `100ms` feels responsive, no perceivable stuttering
- `200ms` visibly choppy progress bar updates

trade-off~
- progress bar updates 10 times per second instead of continuously. In practice, 10 updates / sec is imperceptible to users - looks smooth. The alternative (UI freeze) is **unacceptable**

user-facing impact~
- users can now interact with the app normally during large downloads. previously, the entire app would become unresponsive for 30-60 seconds during a 2GB download. i believe this was reported as a bug by multiple users

alternative considered~
- fe debouncing - tried first, but doesn't prevent `Tauri` event channel saturation. backend throttling is the correct layer to fix this

code quality~
- uses standard Rust Instant for wall-clock time. simple, testable logic. The final progress event ensures UI always shows 100% completion, not stuck at 99.8%

qq - given the dramatic performance improvement, is there any concern about the slightly delayed progress updates?

src/components/settings/general/ModelSettingsCard.tsx
VirenMohindra, line 1: **Contextual Settings Approach**

this component directly addresses your Dec 22 feedback about disabled settings being confusing for Parakeet users

Before (The Problem)~
```tsx
  // Language selector and translation toggle always visible
  <LanguageSelector disabled={!supportsLanguage} />
  <TranslateToEnglish disabled={!supportsTranslation} />
  // + Explanatory text: "This model doesn't support language selection"
```

issues~
- cluttered UI with disabled controls
- users confused why settings exist but don't work
- required reading explanatory text to understand why disabled

after (this implementation)~
```tsx
  // Settings card only shows relevant controls
  {hasSettings ? (
    <>
      {supportsLanguage && <LanguageSelector />}
      {supportsTranslation && <TranslateToEnglish />}
    </>
  ) : (
    <p>This model works automatically...</p>
  )}
```

benefits~
- cleaner UI: only show what's relevant (show / hide vs disable / enable)
- less cognitive load: No need to understand why something is disabled
- better UX pattern: "If you can see it, you can use it"
- extensibility: Easy to add model-specific settings in the future

example UX~
- whisper models: Shows language selector + translation toggle (both work)
- parakeet models: Shows "This model works automatically with English audio" message
- future noonshine: Could show different settings entirely (e.g., CPU vs GPU toggle)

code quality~
- straightforward conditional rendering
- no complex state management, just adapts to `currentModel` prop
- reuses existing `LanguageSelector` / `TranslateToEnglish` components (DRY)
- type-safe via `ModelInfo` interface

Maintainability~
adding a new model type is simple:
1. add fields to `ModelInfo` in Rust (e.g., `supports_gpu_acceleration`)
2. add conditional rendering in `ModelSettingsCard`
3. works automatically for all future models

qq - this follows the "show relevant, hide irrelevant" UX principle rather than "show everything, disable what doesn't work." any concerns about this approach?

scripts/check-translations.cjs
VirenMohindra: benefits~
- caught 8 missing keys during development (manual review missed these)
- prevents runtime errors: missing key = fallback to English (confusing for users)
- scales with codebase: as features grow, manual tracking becomes impossible
- CI Integration: fails builds if translations incomplete (caught before merge)

limitations (by design)~
_validates structural consistency (same keys exist)_
- doesn't validate translation quality (still manual / native speaker review), not really solvable unless we use some open source i18n verification software
- doesn't check for typos or grammatical errors
- doesn't verify interpolation variables match (e.g.,` {{modelName}}`)

VirenMohindra: usage~

# Locally during development
```sh
bun run scripts/check-translations.cjs
```

# CI (added to lint workflow)
```
  - name: Check translations
    run: bun run scripts/check-translations.cjs
```

real-world example (from development)~
❌ missing keys in es/translation.json

code quality~
- pretty small, well-commented
- handles nested objects recursively
- clear error messages with exact key paths
- no external dependencies (pure Node.js)

maintainability~
As new languages are added~
1. add locale file to locales/ directory
2. script automatically detects and validates it
3. CI enforces completeness from day one

alternative considered~
using a translation management platform (`Crowdin`, `Lokalise`). rejected for now - adds complexity and cost for a single maintainer. Can migrate later if team grows

qq - with 36 new keys added in this PR across 12 languages (432 total translations), manual tracking isn't sustainable. is there any concern about automating this validation?

src/stores/modelStore.ts
```diff
@@ -0,0 +1,383 @@
...
+        set((state) => {
+          const next = new Set(state.downloadingModels);
+          next.delete(modelId);
+          return { downloadingModels: next };
+        });
+        return false;
+      }
+    },
+
+    cancelDownload: async (modelId: string) => {
```
VirenMohindra, line 200: **Double-Click Cancel Bug Fix**

this addresses a non-obvious race condition discovered during testing where cancelling downloads would sometimes cause them to auto-resume ("ghost downloads")

the original issue~
1. user clicks "Cancel" on downloading model
2. `cancelDownload()` immediately clears `downloadProgress` Map
3. a few milliseconds later, `loadModels()` is called (triggered by be event)
4. be still shows `is_downloading: true` (hasn't processed cancel yet)
5. `loadModels()` sees backend state and re-adds model to downloadingModels Set
6. UI shows download resuming unexpectedly - "ghost download"

root cause~
- we were checking `downloadProgress.has(modelId)` to determine if something was downloading. But `cancelDownload()` cleared that Map immediately, causing the check to return false before the backend confirmed cancellation

The Fix (This Commit)~
```ts
  // OLD (Commit 92a022b^ - before fix)~
  const isDownloading = downloadProgress.has(model.id); // ❌ Cleared immediately on cancel
```

```ts
  // NEW (Commit 92a022b - after fix)~
  const isDownloading = downloadingModels.has(model.id); // ✅ Stays true until backend confirms
```

why this works~
1. user clicks cancel
2. `cancelDownload()` immediately:
  1. removes from `downloadingModels` Set
  2. removes from `downloadProgress` Map
  3. removes from `downloadStats` Map
3. be call: `commands.cancelDownload(modelId)`
4. `loadModels()` syncs with be
5. intelligent merging (L82-105) prevents re-adding because~
  1. be says not downloading + no progress data = **truly cancelled**
  2. model stays removed from Set

connection to state merging~
- this bug fix and the state merging logic solve different aspects of the same problem: preventing frontend / backend state drift during async operations

testing~
- click cancel -> download stops immediately, doesn't resume
- rapid cancel-download-cancel -> handled gracefully
- cancel near end (95% complete) -> stops without completing
- cancel during network pause -> resumes don't trigger auto-download

user impact~
- before fix: users would click cancel, see download disappear, then mysteriously resume 2-3 seconds later felt broken and unresponsive
- after fix: cancel is immediate and permanent. download only resumes if user explicitly clicks download again

code quality~
**Three-phase cleanup ensures atomic state updates~**

```ts
  // Phase 1: Clear frontend state (immediate UI feedback)
  set((state) => ({ /* remove from all Maps/Sets */ }));

  // Phase 2: Tell backend to cancel
  await commands.cancelDownload(modelId);

  // Phase 3: Sync to ensure consistency
  await get().loadModels(); // Uses intelligent merging
```

this pattern is now consistent across all download state mutations (start, cancel, complete)

qq - the fix was subtle to track down - took multiple iterations to identify the race condition. any concerns about the approach or edge cases I might have missed?

cjpais:
You're awesome thank you for this, I'll have some time in a few days to take a more solid look! Let's get this merged!

cjpais:
Okay @VirenMohindra @virenpepper, I've started to review this and I have some notes

* I think the download/delete right now feels a bit awkward, mainly it causes quite a bit of layout shift. We really need to keep the layout consistent I think. Perhaps it's moved to a slightly different location? Maybe the bottom right of the card? I'm not quite sure, but I think there is some better place.
* I removed the filters for translation/etc for now, it just felt a bit overkill to me. However this just leaves the language selector which also feels a bit awkward visually now. Not sure the best way forward on this one. I was thinking it could potentially be a multiselect instead of a dropdown itself? Basically you may want to filter for a few languages? It could potentially just take the entire row? Not really sure
  * one note on this too, is that possibly it makes sense to have sort? 'default', 'speed', 'accuracy', kind of thing? not sure... I think we need to have a good default ordering too. right now it feels a bit random. I also am wondering if there should be a search for the model itself? maybe that makes sense instead of the multi language select and leaving that as a dropdown?
* I saw a bunch of <button> with hard coded values.. we really should try to use the <Button> UI component everywhere. Generally we need to have the UI have all of its own components and use those. We don't need to refactor all of this now, but this is something that will become very important for maintainability and helping others contribute most easily. 
* the cancel button for downloading I dont think reflects the state properly. if you hit it, the bottom still has downloading, but its just stuck. and then you resume the download and the bottom doesnt resume immediately either. that needs testing and fixing.
* one last note is that for the models, im trying to figure out if there is a better way to distinguish between models which are downloaded and can be downloaded. it feels a bit weird to me these interactions are mixed right now, but we can continue to iterate on this a bit

VirenMohindra:
thank you for the comprehensive feedback! going line by line. opted to divide each work item by commit

> * I think the download/delete right now feels a bit awkward, mainly it causes quite a bit of layout shift. We really need to keep the layout consistent I think. Perhaps it's moved to a slightly different location? Maybe the bottom right of the card? I'm not quite sure, but I think there is some better place.

agree. bottom works. adjusted the styles a bit and fixed floating accuracy / speed bars when card expands during download (9024719). some slight layout shifting will occur but this IMO is a much better experience

> * I removed the filters for translation/etc for now, it just felt a bit overkill to me. However this just leaves the language selector which also feels a bit awkward visually now. Not sure the best way forward on this one. I was thinking it could potentially be a multiselect instead of a dropdown itself? Basically you may want to filter for a few languages? It could potentially just take the entire row? Not really sure

im also unsure. does top right not work? maybe we can iterate more once this PR lands

>   * one note on this too, is that possibly it makes sense to have sort? 'default', 'speed', 'accuracy', kind of thing? not sure... I think we need to have a good default ordering too. right now it feels a bit random. I also am wondering if there should be a search for the model itself? maybe that makes sense instead of the multi language select and leaving that as a dropdown?

possible to address these enhancements in a follow up PR? 

> * I saw a bunch of `<button>`  with hard coded values.. we really should try to use the `<Button />  UI component everywhere. Generally we need to have the UI have all of its own components and use those. We don't need to refactor all of this now, but this is something that will become very important for maintainability and helping others contribute most easily.

nah you're right. refactored and added  `primary-soft` and `danger-ghost` variants (5429f0d)

> * the cancel button for downloading I dont think reflects the state properly. if you hit it, the bottom still has downloading, but its just stuck. and then you resume the download and the bottom doesnt resume immediately either. that needs testing and fixing.

ah good catch. cancel button now properly clears the bottom progress indicator (960d387) also fixed a race condition where deleting a model would interrupt another model's extraction (807309e)

> * one last note is that for the models, im trying to figure out if there is a better way to distinguish between models which are downloaded and can be downloaded. it feels a bit weird to me these interactions are mixed right now, but we can continue to iterate on this a bit

split models into "your models" and "available to download" sections for clearer distinction (3a73103)

_also fixed all missing translation keys that were failing ci (including the new ar locale)_

let me know if you want any tweaks!

edwche10:
Hi VirenMohindra 

My name is Edward, and I’m a product designer. I’d like to contribute to this PR and help move it forward.

If you’re open to it, I’m happy to work with you to refine and finish the design. I’ve already discussed a few possible changes with CJ.

I’ve attached an image showing a design direction I explored Figma. Not everything needs to be implemented, but it should give a sense of what’s possible.

If you prefer, we can also continue the discussion on Discord. I’d be glad to collaborate there and help finalize the design together.

Handy

cjpais:
Just adding some context from email

Edward
> UI feedback:
>
> Model management should live in one place in Settings.
> There is no need to repeat it in a dropdown (1).
> Filters and search are not needed (2).
> There are only seven models.
> Scanning is faster than filtering.
> Language selection feels unnecessary (3).
> In most cases, the AI model already handles language automatically. (When I talk in a different language, it understands the language automatically.)
> Please correct me if this assumption is wrong.
> The delete icon does not feel aligned with the rest of the UI on this page (4).
> Reusing the existing delete action would keep the interface cleaner.
> I added a visual indicator so the user can see which model is active (5).
> It is also important to make it clear how the user sets the active model.
> For example, by clicking on the model card itself.

CJ
> 1: Right now I want to keep both, but management of the dropdown to just be model selection. This may change in the future, but trying to make most of the code stay the same for now.
> 2: More models will be added soon and this UI is in service of making that slightly more accessible in the future.
> 3: I get a TON of messages about language support, and this was a critical thing to add so people are able to filter their language. Auto does not work well for non-western languages. Also more models in the future will make this messier.
> 4: Big yes, this is a big issue and your cleanup is good. Curious if there are very clear destructive states
> 5: Also good to have, the dot in the bottom where it was before shows a few types of state as well (active, loading, unloaded) so we may need to carry this forward in some way.
>
> I think another thing there is not clear distinction between downloaded models and models you can download. Not sure what that looks like.

edwche10:
Hi people, I add here a design with two options for the screen selection. Let me know what you think.

https://youtu.be/OFnDkcqEzi8

cjpais:
I just want to say a huge shout out for all this work! Thank you for bearing with me and getting this merged, but I think it's good to go. Super excited to get this in. I think this improves a bunch of UI and leaves a lot more room for future improvements as well. Thank you!

PR #520, opened 2026-01-04
feat: Add custom macOS Fn key listener to enable its use as a shortcut.

its-bash33r-here:
I wanted to use the Fn key for voice transcription as WhisperFlow does, but it wasn't available in the shortcut picker. Regular shortcut APIs can't detect the Fn key because it's a hardware modifier, so I implemented a native macOS solution using CGEventTap that also blocks the emoji picker from appearing.

### DIFF
--- a/bun.lock
+++ b/bun.lock
@@ -1,5 +1,6 @@
 {
   "lockfileVersion": 1,
+  "configVersion": 0,
   "workspaces": {
     "": {
       "name": "handy-app",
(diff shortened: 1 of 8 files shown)


cjpais:
Hey thanks for this. I'm also rewriting the keyboard backend to handle this in a more generalized way. I'm going to close this in favor of my PR as it's going to be a fairly extensive rewrite aiming to fix many outstanding issues

its-bash33r-here:
yeah no problem @cjpais

PR #543, opened 2026-01-06
fix: bottom overlay positioned partially off-screen

antons:
## Before Submitting This PR

## Human Written Description

This commit reverts a change from 0068b6b which caused overlay to be shown partially off-screen when enabling System Settings > Desktop & Dock > Automatically hide and show the Dock.

The commit that introduced this issue added a comment “don't subtract the overlay height it puts it too far up”. I’m not a fan of overlapping the Dock, but this is subjective. I can amend the PR to keep the overlay position as is when the Dock is visible.

## Before with the Dock hidden
before-crop

## After with the Dock hidden

after-crop

## Before
dock-before-crop

## After

dock-after-crop

## Wispr Flow

flow-crop

### DIFF
--- a/src-tauri/Cargo.lock
+++ b/src-tauri/Cargo.lock
@@ -6581,7 +6581,7 @@ dependencies = [
 [[package]]
 name = "tauri-runtime"
 version = "2.9.1"
-source = "git+https://github.com/antons/tauri.git?branch=fix%2Foverlay-bottom-position#af6f5a91105f696a1b76a5db6ba936a30bbf80c6"
+source = "git+https://github.com/cjpais/tauri.git?branch=handy-2.9.1#8ba98c7b543ec155961e1840641b1f28c548b8dd"
 dependencies = [
  "cookie",
  "dpi",
@@ -6605,7 +6605,7 @@ dependencies = [
 [[package]]
 name = "tauri-runtime-wry"
 version = "2.9.1"
-source = "git+https://github.com/antons/tauri.git?branch=fix%2Foverlay-bottom-position#af6f5a91105f696a1b76a5db6ba936a30bbf80c6"
+source = "git+https://github.com/cjpais/tauri.git?branch=handy-2.9.1#8ba98c7b543ec155961e1840641b1f28c548b8dd"
 dependencies = [
  "gtk",
  "http",
@@ -6659,7 +6659,7 @@ dependencies = [
 [[package]]
 name = "tauri-utils"
 version = "2.8.0"
-source = "git+https://github.com/antons/tauri.git?branch=fix%2Foverlay-bottom-position#af6f5a91105f696a1b76a5db6ba936a30bbf80c6"
+source = "git+https://github.com/cjpais/tauri.git?branch=handy-2.9.1#8ba98c7b543ec155961e1840641b1f28c548b8dd"
 dependencies = [
  "anyhow",
  "brotli",
--- a/src-tauri/Cargo.toml
+++ b/src-tauri/Cargo.toml
@@ -102,9 +102,9 @@ gtk-layer-shell = { version = "0.8", features = ["v0_6"] }
 gtk = "0.18"
 
 [patch.crates-io]
-tauri-runtime = { git = "https://github.com/antons/tauri.git", branch = "fix/overlay-bottom-position" }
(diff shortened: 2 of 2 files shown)


VirenMohindra (approved):
nice work!

src-tauri/src/overlay.rs
```diff
@@ -121,8 +121,7 @@ fn calculate_overlay_position(app_handle: &AppHandle) -> Option<(f64, f64)> {
         let y = match settings.overlay_position {
             OverlayPosition::Top => work_area_y + OVERLAY_TOP_OFFSET,
             OverlayPosition::Bottom | OverlayPosition::None => {
-                // don't subtract the overlay height it puts it too far up
-                work_area_y + work_area_height - OVERLAY_BOTTOM_OFFSET
+                work_area_y + work_area_height - OVERLAY_HEIGHT - OVERLAY_BOTTOM_OFFSET
```
VirenMohindra, line 171: how does this look on windows / linux?

antons: Thank you for taking a look. Unfortunately, I don’t have access to either.

cjpais:
I can ping other folks to test but we might be able to just feature flag this off for MacOS or something

cjpais:
@antons I'm curious if we can do any kind of dock detection logic or even get system information from the dock height because I suspect that this doesn't work properly across all systems and screens. Mainly because you can change the dock height itself, and this is just a constant currently. At least in my case, I have the dock automatically hiding and when it's showing again I still have the overlap with the overlay being on the dock as well, which is not ideal.

Screenshot 2026-01-10 at 11 04 56 AM

Possibly this is something we could do via `objc2` if not exposed nicely through Tauri?

borzov:
Nice fix! This resolves a real issue with auto-hiding Dock — the overlay now stays fully on-screen without clipping. Personally prefer subtracting overlay height from position calculation, aligns with HIG. 👍

antons:
@cjpais I haven’t looked into the code on the Dock height, but tried resizing the always-visible Dock just now and the overlay seems to correctly position itself outside the Dock no matter what size it is. 

For auto-hiding Dock, animating the overlay to get out of the way in case the Dock appears is probably possible. I see this as low priority, because I almost never switch apps while actively recording.

cjpais:
@antons oh interesting I will have to try this PR again, it did not work nicely for me when I tried on my machine

cjpais:
im going to review this again in the next few days probably, also curious if [tauri_plugin_positioner](https://docs.rs/tauri-plugin-positioner/latest/tauri_plugin_positioner/index.html) may help

cjpais:
@borzov @VirenMohindra @antons for whatever reason on my machine the behavior seems correct without the fix, and with the fix makes it place more awkward. Without the fix the behavior is not like this for you? This is how it is for me without the fix

Screenshot 2026-02-06 at 4 53 09 PM

antons:
@cjpais I’ve been experiencing this bug, as shown on my screenshots, consistently on two different Macs with macOS 26.3 betas and RC. Maybe on macOS 26.2 before that, but I’m not sure. Both Macs (mini and Studio) connected to a single external screen. Seems like there is a difference in either software or hardware configuration that causes this.

cjpais:
Interesting thanks for sharing more details

antons:
I have also now tried showing the menu bar (normally I have it automatically hidden), and changing screen scaling to match native pixel size, 2x retina, and then smaller scaled, and couldn’t get the overlay to render in a different way.

antons:
@cjpais After more testing and another round of investigation, I can reliably reproduce this bug when the menu bar is set to be automatically hidden in System Settings. 

A full fix needs both:
1. this PR
2. the fix for a Tauri bug that already has an open PR at tauri-apps/tauri#14655.
  
I pushed a temporary Tauri crate override for you to test. 

That Tauri PR has been open for a while with no recent activity. I’m new to Tauri, and I’m not sure what is the best way to proceed. Handy’s own fork of Tauri?

cjpais:
@antons thanks for this, it works great on my system as well now. Let me know if with my fork everything works as expected for you. I'll merge it, I also kicked off a test build for this PR just to make sure CI and everything works fine

I posted a comment there just showing support for that PR, hopefully it will be merged

antons:
@cjpais Tested. Looks good to me!

cjpais:
Awesome thank you! Merged

PR #545, opened 2026-01-07
Add permission step to onboarding

anchpop:
This PR adds an onboarding step on MacOS that explicitly asks the user to grant permissions. Once permissions are granted, it automatically moves on to the next screen. It also fixes a bug where the app would crash if it did not have the necessary permissions, by removing `input::EnigoState::new().expect("...");`

https://github.com/user-attachments/assets/6dcb0448-32d9-4674-b487-4ee9d1758802

The most important architectural change is that `input::EnigoState::new()` is longer be called from `initialize_core_logic`, because this triggers a permission prompt, and I wanted the permission prompt not to appear until after the user clicked a "grant permission" button. (I didn't put too much thought into phrasing – "open settings" might be better, but a permission prompt does appear first.)

See discussion in #544

### DIFF
--- a/src/components/onboarding/AccessibilityOnboarding.tsx
+++ b/src/components/onboarding/AccessibilityOnboarding.tsx
@@ -40,16 +40,13 @@ const AccessibilityOnboarding: React.FC<AccessibilityOnboardingProps> = ({
     microphone: "checking",
   });
   const pollingRef = useRef<ReturnType<typeof setInterval> | null>(null);
+  const timeoutRef = useRef<ReturnType<typeof setTimeout> | null>(null);
+  const errorCountRef = useRef<number>(0);
+  const MAX_POLLING_ERRORS = 3;
 
   const allGranted =
     permissions.accessibility === "granted" &&
     permissions.microphone === "granted";
-
-  const anyNeeded =
-    permissions.accessibility === "needed" ||
-    permissions.microphone === "needed" ||
-    permissions.accessibility === "waiting" ||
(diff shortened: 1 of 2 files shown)


anchpop:
My bad, I didn't realize I had pushed the wrong version of my branch. Updated it now.

anchpop:
Looked over the code, I think it's ready for review now.

I experienced a bug during testing where the app got stuck in the "loading" state according to the model selector, and transcriptions didn't work. It happened on the second open (e.g. open → onboard → close → open → stuck in "loading"). Unfortunately I stopped being able to reproduce it, and I'm not sure if it's related to my changes or a coincidence.

cjpais:
Okay, thank you for this. I'll take a look soon, but give me maybe a week or two. I'm quite busy right now, but I do wanna bring this in for sure.

anchpop:
Oh yeah, probably a good call on the translations haha. I forgot about that. Does the code need to be updated to use them?

cjpais:
Should help to fix #564

anchpop:
@cjpais I can confirm that this PR fixed #564 on my machine (when I tested it)! The crash was caused by this line:

```
fn initialize_core_logic(app_handle: &AppHandle) {
    let enigo_state = input::EnigoState::new().expect("Failed to initialize input state (Enigo)"); // ← crash!
    // [...]
}
```

cjpais:
Thanks yea, I was unable to replicate the crash on my machine even toggling the permission off

PR #550, opened 2026-01-07
feat: add live transcription feature

DevDeepakBhattarai:
## Before Submitting This PR

## Human Written Description

I had an idea  / feature that would really be useful for this app. It is live typing as I speak rather than waiting for the entire speech. Why this matters, is to help user keep track of what has been typed till now and help people from getting lost on what they are saying. Happens to me a lot so I thought other would love it as well.

## Testing
I testing the application by just running it, and it successfully compiled without any errors and this does not touch existing logic, just add a live transcription feature. 

image

### DIFF
--- a/src-tauri/src/input.rs
+++ b/src-tauri/src/input.rs
@@ -122,13 +122,3 @@ pub fn paste_text_direct(enigo: &mut Enigo, text: &str) -> Result<(), String> {
     Ok(())
 }
 
-/// Sends backspace key presses to delete characters.
-/// Used for live transcription to correct previous output.
-pub fn send_backspace(enigo: &mut Enigo, count: usize) -> Result<(), String> {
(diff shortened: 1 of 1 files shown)


cjpais:
I appreciate this PR. I would generally much prefer if we could have had this feature to `transcribe-rs` so we can more broadly support the community. Then use it appropriately in Handy. Also for beta features they at minimum should be in advanced not general. Please let me know if this was written with assistance of AI.

DevDeepakBhattarai:
Thanks for the feedback @cjpais. I can may be try to implement this in `transcribe-rs` as whisper already supports live transcript. It would be a really great feature. But I am a bit new to this rust eco-system, I would really love if you could guide me. And as per this PR, it is written with the assistance of AI.

PR #551, opened 2026-01-07
feat(appearance): add customizable accent themes and overlay styles

brieyla1:
## Before Submitting This PR

  **Please confirm you have done the following:**

  - [x] I have searched [existing issues](https://github.com/cjpais/Handy/issues) and [pull requests](https://github.com/cjpais/Handy/pulls) (including closed ones) to ensure this isn't a duplicate
  - [x] I have read [CONTRIBUTING.md](../CONTRIBUTING.md)

  ## Description

  I found that apps like SuperWhisper and WhisperFlow have really polished, good-looking UIs. Handy is not only fast but also fully local, so I wanted to combine the best of both worlds—bringing a more visually appealing and customizable UI to both the settings page and the recording overlay.

  **Changes:**

  - add new Appearance settings tab in sidebar with accent color and overlay customization
  - add 6 accent color themes (Pink, Blue, Green, Purple, Orange, Teal) applied across app and overlay
  - add 3 overlay style themes (Pill, Minimal, Glassmorphism) with live preview
  - add audio bar customization: count (3-15), size (2-12px), color (accent or preset colors), centered mode
  - add icon visibility toggle for overlay themes
  - create shared overlay components (`AudioBars`, `OverlayDisplay`) used by both preview and actual overlay
  - add real-time event updates so overlay reflects changes immediately without restart
  - add Tauri commands for all new appearance settings with persistence
  - add translations for new appearance settings

  ## Related Issues/Discussions

  Related to UI/UX improvements and customization requests.

  ## Testing

  - [x] verified accent theme switching works across all 6 themes
  - [x] verified overlay themes render correctly (Pill, Minimal, Glassmorphism)
  - [x] verified bar customization updates in real-time (count, size, color, centered)
  - [x] verified icon toggle works on all overlay themes
  - [x] verified preview matches actual recording overlay
  - [x] verified settings persist across app restarts
  - [x] verified lint passes
  - [x] verified cargo check passes

  ## Screenshots/Videos (if applicable)

CleanShot 2026-01-08 at 03 58 05@2x

PR #552, opened 2026-01-08
visual tweak: improve recording visualization with symmetric bars

BerthalonLucas:
- Add symmetric mirror effect for Apple-like audio visualization
- Adjust FFT frequency range (150-2500Hz) to better capture voice
- Fix intermittent GPU compositing issues with transparent backgrounds
- Fix unused import and mut warnings

  - [x] I have searched existing issues and pull requests to ensure this isn't a duplicate

  ## Description

  I thought the recording overlay could look nicer, kind of like Siri or other Apple-style
  visualizations - cleaner and more polished. So I started experimenting with different
  configurations to find what looked best.

  I noticed that only the first few bars on the left were moving when I spoke, which felt
  off. I dug into the code to understand how it worked and discovered the FFT frequency
  distribution. That's when I got the idea to make the bars emanate from the center
  outward - it just looks way better and more natural.

  I also ran into a weird bug where sometimes a visible rectangle appeared around the bars
  (GPU compositing issue), so I fixed that too but it seems to be still here on screenshot but not in prod.

sorry if things like coms, commit etc... seems ai-made, i'm french and not so good in English so i used ai to rewite my thoughts for it to be more clear and understandable. also this is my first PR so please go easy on me.

  ## Testing

  - Tested on Ubuntu 24.04 (X11)
  - Verified symmetric bar animation responds correctly to voice
  - Confirmed GPU compositing fix resolves intermittent opacity rectangle issue

Screenshot from 2026-01-08 00-59-26
Screenshot from 2026-01-08 00-59-18
Screenshot from 2026-01-08 00-59-09

### DIFF
--- a/src-tauri/src/lib.rs
+++ b/src-tauri/src/lib.rs
@@ -320,7 +320,7 @@ pub fn run() {
         )
         .expect("Failed to export typescript bindings");
 
-    let builder = tauri::Builder::default().plugin(
+    let mut builder = tauri::Builder::default().plugin(
         LogBuilder::new()
             .level(log::LevelFilter::Trace) // Set to most verbose level globally
             .max_file_size(500_000)
--- a/src-tauri/src/shortcut.rs
+++ b/src-tauri/src/shortcut.rs
@@ -11,7 +11,7 @@ use crate::managers::audio::AudioRecordingManager;
 use crate::settings::ShortcutBinding;
 use crate::settings::{
     self, get_settings, ClipboardHandling, LLMPrompt, OverlayPosition, PasteMethod, SoundTheme,
-    APPLE_INTELLIGENCE_PROVIDER_ID,
+    APPLE_INTELLIGENCE_DEFAULT_MODEL_ID, APPLE_INTELLIGENCE_PROVIDER_ID,
 };
 use crate::tray;
 use crate::ManagedToggleState;


src/overlay/RecordingOverlay.tsx
```diff
@@ -80,9 +84,8 @@ const RecordingOverlay: React.FC = () => {
                 key={i}
                 className="bar"
                 style={{
-                  height: `${Math.min(20, 4 + Math.pow(v, 0.7) * 16)}px`, // Cap at 20px max height
-                  transition: "height 60ms ease-out, opacity 120ms ease-out",
-                  opacity: Math.max(0.2, v * 1.7), // Minimum opacity for visibility
+                  height: `${Math.min(50, 4 + Math.pow(v, 0.7) * 32)}px`,
+                  opacity: Math.max(0.3, Math.min(1, v * 1.5)),
```
jacksongoode: Why did you remove the transitions? This otherwise causes some flickering.

brieyla1:
this is already part of #551 
Maybe we could join effort ?

src/overlay/RecordingOverlay.tsx
BerthalonLucas: you are right, i forgot to put it back in the CSS because keeping it in the TSX made some ghosting artefacts on top of the bars and moving it to CSS fixed the issue.

BerthalonLucas:
Yeah I saw yours, with the symmetric bars and all, it’s pretty close to what I did too.
I didn’t wanted to duplicate or anything, I just wanted to try to contribute something small.

I also saw [cjpais](https://github.com/cjpais) feedback about not changing the global design/colors,
so I thought maybe a smaller tweak like this had more chances to fit in?
But maybe I’m wrong 😅

If it’s easier to just close mine and keep yours, no problem at all! Let me know.

cjpais:
This smaller tweak is much more likely to make it in. Smaller scoped PRs are almost always easier to review and merge

cjpais:
Okay, so I'd like to pull this in, but I do have some issues on my machine generally. 

The overlay for me has actually been fairly consistently broken where the height of the bars doesn't show up properly. I'm just curious if that's also happening on your system. I really haven't had a chance to address this and I don't know if it's just my system also pinging @jacksongoode if you get the bar height displaying properly on your system

BerthalonLucas:
Yes, I'm experiencing similar issues on Linux (Ubuntu 24.04 + NVIDIA GPU). The transparency and bar rendering is very inconsistent:

I'm going to test on Windows (dual-boot) to see if all three platforms are affected.

BerthalonLucas:
I just built and tested it on Windows and everything is working correctly. The overall overlay transparency is functioning properly and the bar rendering works as well.

jacksongoode:
Hmm, I just tested and it looks alright, but I would be a little weary of making 150Hz the lowend? I just worry with most microphones you get quite noisy in the lowend so the original range seemed a little more reasonable (plus most computer mics are tuned for that range as well).

cjpais:
Yeah, I'm closing this for now.

Issue #522, opened 2026-01-04
[BUG] wtype failures seem silently ignored


dandv:
## Bug Description

On Wayland, I had installed wtype (`wtype/noble,now 0.4-3 amd64 [installed]`), but the "Direct" Paste Method didn't paste/type anything. The logs show that Handy is aware of the wtype failure, but this isn't indicated in the UI. I only happened to look at the CLI logs because I'm using the AppImage, but users who launch without a CLI may never see these logs.

Does `wtype` require some post-install configuration? I couldn't find anything more [in the README](https://github.com/cjpais/Handy?tab=readme-ov-file#linux-notes) than `sudo apt install wtype`. Some notes on when to use dotool vs. wtype would help as well.

Logs:
```
[2026-01-04][11:17:24][handy_app_lib::managers::audio][INFO] Microphone stream initialized in 19.033303ms
[2026-01-04][11:17:24][handy_app_lib::audio_toolkit::audio::recorder][INFO] Using device: Ok("default")
Sample rate: 16000
Channels: 1
Format: F32
[2026-01-04][11:17:27][handy_app_lib::managers::transcription][INFO] Transcription completed in 237ms
[2026-01-04][11:17:27][handy_app_lib::managers::transcription][INFO] Transcription result: Just testing
[2026-01-04][11:17:27][handy_app_lib::clipboard][INFO] Using paste method: CtrlV
[2026-01-04][11:17:27][handy_app_lib::clipboard][INFO] Using wtype for key combo
[2026-01-04][11:17:27][handy_app_lib::actions][ERROR] Failed to paste transcription: wtype failed: Compositor does not support the virtual keyboard protocol
```

## System Information

**App Version:**

0.6.10

**Operating System:**

Ubuntu 24.04

**CPU:**

12th Gen Intel® Core™ i7-12700H

**GPU:**

Graphics Processor 1: NVIDIA GeForce RTX 3050 Ti Laptop GPU
Graphics Processor 2: Intel® Iris® Xe Graphics

cjpais:
It seems like wtype just does not work on your system? We will definitely be adding better handling to fallback in the case of incompatibility. But for your case can you just work around this by uninstalling wtype??

'Compositor does not support the virtual keyboard protocol' - this is a compatibility issue between wtype and your compositor which is not particularly an "issue" with Handy imo

dandv:
[...]

PR #557, opened 2026-01-09
feat(linux): add ydotool support for text input

manimohans:

## Human Written Description

  I'm running Handy on Fedora 43 with GNOME/Wayland. The existing tools (wtype, dotool) either aren't packaged for Fedora or don't work with GNOME's compositor. ydotool is available in Fedora repos and uses uinput, which bypasses compositor limitations. This gives Fedora/GNOME users another option for text input.

## Related Issues/Discussions

https://github.com/cjpais/Handy/issues/522

## Testing

Tested on Fedora 43 (Wayland/GNOME):
  - ydotool detection works (logs show "Using ydotool for direct text input")
  - Requires ydotoold daemon running and user in input group
  - Note: Key combo simulation may have issues on some GNOME setups due
    to keycode mapping, but direct text typing works

### DIFF
--- a/src-tauri/src/clipboard.rs
+++ b/src-tauri/src/clipboard.rs
@@ -59,7 +59,7 @@ fn paste_via_clipboard(
 #[cfg(target_os = "linux")]
 fn try_send_key_combo_linux(paste_method: &PasteMethod) -> Result<bool, String> {
     if is_wayland() {
-        // Wayland: prefer wtype, then dotool
+        // Wayland: prefer wtype, then dotool, then ydotool
         if is_wtype_available() {
             info!("Using wtype for key combo");
             send_key_combo_via_wtype(paste_method)?;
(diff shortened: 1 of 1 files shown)


cjpais:
I can't test this since I'm on MacOS, but I'm assuming you've tested and we can just merge this.

jacobwhall:
Just want to express my thanks here, Handy now works for me using my Fedora Silverblue-based (Bluebuild) OS. I had to uninstall wtype so that Handy would only try using ydotool. This is the custom shortcut I created in GNOME Settings:

image

jacobwhall:
To add to the above, here is how I got this working in Fedora:

```
mkdir -p ~/.config/systemd/user
```

Save the following to `~/.config/systemd/user/ydotoold.service`:

```
[Unit]
Description=ydotool daemon

[Service]
ExecStart=/usr/bin/ydotoold
Restart=on-failure

[Install]
WantedBy=default.target
```

Then enable the service:
```
systemctl --user daemon-reload
systemctl --user enable ydotoold.service
systemctl --user start ydotoold.service
```

asharpaev:
Is there a way to add support for cyrillic symbols for ydotool?

cjpais:
thats an upstream issue not a handy issue

jesanabriah:
> Is there a way to add support for cyrillic symbols for ydotool?

I had this same problem with these characters in Handy. Especially Spanish accents and other similar types of characters. The solution was to use "Clipboard (CTRL+Shift+V)" as the paste method and "Copy to clipboard" for clipboard handling.
It doesn't fix the problem with ydotool — it simply no longer uses it. For almost everything I need, it has the same effect. With ydotool it types directly, but since it doesn't have compatibility with certain symbols, everything ends up looking very bad. However, with this configuration it pastes directly from the clipboard by executing CTRL+Shift+V.
As long as the context you are in is compatible with the clipboard using this shortcut, and as long as you don't mind using the clipboard, then the symbol problem will be solved. This is a temporary workaround until there is a better option for Ubuntu 26.04 with Wayland, for example.

PR #559, opened 2026-01-09
Add "Follow OS Input Language" option

vladstudio:
Adds a new language selection option that automatically detects and uses the current OS keyboard input source for transcription. When the user switches keyboard layouts (e.g., from US to Russian), the app detects this and sets the corresponding language for Whisper.

- New `input_source.rs` module with cross-platform keyboard layout detection (macOS via `defaults`, Windows via PowerShell, Linux via `setxkbmap`)
- Maps ~30 common keyboard layouts to ISO 639-1 codes for each platform
- UI shows resolved language (e.g., "Follow OS (Russian)") and polls every 5s
- Falls back to auto-detect if keyboard layout isn't mapped


vladstudio:
https://github.com/user-attachments/assets/b3221589-7968-49fe-a96a-21c8279e5dba

cjpais:
wtf im so sorry I didn't look at this earlier, this is great!!

would you mind rebasing?!

vladstudio:
I tried!

a-luchaninov:
gosh, I neeed this feature!!!

cjpais:
I know me too! I have some hesitations pulling in mainly due to the extra thread, but I will do another review and pull in. It would be massively helpful. I am getting close to finished with transcribe.cpp so features will start making their way in as I have more time for careful review

koloved:
I was thinking that it would be cool if language switching during recording could be done with a single button instead of a keyboard shortcut. Because if you're typing a lot of text and constantly changing the language, it will be very difficult to keep changing it with the key combination. I understand that this is a system setting, but it's worth thinking about making language switching directly in the overlay with a hotkey. (alt?)

Press the voice recording hotkeys. The program checks which language is set to start the recording; an Overlay is shown along with the current language. Have the ability to press a hotkey, as I already wrote above, and have the overlay immediately display the language change.

image

a-luchaninov:
IMHO: This is extension of original feature - can be done in a separate update. First version should be simple if we want it to be delivered at all.

cjpais:
The extension wont be likely to make it into the app soon. There are enough complications around adding the feature that I'm not willing to accept.

The original feature will be merged eventually, but it needs more review passes still.

PR #570, opened 2026-01-12
feat: add shortcut for retyping last transcription

johnmalek312:
create a shortcut for retyping last transcription. sometimes i forget to click on the textbox before starting transcription.

Shortcut defaults to none, must be configured to use.

### DIFF
--- a/src-tauri/src/actions.rs
+++ b/src-tauri/src/actions.rs
@@ -436,6 +436,56 @@ impl ShortcutAction for CancelAction {
     }
 }
 
+struct RetypeLastAction;
+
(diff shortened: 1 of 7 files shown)


cjpais:
New bindings are not going to be added right now. Please **collect community support first** for any new bindings and fill out the PR form properly.

johnmalek312:
Does it make sense to have a bunch of shortcuts that are disabled by default and any user can enable them?

cjpais:
Please use discussions and read the existing threads. Community support is vital. You're welcome to collect any kind of feedback from the community there. I am overwhelmed with issues and this is very low priority for me without any major community support.

PR #573, opened 2026-01-12
feat(macos): add Fn key support for transcribe shortcut

shahshrey:
## Before Submitting This PR

## Human Written Description

I've been using Handy daily and noticed that pressing a key combination like option+Space every time is inconvenient. I am switching from Whisperflow. Um and I've always been used to use the function key to transcribe and a lot of other tools also use the same pattern. And just having a single key is a lot more useful than having to press a combination. So that's why I think this feels much more convenient

## Summary of Changes

- Adds support for using the macOS **Fn key** as the transcribe shortcut
- Implements a dedicated Fn key listener using `rdev` that runs on a separate thread
- Works with both push-to-talk and toggle recording modes
- Only available on macOS (requires accessibility permissions)

### Technical Details

- **Backend**: New `init_fn_key_listener` and `is_fn_key_listener_available` Tauri commands
- **Backend**: Fn key listener thread using `rdev` to capture Function key events
- **Frontend**: Integration to detect Fn key presses when editing the transcribe shortcut
- Proper suspend/resume handling for the Fn key binding

## Testing

- [x] Set transcribe shortcut to "Fn" on macOS
- [x] Verify Fn key triggers recording in push-to-talk mode
- [x] Verify Fn key toggles recording in toggle mode
- [x] Verify other shortcuts continue to work normally
- [x] Verify feature is disabled on non-macOS platforms

[](https://github.com/user-attachments/assets/36f21b82-92c8-4163-897b-c6814e8b5471)

## AI Assistance

**If AI was used:**

- Tools used: Cursor - Helped with some implementation and testing

### DIFF
--- a/src-tauri/src/lib.rs
+++ b/src-tauri/src/lib.rs
@@ -260,6 +260,8 @@ pub fn run() {
         shortcut::update_custom_words,
         shortcut::suspend_binding,
         shortcut::resume_binding,
+        shortcut::init_fn_key_listener,
+        shortcut::is_fn_key_listener_available,
         shortcut::change_mute_while_recording_setting,
         shortcut::change_append_trailing_space_setting,
         shortcut::change_app_language_setting,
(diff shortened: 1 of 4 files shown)


cjpais:
Closing this because I'm in the process of rewriting the keyboard backend entirely and it will include this and other features in a more unified manner. Also fix other outstanding bugs

PR #617, opened 2026-01-18
Fix audio playback in History tab on Linux

GabeDuarteM:
## Before Submitting This PR

Btw, just wanted to mention that the link for the contributing document above is broken on the template

## Human Written Description

While I was working on the flatpak pr, I noticed that the play/seek functionality on the history tab wasn't working on my machine, after investigating, I saw that it wasn't a flatpak issue, but a linux issue (or at the very least an arch-linux issue). This should fix it, and I also added some missing portuguese translations while I was at it :)

Had to add the fs:scope so we can read from the recordings folder, and also had to switch away from `convertFileSrc`, since that was giving a `MEDIA_ERR_SRC_NOT_SUPPORTED` error when used, and duplicating the encoding of the url (e.g. `/` woudl became `%2F`)

## Related Issues/Discussions

Couldn't find an issue, I just fixed it directly, but I'm happy to also file an issue/discussion if needed

## Testing

Tested the native dev build + the flatpak branch, both worked correctly. Couldn't get the appimage build due to the build errors on my machine (same thing that happened on the flatpak branch, which also happens to me on the main branch, not sure why tbh)

Before:

https://github.com/user-attachments/assets/a206a24f-d760-4d2d-819b-88fee4bd8e82

After:

https://github.com/user-attachments/assets/6568369f-35b1-4465-9efd-7af1f3af66ec

## AI Assistance

**If AI was used:**

- Tools used: OpenCode (opus 4.5)
- How extensively: mostly for debugging, the fix was quite small

### DIFF
--- a/src/components/settings/history/HistorySettings.tsx
+++ b/src/components/settings/history/HistorySettings.tsx
@@ -3,10 +3,14 @@ import { useTranslation } from "react-i18next";
 import { AudioPlayer } from "../../ui/AudioPlayer";
 import { Button } from "../../ui/Button";
 import { Copy, Star, Check, Trash2, FolderOpen } from "lucide-react";
+import { convertFileSrc } from "@tauri-apps/api/core";
 import { listen } from "@tauri-apps/api/event";
+import { platform } from "@tauri-apps/plugin-os";
 import { readFile } from "@tauri-apps/plugin-fs";
 import { commands, type HistoryEntry } from "@/bindings";
 import { formatDateTime } from "@/utils/dateFormat";
+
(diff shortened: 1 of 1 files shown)


cjpais:
thank you for looking into this, but for what it's worth this breaks playback on macOS which we cannot do

maybe detect os and only for linux use this new codepath in the frontend?

GabeDuarteM:
Sure, done on the last commit!
Also revoked the blob on component unmount

cjpais:
Thank you, works on my Mac now

PR #622, opened 2026-01-19
feat(models): add opt-in custom Whisper model discovery

CJHwong:
## Before Submitting This PR

## Human Written Description

I work with mixed-language speech (Chinese and English) and found that the built-in Whisper models struggle with this use case.

The problem was that using a custom model required either modifying the source code or a workaround of renaming the file to match a built-in model name. This PR adds the ability to auto-discover custom Whisper GGML models from the models directory.

To address the concern about users expecting support when custom models don't work, the feature is opt-in only: it lives behind a toggle in Debug settings (off by default).

## Related Issues/Discussions

Fixes #
Discussion: https://github.com/cjpais/Handy/discussions/609

## Community Feedback

https://github.com/cjpais/Handy/discussions/609#discussioncomment-15531080

## Testing

Tested with:
- `MediaTek-Research/Breeze-ASR-25` from Hugging Face (GGML version)
- Invalid file - fails gracefully with "Failed to load..." error, no crash
- Toggle off (default) - no custom models discovered
- Toggle persistence - setting saves correctly across restarts

1. Toggle off

Screenshot 2026-01-19 at 5 15 02 PM

2. Successfully loaded

Screenshot 2026-01-19 at 5 16 10 PM

3. Load failed

Screenshot 2026-01-19 at 5 16 20 PM

## AI Assistance

**If AI was used:**

- Tools used: Claude Code
- How extensively: Wrote most of the code and handled conflict resolution, 100% human-reviewed


cjpais:
I see you made some tweaks to the UI as well. You might wanna wait for #478! It will probably come in first and then I'll probably be pulling this in. That PR and probably subsequent PRs are going to be changing this bit of UI quite a bit.

cjpais:
@CJHwong mind refactoring this code, theres a new model ui

VirenMohindra (commented):
nice feature - the rust backend for model discovery is well-designed. `discover_custom_whisper_models()` function handles edge cases well (hidden files, non-`.bin` files, directories, predefined filenames). tests are a solid addition too!

that said, main has changed significantly since this PR was created as CJ mentioned (#478 merged a new models settings page). here's what needs updating for the rebase~

1. won't compile on current main
`ModelInfo` now has 3 additional **required** fields. `discover_custom_whisper_models` function needs to set these~

```rust
supports_translation: false,
is_recommended: false, 
supported_languages: vec![],  // or populate with whisper's supported languages
```

2. frontend likely needs a full rewrite
`ModelDropdown.tsx` was completely overhauled by by PR (sorry 😬 ) - it's now a simple dumb component that only shows downloaded models. all download management moved to a new `ModelsSettings` page

for the rebase i'd suggest~
- **ModelDropdown**: just add custom models alongside regular downloaded models with a "custom" badge. keep it minimal and simple
- **ModelsSettings**: add a "custom models" section here (this is where users manage models now)

3. `bindings.ts` should be regenerated

the rust backend logic is sound - main work is adapting the frontend to the new UI architecture. happy to help if you have questions about the new component structure

src/i18n/locales/en/translation.json
```diff
@@ -19,6 +19,7 @@
     "recommended": "Recommended",
     "download": "Download",
     "downloading": "Downloading...",
+    "customModelDescription": "Not officially supported",
```
VirenMohindra, line 23: missing i18n keys for all _new_ 15 non-English locales (`i18next` falls back to english, so not a user-crash, but `bun run check:translations` _will_ fail CI)

src-tauri/src/managers/model.rs
```diff
@@ -331,6 +340,121 @@ impl ModelManager {
...
+                    filename,
+                    url: None, // Custom models have no download URL
+                    size_mb,
+                    is_downloaded: true, // Already present on disk
+                    is_downloading: false,
+                    partial_size: 0,
+                    is_directory: false,
+                    engine_type: EngineType::Whisper,
+                    accuracy_score: 0.75,
+                    speed_score: 0.75,
```
VirenMohindra: the **hardcoded** is misleading for _unknown_ models - think we should hide the scores or using a sentinel

src/components/model-selector/ModelDropdown.tsx
```diff
@@ -35,9 +36,22 @@ const ModelDropdown: React.FC<ModelDropdownProps> = ({
...
   const { t } = useTranslation();
-  const availableModels = models.filter((m) => m.is_downloaded);
-  const downloadableModels = models.filter((m) => !m.is_downloaded);
-  const isFirstRun = availableModels.length === 0 && models.length > 0;
+
+  // Split models into three groups: downloaded (with URL), custom (no URL), downloadable
+  const downloadedModels = models.filter(
+    (m) => m.is_downloaded && m.url !== null,
+  );
+  const customModels = models.filter((m) => m.url === null);
```
VirenMohindra: this check to identify custom models is fragile - a dedicated `is_custom` field on `ModelInfo` would be more robust

src-tauri/src/managers/model.rs
```diff
@@ -331,6 +340,121 @@ impl ModelManager {
         Ok(())
     }
 
+    /// Discover custom Whisper models (.bin files) in the models directory.
+    /// Skips files that match predefined model filenames.
+    fn discover_custom_whisper_models(
+        models_dir: &PathBuf,
```
VirenMohindra: change `&PathBuf` to `&Path` in function signature per rust convention

cjpais:
Hey @CJHwong fwiw, I added breeze asr support directly, it'll be out in the next version

CJHwong:
@VirenMohindra Thanks for the thorough review! All three points addressed:

  1. Merged main and added the 3 missing fields (supports_translation, is_recommended, supported_languages) to custom model construction.
  2. Fully integrated with the new models page. I added a dedicated "Custom Models" section in ModelsSettings, and a "Custom" badge in ModelDropdown. Your new models settings page is genuinely great work btw, made it really easy to slot custom models in.
  3. Bindings updated with is_custom on ModelInfo, custom_models_enabled on AppSettings, and changeCustomModelsEnabledSetting command.

Also btw, fixed a pre-existing bug where the language filter disappears when no models match the selected language due to the filter being nested inside the downloaded models section, so it vanished along with it.

src-tauri/src/managers/model.rs
CJHwong: Fixed.

src/components/model-selector/ModelDropdown.tsx
CJHwong: Yep, way cleaner.

src-tauri/src/managers/model.rs
CJHwong: Good call. Changed to 0.0 sentinel for both scores. ModelCard now hides score bars entirely when both are 0, and also hides language tags when supported_languages is empty.

src/i18n/locales/en/translation.json
CJHwong: Added all 5 new keys to all 16 locales with proper translations. Tried my best validating them through Google Translate and LLM.

CJHwong:
> Hey @CJHwong fwiw, I added breeze asr support directly, it'll be out in the next version

Thanks for adding Breeze natively!

src-tauri/src/shortcut/mod.rs
```diff
@@ -978,3 +978,13 @@ pub fn change_app_language_setting(app: AppHandle, language: String) -> Result<(
...
+
+#[tauri::command]
+#[specta::specta]
+pub fn change_custom_models_enabled_setting(app: AppHandle, enabled: bool) -> Result<(), String> {
+    let mut settings = settings::get_settings(&app);
+    settings.custom_models_enabled = enabled;
+    settings::write_settings(&app, settings);
+
+    Ok(())
+}
```
VirenMohindra: **bug: app breaks on restart when custom model was selected.** if the user has a custom model selected and toggles this off, `selected_model` in settings still points to the custom model. on restart, discovery is skipped so the model doesn't exist → `Failed to load model: Model not found` and transcription is completely broken until the user manually selects another model.

**suggested fix:** when disabling, check if the currently selected model is custom. if so, reset `selected_model` to an empty string (which triggers `auto_select_model_if_needed` to pick a valid default on next startup):

```rust
if !enabled {
    let models = app.state::<ModelManager>().get_available_models();
    if let Some(model) = models.iter().find(|m| m.id == settings.selected_model) {
        if model.is_custom {
            settings.selected_model = String::new();
        }
    }
}
```

src-tauri/src/shortcut/mod.rs
```diff
@@ -978,3 +978,13 @@ pub fn change_app_language_setting(app: AppHandle, language: String) -> Result<(
...
     Ok(())
 }
+
+#[tauri::command]
+#[specta::specta]
+pub fn change_custom_models_enabled_setting(app: AppHandle, enabled: bool) -> Result<(), String> {
+    let mut settings = settings::get_settings(&app);
+    settings.custom_models_enabled = enabled;
+    settings::write_settings(&app, settings);
+
```
VirenMohindra: **bug: custom models remain visible in UI after toggle-off (without restart).** toggling off only persists the setting — it doesn't emit a `model-state-changed` event or remove custom models from the `ModelManager`'s cached `available_models`. the frontend never refreshes.

**suggested fix:** after `write_settings()`, either:
1. remove custom models from the `available_models` mutex and emit `model-state-changed` so the frontend refreshes, or
2. at minimum emit the event and have `get_available_models()` filter based on the current `custom_models_enabled` setting

VirenMohindra (changes requested):

VirenMohindra:
Screenshot 2026-02-08 at 11 09 37 AM

separately, @edwche10 wanna take a look at the custom models UI? this PR adds opt-in custom whisper model discovery (behind a debug toggle). a few design questions~
1. custom models are marked with a `CUSTOM` label inline with the model name in the dropdown. does this work or should we delineate differently?
2. section order is now Downloaded -> Custom -> Available to Download. custom models are local so they sit alongside downloaded models rather than available ones. does this work?
3. should we surface anything on the card to indicate user-sourced models? e.g. a reference link to where they got the model
4. custom models currently bypass the language filter (since we don't know what languages they support). is that fine for now, or should we eventually let users tag their custom models with metadata (languages, accuracy, etc) via some UI workflow? currently as it stands in the PR today we would need to drop the custom model directly inside the models/ directory which is not the _best_ UI

and finally maybe premature but: should we add a new filter section on the models page to filter against custom models? there's only ever going to be a handful of custom models, a filter section feels like overkill for a debug feature. i'd prob say YAGNI for now

src-tauri/src/shortcut/mod.rs
CJHwong: Fixed. The toggle-off path now resets `selected_model` when it's a custom model. Also hardened `auto_select_model_if_needed` to clear any stale selection that doesn't exist in `available_models`, which covers the case where a custom model file is deleted from disk too.

src-tauri/src/shortcut/mod.rs
CJHwong: Fixed. Toggle now adds/removes custom models from the mutex and emits `model-state-changed` so the UI refreshes instantly.

VirenMohindra: when the toggle is on but no custom .bin files exist in the models directory, the "Custom Models" section is either empty or hidden entirely with no guidance

think we should add a small info box / empty state in `ModelsSettings.tsx` that tells the user where to place their custom model files (e.g. "place .bin whisper models in ~/Library/Application Support/com.pais.handy/models/") so they're not left wondering why nothing shows up

VirenMohindra (approved):
LGTM! tested on MacOS and works well!

custom models with supported_languages: vec![] will get filtered out when the language filter is set to anything other than "all". the filteredModels memo filters by `model.supported_languages.includes(languageFilter)`, which returns false for an empty array

this means if a user selects e.g. "English" in the filter, their custom models disappear. **not a blocker**, but worth noting custom models could arguably bypass the language filter since their capabilities are unknown

PR #455, opened 2025-12-14
feat: add text replacements feature

schmurtzm:
## ✨ New Feature: Text Replacements

This new feature introduces a **simple and intuitive interface** for performing text replacements within Handy.

To help better understand this feature in practice, here is a tutorial video (with sound):

https://github.com/user-attachments/assets/73820022-d2cb-4ce6-b3f4-a334a83b4a1d

---

## 🎯 Purpose and Capabilities

The interface is designed to be **simple, offline, powerful, and intuitive**.  
It allows users to:

- Replace one term with another
- Perform text expansion
- Manage punctuation cleanly
- Insert the current date or time
- Execute command lines from a spoken trigger

---

## 🛠️ Replacement Management

It supports both **basic** and **advanced** usage through:

- Graphical options (trimming punctuation and/or whitespace before and after the matched text)
- Regular expression support
- Magic commands (currently a limited set is available, but they can be easily extended thanks to a simple and scalable code-side implementation)

The Replacements interface also provides tools to efficiently manage rules:

- Filter existing replacements
- Export and import replacement lists to enable easy sharing within the Handy community
- For each replacement, users can:
  - Enable or disable it
  - Edit it
  - Duplicate it
  - Delete it

The UI visually highlights:
- Regular expression enabled or not
- Trim punctuation or spaces enabled or not
- Magic command used or not
- Disabled replacements
- Duplicate rules
- Newly imported replacements

---

### Available Magic Commands

- `[lowercase]`  
Converts the entire current phrase to lowercase  
  - Replacement rule example:
"transform in lowercase" → [lowercase]
  - Result example (voice → text result):  
  `"Hello World transform in lowercase"` → `"hello world"`

- `[uppercase]`  
  Converts the entire phrase to UPPERCASE  
  Example:  
  `"Hello"` → `"HELLO"`

- `[capitalize]`  
  Capitalizes the first letter of each word  
  Example:  
  `"jean dupont"` → `"Jean Dupont"`

- `[nospace]`  
  Removes all spaces from the phrase  
  Example:  
  `"a test"` → `"atest"`

- `[date]`  
  Inserts the current date  YYYY-MM-DD (depending your current region settings)

- `[time]`  
  Inserts the current time (HH:MM)

- `[run]"command…"`  
  Executes the specified command and **prevents any transcription output** for that trigger
 The following placeholders can be used inside `[run]` commands:

- `{text}` — full phrase
- `{text_nospace}` — text without spaces
- `{text_nopunctuation}` — text without punctuation
[...]

PR #638, opened 2026-01-21
Add "external script" paste method

piec:
Thanks for making Handy, I've only discovered a few days ago and it's awesome.
A few months ago I searched for something like this but couldn't find anything.

## Before Submitting This PR

## Human Written Description

Adds a new paste method that uses an external script.
That gives me the flexibility to post process the text as I want without adding complexity to Handy.
Examples:
* when the focused window/tab is a chat I don't want capitals, or a final dot 
* depending on the language typography rules are not the same (space before "?")

## Related Issues/Discussions

I'm noticing these after making the PR oops :)
But they complement each other:
* https://github.com/cjpais/Handy/pull/455
* https://github.com/cjpais/Handy/discussions/168

## Testing

I have tested it for a few days on an Arch Linux / xorg / i3 setup.
I think it should work on windows / macos too but cannot test on these platforms.

<details open>
  <summary>Simple external script example</summary>
  
```python
#!/usr/bin/env python
import sys
import os
import subprocess

if len(sys.argv) <= 1:
        os.exit(1)

text = sys.argv[1]

# subprocess.run(['notify-send', '--', text])

# Make first character lowercase
if text:
    text = text[0].lower() + text[1:]

# Remove trailing period (but keep other punctuation)
if text.endswith('.'):
    text = text[:-1]

# French convention: add space before "?" and "!"
import re
text = re.sub(r'(?<! )([?!])', r' \1', text)

subprocess.run(['xdotool', 'type', '--', text], check=True)
```
</details>
I also have a more complex script where I change the behavior depending on the focused windows. With `xdotool getactivewindow getwindowname`.

image

## AI Assistance

**If AI was used:**

- Tools used: Claude code
- How extensively: Claude did the initial boilerplate for settings layout, which I simplified aferwards.

### DIFF
--- a/src/i18n/locales/de/translation.json
+++ b/src/i18n/locales/de/translation.json
@@ -261,8 +261,10 @@
           "clipboardCtrlShiftV": "Zwischenablage (Strg+Umschalt+V)",
           "clipboardShiftInsert": "Zwischenablage (Umschalt+Einfg)",
           "direct": "Direkt",
-          "none": "Keine"
-        }
+          "none": "Keine",
+          "externalScript": "Externes Skript"
+        },
+        "externalScriptPlaceholder": "/pfad/zu/ihrem/skript.sh"
       },
       "typingTool": {
         "title": "Eingabetool",
--- a/src/i18n/locales/fr/translation.json
+++ b/src/i18n/locales/fr/translation.json
@@ -261,8 +261,10 @@
           "clipboardCtrlShiftV": "Presse-papiers (Ctrl+Shift+V)",
           "clipboardShiftInsert": "Presse-papiers (Shift+Insert)",
           "direct": "Direct",
-          "none": "Aucun"
-        }
+          "none": "Aucun",
+          "externalScript": "Script externe"
+        },
+        "externalScriptPlaceholder": "/chemin/vers/votre/script.sh"
       },
       "typingTool": {
         "title": "Outil de frappe",
--- a/src/i18n/locales/it/translation.json
+++ b/src/i18n/locales/it/translation.json
@@ -261,8 +261,10 @@
           "clipboardCtrlShiftV": "Appunti (Ctrl+Shift+V)",
           "clipboardShiftInsert": "Appunti (Shift+Insert)",
           "direct": "Diretto",
-          "none": "Nessuno"
-        }
+          "none": "Nessuno",
+          "externalScript": "Script esterno"
+        },
+        "externalScriptPlaceholder": "/percorso/del/tuo/script.sh"
       },
       "typingTool": {
         "title": "Strumento di digitazione",


cjpais:
Can we make this a Linux only thing?

piec:
Yes, but wouldn't it be useful to macos too?
Windows users probably aren't used to do stuff like this.

cjpais:
I would rather keep options fairly slim where possible. For me everything works more or less perfect on MacOS. It's my primary machine and I have very few if any issues

piec:
@cjpais done
I only hid it in the UI, if in any case someone wants force it using the conf file

gabenasci:
Imo it would be nice to still be able to select one of the pre-existing paste methods but also run the script as well (for other use-cases other than just custom paste, like a custom post-transcription script that will run x y z commands). Lots of scripting possibilities here regardless of platform. 

If the user were to use this script to handle only pasting, they could set the paste method to none + add their script that will run after transcription with their own custom pasting method.

Maybe this could be hidden inside "experimental features" instead? Just a thought!

piec:
Hi @gabenasci, maybe we need 2 things:
* This PR: an "external script" paste method which can let users insert text with a custom tool. The intended usage is to insert text not to process text.
* Another PR: an "external script" for post-processing, handy runs it, captures the command output and carries on with other post processing and pasting.

cjpais:
@piec if you want this, could you rebase, I'll merge it as an option for Linux users

piec:
OK, will do :)

piec:
Hi @cjpais, rebased I also added the translations I knew but there are missing ones.

piec:
Related to this previous comment https://github.com/cjpais/Handy/pull/638#issuecomment-3785291802

Are you interested by a PR for processing recognized text in an external command?

cjpais:
@piec not at the moment, I'm mostly interested in PR's which are contributing to fixing the stability issues of the app. I will be putting a pause on most new features coming in.

suuuehgi:
@cjpais Either this is not working on my computer or or I don't get how it's supposed to be used.

I created a small test script `/tmp/test.sh`
```bash
#!/bin/bash
while IFS= read -r line; do
  echo "$line" >> /tmp/test.log
done
```
and set the paste method to the external script but no `/tmp/test.log` appears.

cjpais:
@piec please take a look, I am unable to test this

piec:
Hi @suuuehgi, is the script executable? `chmod +x ...`

Edit: Hmm also actually the text is passed as $1, not to stdin
It might have been better to pass it as process stdin to avoid cmdline max size, it's really huge these days though

suuuehgi:
> Edit: Hmm also actually the text is passed as $1, not to stdin

That was it. I just assumed it to read from stdin as there's no documentation about this feature whatsoever.

Regarding stdin / $1 off the top of my head, I don't have any preferences on this.

Arbitrarily long strings should use stdin but at the same time we're speaking about life transcriptions. That's probably a couple of sentences at max.
I think there's a hard limit of 128 KiB for arguments by the kernel but thats quite a bit of plaintext.

piec:
The only documentation so far is the python example in a collapsible section in the text of the PR at the top of the page 🙂

suuuehgi:
> in a collapsible section

Well hidden. 😅 I always overlooked that when skimming through it.

AlexanderYastrebov:
Hi, I think this is a useful feature (for macos users as well). 

> Adds a new paste method that uses an external script.
That gives me the flexibility to post process the text as I want without adding complexity to Handy.

The original need seems to be about processing the transcription rather that pasting it differently. Therefore I think this feature might be evolved/improved in the following way:
1. pass transcription to the script using stdin - this is a standard unix practice
2. in case exit code is 0, read script stdout and if its not empty - paste using selected method. This way script may preprocess transcription without bothering about paste method as well as use custom paste method by returning empty stdout. AI post processing or what not could be done by such a script as well.

cjpais:
I agree, but if you are advanced enough you can use it on macOS by editing the settings file. This is not a setting for average users and it shouldn't be in the UI

piec:
Hi I also thought making an additional PR to modify the code in order use stdin instead of the arg

AlexanderYastrebov:
> Therefore I think this feature might be evolved/improved

I've created https://github.com/cjpais/Handy/pull/930

piec:
Hello @AlexanderYastrebov, I also did something like this but a bit differently, I will share it on my fork :)

suuuehgi:
Okay, it took me a while to figure this out, but I came up with [this](https://github.com/cjpais/Handy/issues/684#issuecomment-4038938522) script now.
Maybe it is of help to someone.

Issue #641, opened 2026-01-21
[BUG]  It's crashing when you accidentally hit push-to-talk twice in a row.


edwinencomienda:
## Bug Description

The application is just hanging every time you do or trigger the push to talk twice.  And it will just not responding

## System Information

**App Version:**

0.7.0

**Operating System:**

macOS Tahoe
Version 26.2

**CPU:**

Apple M1 Max

**GPU:**

Built-In

Korben00:
I confirm the issue. same problem here on my Mac Studio M4 (Tahoe 26.2)

cjpais:
Thanks for this report. I thought this got solved, but looks like maybe it hadn't. It did take me a few attempts, but I was able to replicate this by just like spamming the button.

cjpais:
Thanks for the confirmations on the reports. Maybe the workaround for the time being is disabling push to talk. I definitely don't run into this in the other mode, which is my primary mode of using the app currently.

It'll probably take me a little bit to get to this, but I think in the worst case we can kind of do a bit of a hack which is just de bouncing the key on the back end and making sure that it just doesn't register the second time. But there's definitely a race condition here that is really what we should be fixing. @VirenMohindra If you're able to help on this and replicate, that would be super helpful as well

Korben00:
Hi, 
I needed it to work, so I made a small fix quickly, but I have no guarantee that there will be no side effect or the code is very clean... 'cause I'm not a developer ;-) . 
So out of respect for your project, I do not do PR but you can go there to look what I did : https://github.com/Korben00/Handy

cjpais:
@Korben00 if you submit a PR it's even better for me! I can review the changes all there at once and decide if I want to pull them in or make changes on them myself. I would really appreciate if you submitted a PR! It's the best way to contribute, I really welcome PRs of any kind even if they ultimately don't make it in for whatever reason. It makes my life much easier

joshribakoff:
If we can land the tests, it should help with thinking things are fixed and help reduce the noise of "you said this was fixed but fyi its not". Its always tricky to balance user reports with refactoring and testing, but I find that once tests and refactoring are prioritized the user reports tend to stabilize.

The first step might be to ensure we run our tests on each change https://github.com/cjpais/Handy/pull/671 and I have shared thoughts in a few different places on expanding the tests and filling our testing gaps.

luckman212:
[...]

PR #644, opened 2026-01-22
Fix recorder start/stop failures and debounce toggle shortcuts

Korben00:
I noticed that in some cases the recorder could fail the start/stop and leave the app in an inconsistent state (microstream stopped, toggle stuck). So I reinforced error management by restoring the microphone stream when it’s fairing, and I stabilized the toggle shortcut with a debounce to avoid double triggers.

## Related Issues/Discussions
Fixes #
Discussion: https://github.com/cjpais/Handy/issues/641

## Testing
- [x] Manual: bun run tauri dev
- [x] Tested toggle bindings (start/stop) and confirmed debounce prevents rapid double-trigger
- [x] Tested recorder start/stop failure paths (by forcing recorder errors / device changes) and confirmed microphone stream is restored when in AlwaysOn mode

## AI Assistance
- [x] AI was used

**If AI was used:**
- Tools used: Windsurf with ChatGPT 5.2 (IDE assistant)
- How extensively: How extensively: Used to brainstorm and implement parts of the code changes, and iterated on fixes/refactors.

### DIFF
--- a/src-tauri/src/shortcut/handler.rs
+++ b/src-tauri/src/shortcut/handler.rs
@@ -90,11 +90,19 @@ pub fn handle_shortcut_event(
                     .duration_since(SystemTime::UNIX_EPOCH)
                     .map_or(0, |d| d.as_millis() as u64);
                 let last_ms = states.last_trigger_ms.get(binding_id).copied().unwrap_or(0);
-                if now_ms.saturating_sub(last_ms) < 250 {
+                const DEBOUNCE_MS: u64 = 250;
+                if now_ms.saturating_sub(last_ms) < DEBOUNCE_MS {
                     return;
+                }
+                states
+                    .last_trigger_ms
+                    .insert(binding_id.to_string(), now_ms);
             }
-            states.last_trigger_ms.insert(binding_id.to_string(), now_ms);
-            } // Lock released here
+
+            states
+                .active_toggles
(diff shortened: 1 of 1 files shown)


joshribakoff:
Debouncing could make sense as a UX improvement, but it doesn't directly fix race conditions - it can mask them and make them harder to test.

virenpepper (approved):
hey, i reviewed this fix in detail and verified it compiles and works

the root cause is in `recorder.rs` - the consumer loop blocks on `sample_rx.recv()` waiting for audio, and only checks commands after receiving a sample. if the audio stream stops producing samples (race condition, device issue), `stop()` blocks **forever** because the command is never processed...

1. recv_timeout(50ms) instead of blocking recv() - critical fix
2. 5s timeout on stop() response - prevents infinite hang
3. debouncing in handler (250ms) - prevents rapid triggers
4. ui state changes after recording actually starts - correct state management
5. early exit in on_end if not recording - prevents invalid stop calls

one minor note: 250ms debounce might feel _slightly_ sluggish for power users, we should prob update to 150/200ms if possible

LGTM nice work! this should fix #641

cjpais:
Closing in favor of https://github.com/cjpais/Handy/pull/824, thank you Korben for submitting this! Sorry I didn't merge earlier

PR #659, opened 2026-01-24
feat: add Paste Last Transcription shortcut

pchalasani:
## Summary

Add a new global keyboard shortcut to paste the most recent transcription from history.

## Motivation

Sometimes the transcription does not make it to where you need it:

- **Clipboard overwritten**: You transcribe something, then copy something else before pasting - your transcription is gone from the clipboard
- **Paste failed**: The target app did not accept the paste (wrong focus, read-only field, etc.)
- **Need to paste again**: You want to paste the same transcription in multiple places
- **Clipboard mode "Do not Modify"**: If you use this setting to preserve your clipboard, you need another way to retrieve the transcription later

With this shortcut, you can always recover your last transcription without having to go to History, find the entry, and copy it manually.

## Implementation

- Default binding: `Ctrl+Option+V` (macOS) / `Ctrl+Alt+V` (Windows/Linux)
- Configurable in General settings alongside the existing Transcribe shortcut
- Added `PasteLastAction` in `actions.rs` that fetches the most recent history entry
- Uses the existing `utils::paste()` function, so it respects paste_method and append_trailing_space settings
- Prefers `post_processed_text` if available, falls back to `transcription_text`
- Silently does nothing if no history exists

## Bonus fix

Fixed TypeScript build errors in `ModelSelector.tsx` by adding explicit type annotations to immer `produce()` callbacks. (Note: upstream main currently fails `bun run build` without this fix due to missing type annotations.)

## Test plan

- [x] Build completes successfully
- [ ] New shortcut appears in General settings
- [ ] Shortcut can be customized  
- [ ] Pressing shortcut pastes the last transcription
- [ ] Works correctly when no history exists (no error)

### DIFF
--- a/src-tauri/src/actions.rs
+++ b/src-tauri/src/actions.rs
@@ -469,6 +469,62 @@ impl ShortcutAction for TestAction {
     }
 }
 
+// Paste Last Action
+struct PasteLastAction;
+
+impl ShortcutAction for PasteLastAction {
+    fn start(&self, app: &AppHandle, binding_id: &str, _shortcut_str: &str) {
+        debug!("PasteLastAction::start called for binding: {}", binding_id);
+
+        let app_clone = app.clone();
+        tauri::async_runtime::spawn(async move {
+            let hm = app_clone.state::<Arc<HistoryManager>>();
+
+            // Fetch history entries (already sorted by timestamp DESC, first is most recent)
+            match hm.get_history_entries().await {
+                Ok(entries) => {
+                    if let Some(entry) = entries.first() {
+                        // Prefer post_processed_text if available, otherwise use transcription_text
+                        let text_to_paste = entry
+                            .post_processed_text
+                            .as_ref()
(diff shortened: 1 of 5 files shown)


cjpais:
First thank you for the contribution! However, broadly speaking, new keyboard shortcuts are not going to be merged at this time. I'm more or less collecting all of the requests in https://github.com/cjpais/Handy/discussions/211

From there we will work on a UI/UX and figure out how we can support what people want. I think the main issue here is that the app is not consistently working, and we need to solve the root cause, not adding band-aid solutions unless we really have to

pchalasani:
Yep understood. Stability is paramount. Is there a specific issue that is tracking this? The only issue I notice (with Parakeet v3) is the frequent stuttering as I noted elsewhere.

cjpais:
I believe there's at least one or two open issues regarding this failure If I get a chance later I'll link them directly, but I think if you go through the issues tab you'll find a few that are similar.

PR #671, opened 2026-01-26
fix: run Rust tests in CI

### LINKED ISSUE #670: [BUG] CI doesn't run existing Rust tests
## Before You Submit

Searched existing issues - didn't find any about CI test coverage.

## Bug Description

The codebase has ~20 Rust unit tests but CI workflows don't run `cargo test`. Tests pass locally but regressions wouldn't be caught automatically.

## System Information

N/A - CI configuration issue, not platform-specific.

joshribakoff:
## Summary

Adds a separate `test.yml` workflow to run the existing ~20 Rust unit tests on PRs.

Fixes #670

## Trade-offs

I considered a few approaches:

1. **Adding to lint.yml** - Kept lint lightweight instead
2. **Adding to build.yml** - Would run tests on all 6+ platform matrix jobs (overkill)
3. **Separate test.yml** - Chose this, runs once on Ubuntu with required deps

The system dependencies (GTK, Vulkan SDK) are copied from build.yml. Open to feedback if you'd prefer a different approach.

## AI Disclosure

AI-assisted: Yes (Claude Code)

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -11,10 +11,13 @@ jobs:
         run: |
           sudo apt-get update
           sudo apt-get install -y libwebkit2gtk-4.1-dev libappindicator3-dev librsvg2-dev libasound2-dev libssl-dev
-          wget -qO- https://packages.lunarg.com/lunarg-signing-key-pub.asc | sudo tee /etc/apt/trusted.gpg.d/lunarg.asc
-          sudo wget -qO /etc/apt/sources.list.d/lunarg-vulkan-1.3.290-noble.list https://packages.lunarg.com/vulkan/1.3.290/lunarg-vulkan-1.3.290-noble.list
-          sudo apt-get update
-          sudo apt-get install -y vulkan-sdk mesa-vulkan-drivers
+
+      - name: Use mock TranscriptionManager (CI only)
+        working-directory: src-tauri
+        run: |
+          # Swap to mock adapter - avoids compiling whisper/Vulkan
+          cp src/managers/transcription_mock.rs src/managers/transcription.rs
+          sed -i '/^transcribe-rs/d' Cargo.toml
 
       - name: Run Rust tests
         working-directory: src-tauri
--- a/src-tauri/src/managers/transcription_mock.rs
+++ b/src-tauri/src/managers/transcription_mock.rs
@@ -1,0 +1,55 @@
+// CI-only mock TranscriptionManager - avoids whisper/Vulkan dependencies.
+// This file is copied over transcription.rs during CI tests.
+// Existing tests don't exercise transcription, so this is safe.
+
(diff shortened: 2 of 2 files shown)


joshribakoff:
It worked! 

This pulls in heavy dependencies, I'm going to make this better.

joshribakoff:
Hmm, that only saved 1 minute. 

Open question: Is optimal / acceptable for Rust unit tests? Should it be run on every single PR, or was unit tests in CI intentionally skipped?

.github/workflows/test.yml
```diff
@@ -0,0 +1,24 @@
...
+        run: |
+          sudo apt-get update
+          sudo apt-get install -y libwebkit2gtk-4.1-dev libappindicator3-dev librsvg2-dev libasound2-dev libssl-dev
+
+      - name: Use mock TranscriptionManager (CI only)
+        working-directory: src-tauri
+        run: |
+          # Swap to mock adapter - avoids compiling whisper/Vulkan
+          cp src/managers/transcription_mock.rs src/managers/transcription.rs
+          sed -i '/^transcribe-rs/d' Cargo.toml
```
joshribakoff, line 20: Using `sed` isn't ideal, but its just a quick way to do a certain pattern, for now.

This general pattern (not the specific approach) would be the plan as I continue making things testable, putting heavy dependencies behind interfaces and decoupling layers so we can focus on things like testing a state machine in isolation without calling out to heavy whisper C++ libraries or actually trying to draw panels on the screen, etc. 

In Michael Feather's book "Working Effectively with Legacy Code", which is a somewhat disparaging name he gives for code without tests, he documents this pattern and names it adding a seam -- a point where the behavior can be changed without editing the code.  


In this case, I'm just punching a hole in the part of the code that brings in a heavy dependency so that the unit tests don't pull in unnecessary dependencies and run slower. I'm not sure of a better way to add a seam here.

In future PRs I can potentially backfill more tests by adding more seams and testing functionality behind those seams, like an async operation manager / state machine (Right now, that same logic might be hard to test *if* it directly calls out to heavy dependencies in addition to managing state. ) This would continue to happen in small, safe, steps, and always open to feedback :)

src-tauri/src/managers/transcription_mock.rs
```diff
@@ -0,0 +1,55 @@
...
+    }
+
+    pub fn initiate_model_load(&self) {}
+
+    pub fn get_current_model(&self) -> Option<String> {
+        None
+    }
+
+    pub fn transcribe(&self, _audio: Vec<f32>) -> Result<String> {
+        Ok(String::new())
```
joshribakoff, line 53: We can also do things like have a mock that simulates a non `Ok` result, to exercise our edge cases. That will come in future PRs, as we build and ensure alignment on approach and pace.

cjpais:
Thanks for all of this, merging

PR #689, opened 2026-01-30
Feat/add remote desktop direct mode on wayland

petit-aigle:
## Before Submitting This PR

---

## Human Written Description

This PR introduces a Wayland-native keyboard injection path implemented on top of the GNOME Remote Desktop portal.  
It adds a fully portal-backed flow for direct typing, including explicit user authorization, permission lifecycle management, compositor-mediated input injection, onboarding integration, and settings UI feedback.

The objective is to enable direct text injection on Wayland using the mechanisms explicitly designed for it, while keeping the interaction constrained to a well-defined, user-approved security boundary.

Today, common Linux workarounds rely on installing tools such as `ydotool`, adding the user to the `input` group, and deploying permissive udev rules that grant write access to `/dev/uinput`. This setup enables global input injection for any process owned by the user, permanently widening the local attack surface and bypassing Wayland’s isolation guarantees.

By contrast, the portal-based approach keeps input injection scoped, revocable, and mediated by the compositor itself. The trust boundary remains explicit and auditable, aligning with Wayland’s original security goals rather than working against them (see discussion: https://github.com/cjpais/Handy/issues/522#issuecomment-3725129696).

The latest implementation also improves character reliability by reading the compositor-provided Wayland keymap and sending real keycodes when the requested character exists in the active layout. Characters that are not available in that keymap fall back to the standard Unicode input sequence. This avoids hardcoding layout-specific cases while still supporting accented characters, uppercase letters, digits, punctuation, and broader Unicode input.

It also supports portal-backed clipboard paste shortcuts (`Ctrl+V`, `Ctrl+Shift+V`, `Shift+Insert`) on Wayland, since `wtype`, `dotool`, and `ydotool` are not functional by default without specific system configuration. This allows these paste methods to work through the Remote Desktop portal as well. Co-developed with @D3nis.

---

## Related Issues / Discussions

Fixes #  
Discussion: https://github.com/cjpais/Handy/issues/522#issuecomment-3725129696

---

## Community Feedback

No dedicated feedback thread yet.  
This change is intentionally scoped to a specific security and architectural axis; broader feedback can be gathered after initial review if needed.

---

## ⚠️ Must be fixed (before merge)

- [x] Some characters with symbols still end up in the wrong position in the string  
- [x] Some characters with symbols don’t appear at all (e.g. “ô”, “û”)  
- [x] Potential issue with uppercase letters or numbers

###  Information about previous fixes

The initial problem was that some accented characters were arriving “split”: the letter on one side and the accent on the other (e.g., “é” sent as U+0065 for “e” then U+0301 for the acute accent). Sent separately, the accent was mispositioned or lost. I therefore glued these pieces back together before sending, so each accented character is typed in a single shot.

That introduced another issue: if the active keyboard language doesn’t include a character (e.g., a layout without accents), pressing that character is ignored.

The implementation now first resolves characters through the current Wayland compositor keymap and sends the corresponding physical keycode plus modifiers. This handles characters that are available in the active keyboard layout in the same way the user would type them manually.

When a character cannot be produced from the active keymap, the code falls back to the Ctrl+Shift+U Unicode input sequence followed by the character’s Unicode code and validation. This gives a layout-aware fast path while keeping a robust fallback for characters that are not present on the current keyboard layout.

---

## Testing

**Manual tests only**

### Authorization, Token & UI Consistency

- [x] Permission request accepted via the portal and correctly reflected in the UI
- [x] Permission request refused via the portal and correctly reflected in the UI
- [x] Token revocation while the application is running, with immediate UI state update
- [x] Token revocation while the application is closed, followed by application restart and verification that:
  - the stored token is checked against the Flatpak permission store at startup
  - the application state is updated accordingly
  - the UI initializes consistently with the result of this check
- [x] Consistent synchronization between in-memory state, persisted settings, and Flatpak portal permissions
- [x] Verification of correct character injection when the portal authorization is present
- [x] Verification of uppercase letters, digits, punctuation, accented French characters, Latin extended characters, and other Unicode characters
- [x] Verification that the onboarding flow surfaces Remote Desktop authorization when direct typing requires it
- [x] Verification that the settings UI reflects authorization state and allows requesting/revoking authorization
- [x] Verification of character injection and portal-backed paste shortcuts

### Useful commands (Flatpak permission store – Remote Desktop)

 **⚠️ Warning:** the commands below operate on the Flatpak permission store for the **Remote Desktop** portal.  
Some of them can remove **all** Remote Desktop authorizations, not just the ones created by Handy.  
Always inspect the current state before running destructive commands, and only proceed if you are sure no other Remote Desktop permissions are in use.

- `flatpak permissions remote-desktop`  
  Lists all entries stored in the Flatpak permission table for the Remote Desktop portal.

- `flatpak permission-remove remote-desktop <OBJECT_ID>`  
  Removes the Remote Desktop portal permission for a **single specific object/application ID**.  
  This is the recommended and safe way to revoke one authorization without affecting others.

- `flatpak permissions remote-desktop | awk '{print $2, ($3=="-"?"":$3)}' | xargs -n2 sh -c 'flatpak permission-remove remote-desktop "$0"'`  
  Iterates over the Flatpak permission table and removes multiple entries via `flatpak permission-remove`.  
  ❗**Highly destructive** ❗: this can wipe all Remote Desktop permissions at once.

Flatpak portal permissions rely on a stable application identifier derived from desktop integration (i.e. a `.desktop` entry).  
In an installed build, this identifier exists and allows Flatpak Permissions to associate Remote Desktop authorizations with the application.  
In a development context, this desktop integration does not exist, so no stable application ID can be resolved and the permission entry may be empty.  
This is expected behavior of the Flatpak permission system in dev environments, not a bug in the implementation.

---

## Platform Coverage

- [x] Linux (GNOME Wayland)
- [ ] Linux (non-Wayland / X11)
- [ ] macOS
- [ ] Windows

---

## Screenshots / Videos

Screenshots showing:
- Remote Desktop authorization in the onboarding permission flow
Remote Desktop authorization onboarding

- Typing authorization settings UI and live authorization state
Typing authorization settings state
Typing authorization settings request state
Typing authorization settings authorized state

- Updated permission and typing authorization flow
Capture vidéo du 2026-06-02 20-23-21

- Typing authorization is only shown when Paste Method is set to Direct and Typing Tool is set to Auto or Remote Desktop
image

- Typing authorization is shown when Wayland is detected for all method paste.
image

- Active typing session indicator (GNOME screen-sharing icon during injection)
handy_wayland_typing_session_indicator

Note: The GNOME screen-sharing indicator appears during injection and may linger briefly after completion. This is expected portal behavior and may warrant a short user-facing note in settings.

---

## AI Assistance

- [x] AI was used

**If AI was used:**

- Tools used: ChatGPT, Codex  
- How extensively: AI tools were used as assistance to get started and to accelerate development. Codex was used in particular to bootstrap and navigate Rust-specific aspects, as I am an experienced Python developer but not a Rust specialist. AI was also used to help generate and translate the newly introduced i18n strings across all supported languages. Each portion of source code generated by AI was analyzed, validated, maîtrisée, challenged, and modified where necessary to ensure a high level of code quality. All architectural and technical decisions were made by me, and I remain fully responsible for the final design, implementation choices, and code.

## Contributors

- @D3nis: add support portal paste shortcuts on Wayland
- @petit-aigle: main contributor


cjpais:
I think we should try to unify the UI for permissions giving into the onboarding similar to how it's done on Mac. Otherwise I'm excited about this PR

petit-aigle:
@cjpais Thanks for the feedback, glad to hear you’re excited about the PR 🙂

I’m not entirely sure what you’re referring to regarding the macOS onboarding flow. I don’t currently have access to a Mac, so I haven’t been able to observe or test how permissions are presented there in this project.

Otherwise, since opening the merge request, I’ve fixed two issues: first, a wording/translation issue where “GNOME Wayland” was used instead of simply “Wayland”, even though Wayland can be used outside of GNOME. Second, I fixed the handling of accented characters, which were previously not supported and are now correctly handled.

Despite this fix, two issues remain: 
1. some accents still end up in the wrong position in the string
2. some accented characters don’t appear at all, e.g., “ô” and “û”

I don’t yet have a reliable fix that doesn’t turn into a mess of edge-case handling.

cjpais:
@petit-aigle you can look at the UI code for this, even if you are unfamiliar. It should be relatively trivial to have it appear also for linux.

bkanuka:
FYI I pulled this down to test (couldn't get anything else to work in Ubuntu + Gnome + Wayland) and this worked perfectly! I also agree that this is the "right way" of doing things. Great work @petit-aigle :rocket:

cjpais:
@bkanuka thanks for the confirmation, I'll probably be taking a look today and trying to merge

delthas (commented):
The README should probably also be updated? Namely https://github.com/cjpais/Handy?tab=readme-ov-file#major-issues-help-wanted and https://github.com/cjpais/Handy?tab=readme-ov-file#linux-notes

delthas:
Tested on my end, works fine on GNOME, except that many/most characters are always dropped (running with a French keyboard layout and French words): all upper case `A` to `Z`, some diacritic letters `â`, `ê` (but `é` and `è` work), `Ç` (but `ç` works), `Ê`, `Â`, also numbers `0123456789`, and some puncutation `.` (but `,` and `'` work).

(Initially I thought the first letter of all my inputs was dropped, but that was just because the first letter was always capital, and upper case letters are dropped on my system.)

Possibly more but that's what I was able to find.

bkanuka:
Interesting! I didn't have that experience at all. Capitals, numbers, punctuation, all came through - but I only speak English

@delthas  could you test in English? Setting gnome to a English layout?  It would be interesting to know if that's the reason things are dropped for you.

petit-aigle:
@delthas @bkanuka 
@cjpais Just to reiterate, please don’t merge yet as mentioned earlier. These issues are still present. I’m going to look into this as soon as I have a bit of time.

> Despite this fix, two issues remain:
> 
>    1. some accents still end up in the wrong position in the string
>    2. some accented characters don’t appear at all, e.g., “ô” and “û”
> 
> I don’t yet have a reliable fix that doesn’t turn into a mess of edge-case handling.

petit-aigle:
Okay, I've fixed the problem normally.

The initial problem was that some accented characters were arriving “split”: the letter on one side and the accent on the other (e.g., “é” sent as U+0065 for “e” then U+0301 for the acute accent). Sent separately, the accent was mispositioned or lost. I therefore glued these pieces back together before sending, so each accented character is typed in a single shot.

That introduced another issue: if the active keyboard language doesn’t include a character (e.g., a layout without accents), pressing that character is ignored.

To work around this, I now use the Ctrl+Shift+U method (example: Ctrl+Shift+U, then 00E9, then Enter → é) followed by the character’s Unicode code, then validation. This types the symbol directly and works regardless of the chosen keyboard language. I think this is the best solution currently available for managing different keyboard configurations. 

@delthas I wasn’t able to reproduce the issue with numbers or with uppercase letters. I am also on GNOME with a French keyboard. Could you test the latest solution and let me know whether the other issues have already been fixed on your side, and whether the problems with numbers and uppercase letters still persist ? Thank you.

delthas:
Hi, I pulled, reran `bun run tauri dev` and am still getting the exact same issues with some characters not working (all upper case A to Z, some diacritic letters â, ê (but é and è work), Ç (but ç works), Ê, Â, also numbers 0123456789, and some puncutation . (but , and ' work)).

delthas:
FWIW I'm running the "French" layout (not "alt.", "legacy", etc.)

petit-aigle:
@delthas 
Ok, based on my investigation and your explanation, I have the impression that the **keysyms (sent letters)** only work if they actually represent a **real physical combination on your keyboard**.

Example:
Shift + a = A

Whereas normally, the **portal** (what is used to transmit key presses to **Wayland**) is supposed to accept the **final character directly**, without requiring a physical combination.
This would explain your behavior for everything related to **uppercase letters and numbers**.

But after the update, if that were the case, then **even accented characters should no longer work**, because special characters are sent as **non-ASCII**, so all accented characters are sent via their **Unicode** representation.

So for example using Ctrl + Shift + U followed by the Unicode value and then Enter.
Based on your examples, this gives

* Ctrl + Shift + U and ea and Enter for ê
* Ctrl + Shift + U and e8 and Enter for è
* Ctrl + Shift + U and e9 and Enter for é
* Ctrl + Shift + U and c7 and Enter for Ç
* Ctrl + Shift + U and c2 and Enter for Â
* Ctrl + Shift + U and cA and Enter for Ê

Basically, what this says is that **è and é work**, so sending **8 and 9 worked**, but they would not work when they are requested directly.

Then why do the others not work?

At this point I’m a bit lost. As I told you, I have the exact same setup as you, so I’m missing some information to go further.

Could you please tell me **which portal you are using**?
You can run:

```bash
ps -ef | grep xdg-desktop-portal-
```

to see the list of installed portals.

Also, in:

```
~/.config/xdg-desktop-portal/portals.conf
```

check whether you have any specific configuration.

With:

```bash
/usr/libexec/xdg-desktop-portal --verbose
```

you should normally see **which portal is actually loaded**.

Could you also please double-check that you have **pulled the latest version**, for example by verifying that:

```bash
git log -1
```

and that the latest commit is indeed:

```
4a00b6e2bc473234d3ae673eb2d833a09f4ca0df
```

Sorry to ask all this. I know there is almost no chance you are not on the correct version, but the fact that **é and è work** really makes it look like it might be the case.

Also, when you launch the new version, please make sure to **kill any Handy instance** that might still be running in the background.

petit-aigle:
@cjpais Hey !
I’ve updated the UI to take the “Typing Tool” into account (i’ve updated the GIF on the PR). I’m waiting for feedback from @delthas, but otherwise everything looks good on my side. I also took the opportunity to clean up the code a bit and better separate the different concerns.

cjpais:
I'm not super happy with the UI flow. Mostly because permissions are generally handled during onboarding (as mentioned before). I understand why they are here, but just wondering if you took a look or thought through how it could be done during the onboarding.

Also be careful with colors, the app pretty much uses exclusively Handy colors and best to consider this. The UI is generally a bit rough and not matching the rest of the app.

petit-aigle:
Ah yeah, that’s true, I’ll look into it as soon as possible, with all the small issues that came up in the meantime, I forgot to get back to this.

cjpais:
No worries thanks

bkanuka:
@panpanpanpaan The GlobalShorcuts portal is for registering keyboard shortcuts though (right?).  The Remote Desktop portal is an acceptable way of allowing a program to act on the user's behalf (in this case, "typing on a keyboard").

xezpeleta:
Thanks!!

I hope this PR can be merged soon, as I cannot get the direct paste method working with wtype on Ubuntu 22.04.

jurcicek:
It works like a charm. Great work! I hope it will get merged soon!

cjpais:
@petit-aigle this needs a rebase and ideally handling permissions flow the same as windows, macOS

petit-aigle:
@cjpais  I don’t have time to take care of this right now, but I haven’t forgotten. I’ll handle it as soon as I have some time freed up.

skaldebane:
I'm curious if this also works with CJK characters? Because these don't seem to work with `ydotool` at the moment, just like accented characters.

**Update:** Clipboard (Ctrl+V) seems to work perfectly for this, at least for my usages, including CJK and accented characters.

petit-aigle:
Hey @cjpais, hey everyone,

I was finally able to free up some time to work on this PR again.

I added the Remote Desktop authorization step to the onboarding flow, as requested. I also updated the PR description with the new screenshots and videos.

I also fixed the issues related to special characters.

For anyone curious about the technical details:

_The original issue came from relying too directly on character/keysyms injection. Depending on the active keyboard layout and compositor behavior, some characters were either dropped, misplaced, or not typed at all. This was especially visible with uppercase letters, digits, punctuation, and accented characters such as `â`, `ê`, `ô`, `û`, `Ç`, `Ê`, and `Â`._
    
_The new strategy is layout-aware. The code now reads the current Wayland compositor keymap, resolves each character to the real physical keycode and modifiers when possible, and sends those keycodes through the Remote Desktop portal. If a character cannot be produced from the active keymap, it falls back to the standard `Ctrl+Shift+U` Unicode input sequence._
    
 _This keeps the implementation portal-based and compositor-mediated, while avoiding hardcoded keyboard-layout special cases as much as possible._
 
 Please feel free to test it and let me know if you run into any issues.

cjpais:
Great, thank you for the update. I will definitely try and take a look soon Maybe one or two weeks. I'm just quite busy at the moment, but when I do get a solid chance. I will definitely Take a look. This is important and we should fix

matbgn:
@petit-aigle Do you have an idea why I get the following discrepancies while auto pasting?

image

Here the settings for the portal to capture my input:

image

petit-aigle:
@matbgn Thanks for the report.

From my investigation, the issue seems to be related to the speed at which characters are sent. Since we do not have an easy/reliable acknowledgement that a character has actually been processed by the target application, some characters can occasionally be handled faster than others, which creates the offset you are seeing.

We do not have precise control over the final processing order once the events are handed to the compositor/application stack, so I added a setting that lets users configure a small delay between key events. This helps ensure each action is processed one by one before the next one is sent.


Could you test again with 5ms first, which is the default value? If characters still get mixed, please increase the value and let me know which delay works reliably for you, and whether this solves the issue on your side.

Thanks again for the feedback.


Capture d’écran du 2026-06-10 17-50-10

> **Next step if this solves the issue:**
>
> Ideally, I would like `5ms` to be enough so we can hardcode it and avoid keeping this extra setting in the UI, since it adds a bit of complexity and users may not naturally find it when facing this issue.
>
> If one low default value is reliable everywhere, then we are good. If slower machines need more time, we should keep the setting so faster machines are not slowed down unnecessarily.

matbgn:
For a reason I don't get the pasting is no more working now, but I suspect something got weird on my machine even if the logs seems fine:

```
[2026-06-10][17:45:29][handy_app_lib::managers::transcription][INFO] Transcription completed in 126ms
[2026-06-10][17:45:29][handy_app_lib::managers::transcription][INFO] Transcription result: Test one two three test
[2026-06-10][17:45:29][handy_app_lib::clipboard][INFO] Using paste method: Direct, delay: 60ms
[2026-06-10][17:45:29][handy_app_lib::clipboard][INFO] Using user-specified Remote Desktop portal
```

Just a small question the logs says 60ms but the UI was set to 20ms, something to fix here?
image

I will not have time to test another round before a long time, so would be glad if someone can give it a shot.

Thanks @petit-aigle to have try to address my issue, which could still be on my own setup only btw.

petit-aigle:
@matbgn 
I understand the confusion: the `delay: 60ms` log is not the 'Typing Delay' shown in the UI.

That existing log comes from the generic paste flow in `clipboard.rs`:

https://github.com/cjpais/Handy/blob/0d29c929d093221e8034779c580cbb0ea2516ac4/src-tauri/src/clipboard.rs#L642-L645

It refers to `paste_delay_ms`, which is used by the general paste pipeline. It is not the new per-key Remote Desktop delay.

To make this clearer, I added a dedicated log for the Remote Desktop path itself, so the configured key event delay is now logged separately

> I amended the latest commit, so if you already pulled the branch before, please refresh it before testing.

D3nis:
**Tested on Fedora 43 / GNOME Wayland (no X11 available) — found a performance issue and a fix**

**Environment:**
- Fedora 43, GNOME 47, Wayland only (X11 session removed in Fedora 43)
- CPU: Intel Core Ultra 9 285K, no GPU
- Keyboard layouts: Russian + German (no English layout installed)

**Problem with the current PR #689 approach:**

The character-by-character typing via Remote Desktop portal is extremely slow — approximately 3-4 characters per second for a typical dictation result (~150 chars). A single voice transcription took **45 seconds** to paste, which makes it unusable for dictation.

**Root cause:** Each character requires a separate D-Bus round-trip with a key press + key release + delay. For non-ASCII characters (Cyrillic, German umlauts) that aren't in the current keyboard layout, it falls back to `Ctrl+Shift+U` + hex digits — which is even more key events per character.

**Fix: clipboard + single Ctrl+V via Remote Desktop portal**

Instead of typing character-by-character, write the full text to clipboard via `wl-copy`, then send a single `Ctrl+V` through the Remote Desktop portal. This gives **~250ms** paste time regardless of text length or keyboard layout.

Tested with:
- Russian text, RU keyboard layout active ✓
- Russian text, DE keyboard layout active ✓
- German text with umlauts (ä, ö, ü, ß), DE layout ✓

**Patch (2 files, ~40 lines):**

In `try_send_key_combo_linux()` in `clipboard.rs`, add Remote Desktop portal as the first option for Wayland before trying `wtype`/`dotool`/`ydotool`:

```rust
if is_remote_desktop_available() {
    info!("Using Remote Desktop portal for key combo");
    remote_desktop::send_ctrl_v()?;
    return Ok(true);
}
```

Add `send_ctrl_v()` to `remote_desktop.rs` — sends keycode 29 (Ctrl) + 47 (V) via `notify_keyboard_keycode`.

Full diff:

```diff
diff --git a/src-tauri/src/clipboard.rs b/src-tauri/src/clipboard.rs
index 97501e4..12b228e 100644
--- a/src-tauri/src/clipboard.rs
+++ b/src-tauri/src/clipboard.rs
@@ -87,7 +87,13 @@ fn paste_via_clipboard(
 #[cfg(target_os = "linux")]
 fn try_send_key_combo_linux(paste_method: &PasteMethod) -> Result<bool, String> {
     if is_wayland() {
-        // Wayland: prefer wtype (but not on KDE), then dotool, then ydotool
+        // Wayland: prefer Remote Desktop portal (works with any layout/language),
+        // then wtype (but not on KDE), then dotool, then ydotool
+        if is_remote_desktop_available() {
+            info!("Using Remote Desktop portal for key combo");
+            remote_desktop::send_ctrl_v()?;
+            return Ok(true);
+        }
         // Note: wtype doesn't work on KDE (no zwp_virtual_keyboard_manager_v1 support)
         if !is_kde_wayland() && is_wtype_available() {
diff --git a/src-tauri/src/remote_desktop.rs b/src-tauri/src/remote_desktop.rs
index d9730bd..d5a9a2c 100644
--- a/src-tauri/src/remote_desktop.rs
+++ b/src-tauri/src/remote_desktop.rs
@@ -629,3 +629,42 @@ pub fn init_authorization(app: &AppHandle) {
+
+/// Sends a Ctrl+V keystroke through the Remote Desktop portal.
+#[cfg(target_os = "linux")]
+pub fn send_ctrl_v() -> Result<(), String> {
+    if !crate::utils::is_wayland() {
+        return Err("not running on Wayland".into());
+    }
+    if !get_authorized() {
+        return Err("authorization not granted".into());
+    }
+    block_on_portal(|| async {
+        let settings = PORTAL_APP_HANDLE.get().map(crate::settings::get_settings);
+        let delay = settings
+            .as_ref()
+            .map(|s| s.remote_desktop_key_event_delay_ms)
+            .unwrap_or(crate::settings::DEFAULT_REMOTE_DESKTOP_KEY_EVENT_DELAY_MS);
+
+        let (proxy, session) = open_session_async(false).await?;
+
+        // KEY_LEFTCTRL=29, KEY_V=47
+        proxy.notify_keyboard_keycode(&session, 29, KeyState::Pressed).await
+            .map_err(|e| format!("ctrl_v: ctrl press failed: {}", e))?;
+        if delay > 0 { tokio::time::sleep(Duration::from_millis(delay)).await; }
+
+        proxy.notify_keyboard_keycode(&session, 47, KeyState::Pressed).await
+            .map_err(|e| format!("ctrl_v: v press failed: {}", e))?;
+        if delay > 0 { tokio::time::sleep(Duration::from_millis(delay)).await; }
+
+        proxy.notify_keyboard_keycode(&session, 47, KeyState::Released).await
+            .map_err(|e| format!("ctrl_v: v release failed: {}", e))?;
+        if delay > 0 { tokio::time::sleep(Duration::from_millis(delay)).await; }
+
+        proxy.notify_keyboard_keycode(&session, 29, KeyState::Released).await
+            .map_err(|e| format!("ctrl_v: ctrl release failed: {}", e))?;
+
+        let _ = close_session_async(&session).await;
+        Ok(())
+    })
+}
```

Happy to open a separate PR on top of this one if useful.

koloved:
It looks like this requires testing on KDE Wayland , Judging by the number of users KDE should be fixed accordingly. It would be good to have some kind of unified fix, or two very similar fixes that represent a single way of resolving issues. More precisely, one could say a unified interface.

D3nis:
@koloved 
Thank you for the feedback! I've updated my PR to include a `kwtype` fallback for KDE Wayland.

The updated patch is in PR  https://github.com/cjpais/Handy/pull/1545

The fallback chain in `try_send_key_combo_linux()` is now:

1. **Remote Desktop portal** — compositor-agnostic, works on both GNOME and KDE Wayland
2. **KDE Wayland fallback**: `kwtype` if portal is unavailable  
3. **GNOME/other Wayland**: `wtype` → `dotool` → `ydotool`

I don't have a KDE setup to test the `kwtype` fallback — if someone with KDE Plasma Wayland could verify, that would be great.

petit-aigle:
@D3nis Thanks a lot for the detailed testing and patch. This is an excellent idea.

I integrated and tested your commit, and I allowed myself to add a few small fixes to keep it aligned with the current PR:

- the Remote Desktop portal key combo path now follows the selected paste method instead of always sending `Ctrl+V`. It supports `Ctrl+V`, `Ctrl+Shift+V`, and `Shift+Insert`;
- the typing authorization UI is now shown on Wayland for these clipboard paste modes too, not only for `Direct`.

image


To keep this within the scope of the same feature, I also renamed it so it better matches the current PR as a whole.

@D3nis when you have a chance, feel free to retest and confirm whether this works correctly on your setup.

D3nis:
> **Note:** PR #1545 is not a replacement for PR #689 — it builds *on top of* it. PR #689 introduces the Remote Desktop portal infrastructure (authorization, session management, character injection). PR #1545 adds one additional function `send_ctrl_v()` that reuses that same portal infrastructure for clipboard-based paste, which is ~180x faster for typical dictation output.

Without PR #689 merged first, PR #1545 has no portal infrastructure to rely on. They are complementary, not competing.

Thanks for the detailed writeup — very helpful to understand the full picture!

After testing PR #689 on Fedora 43 / GNOME Wayland with Russian and German keyboard layouts, I ran into the performance issue: character-by-character injection via the portal was ~3-4 chars/sec, making a 150-char transcription take ~45 seconds. The `Ctrl+Shift+U` fallback for non-layout characters made it even slower.

I think both approaches have their place:

- **Your approach (character-by-character)**: correct for apps that don't support clipboard paste (some terminal emulators, password fields, etc.)
- **Our approach (clipboard + single Ctrl+V via portal)**: ~250ms regardless of text length or keyboard layout — better for the common dictation use case

I've opened PR #1545 which adds `send_ctrl_v()` via the Remote Desktop portal as the **first option** in the `paste_method: ctrl_v` flow, while keeping your character-by-character implementation intact as the `paste_method: direct` path. The two approaches coexist cleanly.

Regarding your open issue with uppercase letters/numbers — we observed similar artifacts (`и43fсользовать`) when using `paste_method: direct` with the portal. Switching to `paste_method: ctrl_v` resolved this completely, which suggests the issue may be specific to the character-by-character path rather than the portal itself.

PR #1545: https://github.com/cjpais/Handy/pull/1545

petit-aigle:
@D3nis Thanks for the clarification.

I understood your earlier comment as leaving open the possibility of integrating this directly here, and since I think this is a very good addition to the same Wayland portal work, I went ahead and included it in PR #689.

That said, since this initiative comes from you, I’ll leave the final choice to you. Personally, I think it is coherent to keep everything in the same PR because it relies on the same Remote Desktop portal authorization/session infrastructure. But if you prefer keeping this work only in PR #1545, just let me know and I’ll remove my commit from this PR.

Before doing that, I would suggest cherry-picking or porting the follow-up adjustments I added:
- the portal key combo now follows the selected paste method: `Ctrl+V`, `Ctrl+Shift+V`, or `Shift+Insert`;
- the typing authorization UI is now also shown for these Wayland clipboard paste modes, not only for `Direct`;
- a few internal wording/comments were generalized from “direct typing” to “keyboard input”.

Let me know which option you prefer, and I’ll adapt the branch accordingly.

D3nis:
Thank you for including this and for the improvements — especially supporting all paste methods and showing the authorization UI for clipboard modes, which I had missed (no use cases for me)

I'm happy to keep everything in PR #689. It makes sense architecturally since it all relies on the same portal infrastructure. PR #1545 can be closed.

One small request — if possible, could you add a co-author attribution to the commit?

```
Co-authored-by: D3nis <p.d@gmx.de>
```

This is optional of course, but it would be nice to have the contribution reflected in the commit history. ;)
Thanks again for the collaboration!

petit-aigle:
Of course, that’s fair. I should have done it automatically, sorry about that.
I also took the liberty of doing the opposite: I set you as the main author of the commit and added myself as co-author. ;)

koloved:
I created a flatpack with this PRm. It works very slowly, and the insertion happens differently in different applications. As you can see, there are errors @petit-aigle @D3nis 

thats flatpack: [handy_0.8.3_x86_64.zip](https://github.com/user-attachments/files/29412608/handy_0.8.3_x86_64.zip)


https://github.com/user-attachments/assets/6a8f44e4-9e76-46fc-84db-0e2f5d4c26cc


also notice there is Ю paste instead of . (dot) sometimes
ok i understand , the paste symbols depends on keyboard layout, thats the problem

PS: In my opinion, this 'remote desktop portal' is just another paste method that happens to work for me in Flatpak, but it still comes with a lot of issues. Out of all the available text pasting methods, the only one that actually works for me in Flatpak on an atomic system is this one: https://github.com/cjpais/Handy/pull/1395. With this patch, pasting is instant.

D3nis:
@koloved ,  The issue you're experiencing is exactly what we solved. The problem is the **paste method**, not the portal itself.

**What's happening:** You're using `paste_method: direct` which types character-by-character through the portal — this is slow AND layout-dependent (that's why you get "Ю" instead of "." when Russian layout is active).

**The fix:** Switch to `paste_method: ctrl_v` with `typing_tool: auto`. This uses:
1. `wl-copy` to put text in clipboard instantly
2. A single `Ctrl+V` keystroke via the portal (~250ms total, layout-independent)

In your Handy settings, change **Paste Method** to **Clipboard (Ctrl+V)**.

Or edit `~/.var/app/computer.pais.Handy/data/com.pais.handy/settings_store.json` directly (stop Handy first):
```json
"paste_method": "ctrl_v",
"typing_tool": "auto"
```

This is exactly what PR #1545 / the updated PR #689 implements. With this configuration:
- Russian text, RU layout ✓ ~250ms
- Russian text, DE layout ✓ ~250ms  
- German umlauts (ä, ö, ü, ß) ✓ ~250ms

The layout no longer matters because the text goes through the clipboard, not character-by-character keystrokes.

---

**For the global hotkey on GNOME Wayland:**

Set **Keyboard Implementation** to **Handy Keys** in Handy settings, then add a Custom Shortcut in GNOME Settings → Keyboard → Custom Shortcuts:

```
Name:    Handy toggle
Command: handy --toggle-transcription
Shortcut: Alt+X  (or any key you prefer)
```

This works because Handy Keys listens to input directly, while GNOME shortcut triggers `handy --toggle-transcription` as an external command. No need for `ydotool` or any other input tool for the hotkey itself.

koloved:
@D3nis Thanks for the explanation. I just didn't know what settings to configure, so I pulled in another patch ([https://github.com/cjpais/Handy/pull/1395](https://www.google.com/search?q=https%3A%2F%2Fgithub.com%2Fcjpais%2FHandy%2Fpull%2F1395)) just to be able to use the app. Now I’m not sure which of these the program is actually using. I really just want to have a working, flawless app that just works in Wayland. Especially for atomic systems, which are becoming more and more relevant nowadays—like SteamOS. I believe getting onto Flathub is incredibly important

I build another Flathub version for testing. Now it runs entirely on Wayland without X11, and it can read hotkeys. It will ask you to enable them on the first run. https://github.com/cjpais/Handy/pull/1287

[handy_0.8.3_x86_64.zip](https://github.com/user-attachments/files/29419328/handy_0.8.3_x86_64.zip)

dansleboby:
Just checking in on this PR, I’m really looking forward to this feature being merged.

Wayland input is still a major pain point for me, and this PR seems to address it in a much cleaner way than the current workarounds. From the discussion, it looks like most of the functional issues have already been addressed.

Is the main remaining blocker now just rebasing/resolving conflicts with `main` and getting a final review?

Thanks to everyone who has worked on this.

cjpais:
I will try and take a look soon, it's 2500 lines of code, so I need to take a lot of time to review. I don't like to pull in things I haven't tested. It takes a lot of time for changes of this size

Haptein:
> @D3nis Thanks for the explanation. I just didn't know what settings to configure, so I pulled in another patch ([https://github.com/cjpais/Handy/pull/1395](https://www.google.com/search?q=https%3A%2F%2Fgithub.com%2Fcjpais%2FHandy%2Fpull%2F1395)) just to be able to use the app. Now I’m not sure which of these the program is actually using. I really just want to have a working, flawless app that just works in Wayland. Especially for atomic systems, which are becoming more and more relevant nowadays—like SteamOS. I believe getting onto Flathub is incredibly important
> 
> I build another Flathub version for testing. Now it runs entirely on Wayland without X11, and it can read hotkeys. It will ask you to enable them on the first run. #1287
> 
> [handy_0.8.3_x86_64.zip](https://github.com/user-attachments/files/29419328/handy_0.8.3_x86_64.zip)

This has been working flawlessly for me, finally. I tried it on Fedora 43 with Gnome/Wayland. Fwiw, in my case the default 5ms delay was overkill and made text output quite slow, 0ms worked just fine.

PR #704, opened 2026-02-03
added more 3 vars for LLM post-processing transcript

khanhicetea:
## Before Submitting This PR

## Human Written Description

We have more variables in prompt when post-processing:

$time_local : current time local string
$current_app : use for identify the purpose of work, can teach the LLM about in prompt template
$short_prev_transcript : a trim about 200 words last transcript in same app (or expired in 5 mins), so LLM has more context about what next.
$language : Selected transcription language for Whisper models (e.g., "en", "zh-Hans"), or "auto" for other models (Parakeet, Moonshine)

## Related Issues/Discussions

Fixes #168 
Discussion:

## Testing

Using this prompt:

```
Current App : ${current_app}
Current Local Time : ${time_local}

Clean this transcript:
1. Fix spelling, capitalization, and punctuation errors
2. Convert number words to digits (twenty-five → 25, ten percent → 10%, five dollars → $5)
3. Replace spoken punctuation with symbols (period → ., comma → ,, question mark → ?)
4. Remove filler words (um, uh, like as filler)
5. Keep the language in the original version of ’Transcript' (if it was french, keep it in french for example)

Preserve exact meaning and word order. Do not paraphrase or reorder content.

Return only the cleaned transcript.

Last transcript in same app (for better understanding context) : ${short_prev_transcript}

Transcript:
${output}
```

## AI Assistance

**If AI was used:**

- Tools used: Claude Opus 4.5 Thinking
- How extensively:


khanhicetea:
btw, I plan to have 1 more var named '${language}' which is picked by user in settings for multi-lang model. Should I push it ?

cjpais:
Sure. If you can, can you also add documentation of this feature in the readme or somewhere

khanhicetea:
- Added $language (for Whisper models), fallback into 'auto' for everything else
- Added docs into readme (**Experimental > Post Processing** section)

I tested on my mac works (not sure about Win and Linux)

This template can provide more ways to do with transcribe (fix typo, grammar, even translation or dual translation output) 😄

src-tauri/src/active_app.rs
```diff
@@ -0,0 +1,124 @@
+/// Module for getting the frontmost/active application name.
+/// This is platform-specific and returns the name of the application
+/// that has keyboard focus when the user starts transcribing.
+
+#[cfg(target_os = "macos")]
+pub fn get_frontmost_app_name() -> Option<String> {
```
VirenMohindra, line 7: hmm so macOS impl returns app name via `localizedName` (ie "Safari"), but Windows uses `GetWindowTextW` which returns the window title (ie "Stack Overflow - How to do X - Google Chrome"). this means `${current_app}` has different semantics per platform, and `${short_prev_transcript}` context keying by app name would create separate entries per browser tab on Windows. maybe consider using `GetModuleFileNameExW` or querying the process name instead?

src-tauri/src/actions.rs
```diff
@@ -90,9 +110,20 @@ async fn maybe_post_process_transcription(
         provider.id, model
     );
 
-    // Replace ${output} variable in the prompt with the actual text
-    let processed_prompt = prompt.replace("${output}", transcription);
```
VirenMohindra: the change logs the entire prompt including transcription text AND app name. this is a privacy concern since the log file could contain sensitive speech content

recommend reverting

src-tauri/Cargo.toml
```diff
@@ -95,6 +95,7 @@ windows = { version = "0.61.3", features = [
 
 [target.'cfg(target_os = "macos")'.dependencies]
 tauri-nspanel = { git = "https://github.com/ahkohd/tauri-nspanel", branch = "v2.1" }
+objc = "0.2"
```
VirenMohindra, line 99: crate is deprecated in favor of `objc2` which is already a transitive dep in the project. maybe we look into `objc2-app-kit` or a simpler approach for the macOS active app detection

src-tauri/src/actions.rs
```diff
@@ -3,21 +3,26 @@ use crate::apple_intelligence;
...
 use once_cell::sync::Lazy;
 use std::collections::HashMap;
-use std::sync::Arc;
+use std::sync::{Arc, Mutex};
 use std::time::Instant;
 use tauri::AppHandle;
 use tauri::Manager;
 
+/// Tracks the frontmost application captured at recording start, keyed by binding_id
+static RECORDING_APP_CONTEXT: Lazy<Mutex<HashMap<String, String>>> =
```
VirenMohindra, line 25: nit: entries are inserted in `start()` and removed in `stop()`, but `CancelAction` doesn't clean up its entry. over many cancelled recordings this could accumulate stale entries. minor since the values are small strings, but worth adding cleanup in the cancel path too

VirenMohindra:
solid feature idea, the context variables will make post-processing much more useful. some notes below~

**merge conflicts + coordination with #706**

this PR is currently conflicting against main and also has a semantic conflict with #706 (structured outputs). that PR splits prompts into system prompt + user content, so the variable substitutions here (`${current_app}`, `${time_local}`, etc) will need to work in both the legacy `${output}` path and the structured output system prompt path. recommend merging #706 first and rebasing this on top

cjpais:
I rebased, I haven't tested. @VirenMohindra mind taking another look. I know #770 is also somewhere along this track as well

PR #706, opened 2026-02-03
feat: implement structured outputs for post-processing providers

Schreezer:
## Summary

This PR implements structured outputs support for Cerebras, OpenRouter, OpenAI, and Apple Intelligence providers in the post-processing pipeline.

### Core Implementation
- **Structured Outputs**: Added JSON schema support to enforce consistent response format
- **System Prompt Architecture**: Changed from `` variable substitution to system prompt + user content approach
- **Provider Support**: 
  - Cerebras ✅
  - OpenRouter ✅
  - OpenAI ✅
  - Apple Intelligence ✅

### Technical Details
- Added `send_chat_completion_with_schema()` in `llm_client.rs` for structured output support
- Updated `actions.rs` to detect supported providers and use structured outputs
- Modified Apple Intelligence Swift code to accept separate system prompts
- All providers now use consistent format: system instructions + user transcription

### Benefits
- **Type Safety**: Guaranteed JSON schema adherence
- **Consistent Format**: No more parsing errors or hallucinated fields
- **Better Architecture**: Clean separation between instructions and content
- **Fallback Support**: Graceful degradation if JSON parsing fails

## Testing
- All 23 existing tests pass ✅
- No compiler warnings ✅
- Code compiles successfully on macOS with Apple Intelligence support

## Notes
- Legacy providers (Anthropic, Groq, Custom) continue using `` variable substitution
- Apple Intelligence uses native Swift structured outputs via `CleanedTranscript` struct

### DIFF
--- a/src-tauri/src/settings.rs
+++ b/src-tauri/src/settings.rs
@@ -481,7 +481,7 @@ fn default_post_process_providers() -> Vec<PostProcessProvider> {
             base_url: "https://api.groq.com/openai/v1".to_string(),
             allow_base_url_edit: false,
             models_endpoint: Some("/models".to_string()),
-            supports_structured_output: true,
+            supports_structured_output: false,
         },
         PostProcessProvider {
             id: "cerebras".to_string(),


Schreezer:
It can be easily tested by setting the provider to either Cerebras, OpenRouter, OpenAI, or Apple Intelligence. Most of these providers now offer models that support structured output, so simply running the model with one of these providers should be a sufficient test. The reason for adding this was that many times, when using cheap LLMs, it caused issues because 10-20% of the time, it failed to return the correct format. It’s much better now.

cjpais:
Thanks this is good, I will take a look and merge when I get a chance

VirenMohindra (commented):
nice work on this! the architectural shift to system prompt + user content is solid and the separation is actually better for prompt injection resistance too. went through the code and have some feedback below~

**settings migration for existing users**

`supports_structured_output` is added with `#[serde(default)]` which means existing users who upgrade will have all providers deserialized with `false`. `ensure_post_process_defaults()` checks if a provider exists by ID but does **NOT** sync new fields onto existing entries. so the feature will be dead on arrival for anyone who isn't a fresh install. think we need a migration step that updates the field on existing providers from the defaults

**per-provider vs per-model structured output support**

marking OpenRouter and Groq as `supports_structured_output: true` at the provider level is risky. OpenRouter's [docs](https://openrouter.ai/docs/guides/features/structured-outputs) say only select models support `json_schema`. Groq's [docs](https://console.groq.com/docs/structured-outputs) limit it to specific models (llama 4 variants, kimi k2). users who pick an unsupported model will hit API errors. could either make this per-model, or (simpler) add a fallback to legacy mode when the structured output call fails

**fallback on API error**

when the structured output API call fails (~line 180 in actions.rs), it returns `None` and the user gets raw unprocessed text. would be more resilient to fall back to the legacy `${output}` substitution path so the user still gets post-processing even if structured outputs aren't supported for their specific model.

**ordering note**

PR #704 also modifies `actions.rs` with new template variables (`${current_app}`, `${time_local}`, etc). those will need to work in both the legacy path and the structured output system prompt path. whichever merges second will need to account for this. happy to help coordinate

src-tauri/src/actions.rs
```diff
@@ -27,6 +27,21 @@ pub trait ShortcutAction: Send + Sync {
...
 
+/// Strip invisible Unicode characters that some LLMs may insert
+fn strip_invisible_chars(s: &str) -> String {
+    s.replace(['\u{200B}', '\u{200C}', '\u{200D}', '\u{FEFF}'], "")
+}
+
+/// Build a system prompt from the user's prompt template.
+/// Replaces `${output}` with a reference to the user message so the prompt
+/// still reads sensibly when the transcription is sent separately.
+fn build_system_prompt(prompt_template: &str) -> String {
```
VirenMohindra, line 55: replacing `${output}` with `"the user's message"` makes the system prompt read a bit oddly, especially with the default template which ends with `Transcript:\nthe user's message`. since the transcription arrives as a separate user message, the system prompt just needs the instructions. would `.replace("${output}", "").trim()` be cleaner here? then the system prompt is purely instructional and the LLM sees the transcription as the user message naturally.

src-tauri/src/settings.rs
```diff
@@ -408,34 +410,39 @@ fn default_post_process_providers() -> Vec<PostProcessProvider> {
...
             models_endpoint: Some("/models".to_string()),
+            supports_structured_output: true,
         },
         PostProcessProvider {
             id: "openrouter".to_string(),
             label: "OpenRouter".to_string(),
             base_url: "https://openrouter.ai/api/v1".to_string(),
             allow_base_url_edit: false,
             models_endpoint: Some("/models".to_string()),
+            supports_structured_output: true,
```
VirenMohindra, line 468: not all OpenRouter models support `json_schema` response format. the [docs](https://openrouter.ai/docs/guides/features/structured-outputs) note it's per-model, and some providers behind OpenRouter may fall back to `json_object` mode or reject the parameter entirely. same concern applies to Groq (line 437) — only specific models support structured output per their [docs](https://console.groq.com/docs/structured-outputs).

src-tauri/src/actions.rs
```diff
@@ -90,63 +105,137 @@ async fn maybe_post_process_transcription(
...
+                }
+            },
+            "required": ["transcription"],
+            "additionalProperties": false
+        });
+
+        match crate::llm_client::send_chat_completion_with_schema(
+            &provider,
+            api_key,
+            &model,
```
VirenMohindra, line 187: this returns `None` on API failure, which means the user gets raw unprocessed text. since the most likely failure mode is the model not supporting `response_format`, falling back to the legacy `${output}` substitution path would give users post-processing even when structured outputs aren't available for their model. something like:
```rust
Err(e) => {
    warn!("Structured output failed for '{}': {}. Falling back to legacy mode.", provider.id, e);
    // Fall through to legacy path below
}
```

VirenMohindra: im gonna send examples of this if possible and steps to repro

Schreezer: no need... i get the point, so i will apply the fallback mechanism

cjpais:
I suspect this will cause some issues, but since it's alpha we will deal with it when it happens.

PR #731, opened 2026-02-07
feat: add "Unload Model" in tray

AlexP11223:
## Before Submitting This PR

## Human Written Description

Added "Unload Model" button in the tray menu, similar to the "Unload model" after X minutes in the settings, just for manually unloading at any time.
This can be useful when starting doing something that requires the video memory occupied by the model, such as launching a game.

## Testing

1. Transcribe something so that the model is loaded. (press ctrl+space)
2. Click "Unload Model" in tray.
3. See that the model is unloaded and the video memory is freed.
4. Click "Unload Model" in tray again.
5. See a log message that there is no model loaded. 

image

## AI Assistance

**If AI was used:**

- Tools used: Claude Code (Opus)
- How extensively: I used it for planning and generating the code, because I do not have much experience with Rust and not familiar with the project. But everything was reviewed and tested.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -105,7 +105,7 @@ jobs:
         if: contains(inputs.platform, 'ubuntu-24.04') && !contains(inputs.platform, 'arm')
         run: |
           sudo apt-get update
-          sudo apt-get install -y libappindicator3-dev librsvg2-dev patchelf libasound2-dev libopenblas-dev libx11-dev libxtst-dev libxrandr-dev libgtk-layer-shell-dev \
+          sudo apt-get install -y libappindicator3-dev librsvg2-dev patchelf libasound2-dev libopenblas-dev libx11-dev libxtst-dev libxrandr-dev libgtk-layer-shell0 libgtk-layer-shell-dev \
             libwebkit2gtk-4.1-0=2.44.0-2 \
             libwebkit2gtk-4.1-dev=2.44.0-2 \
             libjavascriptcoregtk-4.1-0=2.44.0-2 \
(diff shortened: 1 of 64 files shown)


AlexP11223:
> nice work! imo this seems niche enough to warrant throwing it in the Advanced panel page. we lean and try and avoid over cluttering the tray with specific settings and keeping it super simple

The Advanced tab only contains settings, maybe it would be weird to add an action there. And it takes much more time to open, making it inconvenient for those who need it. Since they may need to use it often, such as every day (e.g. playing a game in the evening) or several times per day,

cjpais:
okay, just stepping back, is auto-model unloading not good enough? this setting already exists. I have mine set at immediately fwiw, so it never consumes memory in the background

im still open to pulling this in, but as mentioned in discord, we should think about greying this out if the model is already unloaded. sure it may do nothing, but its not very helpful for the user. it will definitely not be pulled in without this.

also fwiw, it might make sense to instead: add a model selector in the tray, which then has 'none' as an option to unload the model.

AlexP11223:
> okay, just stepping back, is auto-model unloading not good enough? this setting already exists. I have mine set at immediately fwiw, so it never consumes memory in the background

I think it can feel less responsive. :thinking: And probably does not make much sense to do all these operations every time, such as battery consumption etc.

> we should think about greying this out if the model is already unloaded

implemented

cjpais:
@AlexP11223 just curious if you have experienced it feeling less responsive, or is this speculation? on my system it in not noticeable as the model loads in 300-600ms and this happens as soon as you hit the button to start the microphone. unless you are speaking something under 1second it should not be noticeable. my system is very fast though so maybe the model load time could be much higher on other systems? mostly just curious

thanks ill merge soonish

AlexP11223:
I remember reading about it being an issue here https://www.ksred.com/soupawhisper-how-i-replaced-superwhisper-on-linux/ , but maybe it also depends on the Whisper implementation, so maybe it is slower for `faster-whisper`.
For me, it seems similar to your system. But still loading all that every time I dictate something feels weird. :D

cjpais:
They have a pretty naive way of model stuff I think, the mic/model loading/inference path in Handy is fairly well optimized for speed and should be relatively well suited for streaming in the future

PR #748, opened 2026-02-09
docs: clarify Linux gtk-layer-shell runtime requirements

datenzar:
## Summary
- document the Linux startup failure caused by missing `libgtk-layer-shell.so.0` in `README.md` with distro-specific runtime package names and install commands
- align `BUILD.md` Linux dependency instructions so Debian/Ubuntu and Fedora users install the required `gtk-layer-shell` runtime packages
- note the Ubuntu/Debian build-time dev package separately to reduce confusion between runtime and development dependencies

## Why
Handy links `gtk-layer-shell` on Linux, but current docs did not clearly call out the runtime package on all distros. Users can hit a hard loader failure before app startup, so this change provides direct, actionable install guidance.

### DIFF
--- a/BUILD.md
+++ b/BUILD.md
@@ -32,12 +32,13 @@ This guide covers how to set up the development environment and build Handy from
   ```bash
   # Ubuntu/Debian
   sudo apt update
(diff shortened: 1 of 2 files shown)


cjpais:
Thanks

Issue #750, opened 2026-02-09
[BUG] handy 0.7.2 can't start on Ubuntu 22.04


niko-lay:
## Bug Description

handy 0.7.2 can't start on Kubuntu 22.04 on X11, previous 0.7.1 works fine. Error is
```
handy 
handy: error while loading shared libraries: libgtk-layer-shell.so.0: cannot open shared object file: No such file or directory
```
Installing package `libgtk-layer-shell0` helps, even it useless on X11. Looks like .deb dependencies should be updated 
## System Information

**App Version:**

Handy_0.7.2_amd64.deb

**Operating System:**

```
lsb_release -a
No LSB modules are available.
Distributor ID:	Ubuntu
Description:	Ubuntu 22.04.5 LTS
Release:	22.04
Codename:	jammy
```

**CPU:**

Intel(R) Core(TM) i5-1135G7 CPU 

**GPU:**

datenzar:
Hi,

I also stumbled over this issue this mrorning. The problem is a newly introduced dependency which isn't in the packages yet. This might fix your issue (see #748)
`sudo apt install libgtk-layer-shell0`

cjpais:
@datenzar i bet we need to update the CI dependencies as well, do you mind taking a look

datenzar:
Sure, but it'll be my first contribution in that domain. I was just about to finish the AUR update

cjpais:
@niko-lay Would you mind trying 0.7.3? I think this got fixed by @datenzar

niko-lay:
Unfortunately 0.7.3 version has same issue, .deb package does not contains dependency for `libgtk-layer-shell0`
```shell
$ dpkg-deb --info  ~/Downloads/Handy_0.7.3_amd64.deb 
 new Debian package, version 2.0.
 size 20147514 bytes: control archive=905 bytes.
     198 bytes,     9 lines      control              
    1508 bytes,    20 lines      md5sums              
 Package: handy
 Version: 0.7.3
 Architecture: amd64
 Installed-Size: 46753
 Maintainer: cjpais
 Priority: optional
 Depends: libappindicator3-1, libwebkit2gtk-4.1-0, libgtk-3-0
 Description: Handy
  (none)
```
but binary is linked with this lib
```shell
$ ldd `which handy`
	linux-vdso.so.1 (0x00007ffcfa92d000)
	libgtk-layer-shell.so.0 => not found
[...]

PR #752, opened 2026-02-09
fix(ci): install gtk-layer-shell runtime package for Ubuntu builds

datenzar:
## Summary
- Add `libgtk-layer-shell0` to Ubuntu 22.04, Ubuntu 24.04, and Ubuntu 24.04 ARM dependency install steps in CI.
- Keep `libgtk-layer-shell-dev` while explicitly ensuring the runtime shared library package is present.
- Add a verification step on Ubuntu CI jobs that checks both package install status and `libgtk-layer-shell.so.0` visibility.

## Why
- Recent release packaging missed the runtime dependency required for `libgtk-layer-shell.so.0`, causing Linux startup/runtime failures.
- This hardens CI so missing runtime package regressions are caught during builds.

## Validation
- Local sanity check: `git diff --check`


cjpais:
thanks kicked off a test build for this

datenzar:
> thanks kicked off a test build for this

Ok, I tried it on my repo but it seems that there's more config required to make it pass.
https://github.com/datenzar/Handy/actions/runs/21825443709

cjpais:
Didn't check the logs but maybe remove the verification of the .so

datenzar:
> Didn't check the logs but maybe remove the verification of the .so

Ok, here's a commit with disabled signing in Build Test workflow
https://github.com/datenzar/Handy/actions/runs/21827598579

If that suceeds, we shoul be fine.

cjpais:
It looks like its due to you not having the signing keys which makes sense :P

cjpais:
My test build ran and you should be able to check if it works properly on Ubuntu here: https://github.com/cjpais/Handy/actions/runs/21826158058

datenzar:
2 warnings in the `ubuntu-24.04-arm` build but that should be unrelated. I think we're good.

cjpais:
were you able to download it and test on your machine, just want to confirm @datenzar

datenzar:
I did a smoke test and can confirm that the AppImage of `handy-pr-752-ubuntu-24.04-x86_64-unknown-linux-gnu.zip` is working. Unfortunately I don't have my env setup to test the other distros.

cjpais:
That's fine as long as it works we will call it good enough hopefully. Should close #750

PR #572, opened 2026-01-12
feat(linux): add Wayland support with GNOME system shortcuts

juulieen:
## Summary
- Add detection of Wayland session for UI decisions
- Add GNOME keyboard shortcut configuration via gsettings (uses SIGUSR2 signal)
- Add Wayland-specific UI in settings with warning notice explaining the limitation
- Disable Push-to-Talk on Wayland (only toggle mode available)

## Context
On Wayland, global shortcuts don't work natively due to security restrictions. This PR provides a workaround by allowing users to configure a GNOME system shortcut that sends SIGUSR2 to toggle transcription.

## Test plan
- [x] Test on Wayland session (Ubuntu 24.04+)
- [x] Verify GNOME shortcut configuration works
- [x] Verify Push-to-Talk is disabled with explanation
- [ ] Verify non-Wayland systems are unaffected

🤖 Generated with [Claude Code](https://claude.ai/code)

PR #759, opened 2026-02-10
feat(linux): add SIGUSR1 signal for transcribe_with_post_process

### LINKED ISSUE #758: Add SIGUSR1 signal support for transcribe_with_post_process
## Summary

PR #355 added a dedicated keyboard shortcut (`Ctrl+Shift+Space`) for `transcribe_with_post_process`, and the action is properly registered in the `ACTION_MAP`. However, the signal handler in `signal_handle.rs` still only maps `SIGUSR2` to `"transcribe"`. There is no signal-based way to trigger `transcribe_with_post_process`.

## Use Case

On Linux desktop environments where Tauri's global shortcut capture doesn't work reliably (e.g., Pop!_OS, certain GNOME/Wayland setups), users rely on system-level custom keyboard shortcuts that send `pkill -SIGUSR2 -x handy` to trigger transcription. This is also the approach documented in the README for Sway and the approach taken by PR #572 for Wayland/GNOME integration.

Currently, these users can only trigger basic transcription via signal. There is no way to trigger `transcribe_with_post_process` externally, since no second signal is registered.

## Proposed Solution

Add `SIGUSR1` handling in `signal_handle.rs`, mapped to `"transcribe_with_post_process"`:

- `SIGUSR1` → `"transcribe_with_post_process"` (transcription + AI post-processing)
- `SIGUSR2` → `"transcribe"` (basic transcription, unchanged)

This mirrors the existing `SIGUSR2` implementation pattern and would allow users to set up two system-level shortcuts:

```
# Basic transcription
pkill -SIGUSR2 -x handy

# Transcription with post-processing
pkill -SIGUSR1 -x handy
```

## Impact

This would also benefit PR #572 (Wayland/GNOME system shortcuts), which currently only configures `SIGUSR2` for basic transcription.

arimxyer:
## Summary

- Adds `SIGUSR1` signal handling that maps to the `transcribe_with_post_process` action, enabling Linux users to trigger transcription with post-processing via system signals (e.g. `pkill -SIGUSR1 -x handy`)
- Mirrors the existing `SIGUSR2` toggle pattern (start/stop with `ManagedToggleState`)
- Signal mapping: **SIGUSR1** → `transcribe_with_post_process`, **SIGUSR2** → `transcribe` (unchanged)

Closes #758

## Context

Linux desktop users (Pop!_OS, GNOME, Wayland) rely on `pkill -SIGUSR2 -x handy` for basic transcription because Tauri's global shortcut capture doesn't always work under Wayland. This adds a second signal so they can also trigger post-processing transcription without a keyboard shortcut, which also benefits Wayland/GNOME support efforts (PR #572).

## Changes

- **`src-tauri/src/signal_handle.rs`**: Import `SIGUSR1`, add match arm mapping it to `transcribe_with_post_process`
- **`src-tauri/src/lib.rs`**: Import `SIGUSR1`, register it in `Signals::new` alongside `SIGUSR2`

## Test plan

- [ ] Verify `pkill -SIGUSR2 -x handy` still toggles basic transcription (no regression)
- [ ] Verify `pkill -SIGUSR1 -x handy` toggles transcription with post-processing
- [ ] Confirm both signals respect the toggle state (send twice = start then stop)

### DIFF
--- a/README.md
+++ b/README.md
@@ -120,9 +120,9 @@ Without these tools, Handy falls back to enigo which may have limited compatibil
 - If you are having trouble with the app, running with the environment variable `WEBKIT_DISABLE_DMABUF_RENDERER=1` may help
 - You can manage global shortcuts outside of Handy and still control the app via signals, which lets Wayland window managers or other hotkey daemons keep ownership of keybindings:
 
-  | Signal    | Action                              | Example                         |
-  | --------- | ----------------------------------- | ------------------------------- |
-  | `SIGUSR2` | Toggle transcription                | `pkill -USR2 -n handy`         |
+  | Signal    | Action                                    | Example                |
+  | --------- | ----------------------------------------- | ---------------------- |
+  | `SIGUSR2` | Toggle transcription                      | `pkill -USR2 -n handy` |
   | `SIGUSR1` | Toggle transcription with post-processing | `pkill -USR1 -n handy` |
 
   Example Sway config:


cjpais:
Thanks, I simplified and rebased. This is a good and obvious change to have.

PR #765, opened 2026-02-10
feat: add configurable auto-submit after transcription paste

sasha-computer:
## Summary
- Port PR #189 onto `v0.7.3` by adding an opt-in Auto Submit setting that sends Enter/Ctrl+Enter/Cmd+Enter after transcription paste.
- Refine Auto Submit UX to a single dropdown with an explicit `Off` option (default), then `Enter`, `Ctrl+Enter`, and OS-aware `Cmd+Enter`/`Super+Enter` labels.
- Update tooltip and i18n copy (en/es/fr/vi) to clarify Meta-key behavior across macOS vs Windows/Linux while preserving existing paste and clipboard handling behavior.

## Test plan
- [x] `cargo test --manifest-path src-tauri/Cargo.toml`
- [x] `bun run lint`
- [x] `bun run build`
- [x] `bun run check:translations`
- [x] `bun run test:playwright`
- [x] Manual local app validation of Advanced > Output > Auto Submit behavior on macOS (Off/Enter/Ctrl+Enter/Cmd+Enter)

Made with [Cursor](https://cursor.com)

### DIFF
--- a/src/i18n/locales/ar/translation.json
+++ b/src/i18n/locales/ar/translation.json
@@ -230,6 +230,17 @@
           "copyToClipboard": "نسخ إلى الحافظة"
         }
       },
+      "autoSubmit": {
+        "title": "إرسال تلقائي",
+        "description": "إرسال مجموعة المفاتيح المحددة تلقائياً بعد إدراج النص. Cmd+Enter ينطبق على macOS، بينما يستخدم Windows/Linux مفتاح Super+Enter.",
+        "options": {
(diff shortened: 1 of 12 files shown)


cjpais:
I tested this on my machine and it works great and the UI and UX make sense to me.

sasha-computer:
> I tested this on my machine and it works great and the UI and UX make sense to me.

That's great, and I've already downloaded the new release. Thanks so much - what a time to be alive for OSS.

Issue #779, opened 2026-02-10
[BUG] Enigo-based paste methods (CtrlV, CtrlShiftV, ShiftInsert) cause TranscriptionManager stall on Linux Wayland


arimxyer:
## Description

On Linux Wayland (COSMIC desktop / cosmic-comp), all enigo-based paste methods (`ctrl_v`, `ctrl_shift_v`, `shift_insert`) cause the TranscriptionManager to stall/deadlock on subsequent transcription attempts. The first transcription completes successfully, but the second `TranscribeAction::start` hangs indefinitely — no transcription output, no error, the app just stops responding to transcription requests.

`paste_method: "none"` and `paste_method: "direct"` (wtype) do **not** exhibit this issue.

## Environment

- **Handy version:** 0.7.3
- **OS:** Pop!_OS 24.04 (Ubuntu-based)
- **Desktop:** COSMIC (smithay-based compositor)
- **cosmic-comp:** `0.1~1770304347~24.04~0e97ddb`
- **wayland-protocols:** 1.45
- **Kernel:** 6.18.7-76061807-generic
- **Trigger method:** SIGUSR2 via COSMIC keyboard shortcut

## Steps to Reproduce

1. Set `paste_method` to `ctrl_shift_v` (or `ctrl_v` / `shift_insert`)
2. Set `clipboard_handling` to `copy_to_clipboard`
3. Trigger a transcription — first one completes successfully
4. Trigger a second transcription — `TranscribeAction::start` hangs with no output

## Log Evidence

**With `paste_method: CtrlShiftV` — stalls on second attempt:**

```
[17:43:15] TranscribeAction::start called for binding: transcribe
[17:43:15] TranscribeAction::start completed in 232ms
[17:43:16] Shutting down TranscriptionManager
[17:43:19] Transcription result: Testing, testing. One, two, three.
[17:43:19] Using paste method: CtrlShiftV, delay: 60ms
[17:43:21] Idle watcher thread shutting down gracefully
[17:43:21] Shutting down TranscriptionManager
[17:43:28] TranscribeAction::start called for binding: transcribe
           ← NOTHING AFTER THIS — stalled/deadlocked
```

**With `paste_method: Direct` (wtype) — works consecutively:**

```
[17:49:43] TranscribeAction::start called for binding: transcribe
[17:49:45] Transcription result: Testing.
[17:49:45] Using paste method: Direct, delay: 60ms
[...]

PR #781, opened 2026-02-11
try minimal fix

cjpais:
for #779


boeserwolf:
Tested - works like a charm!

Issue #308, opened 2025-11-06
[BUG] Cannot get default EGL display: EGL_SUCCESS


jesses2025smith:
## Before You Submit

**Please search [existing issues](https://github.com/cjpais/Handy/issues) to avoid duplicates.** Your bug may already be reported! Right now it's just me maintaining this project so many issues can be overwhelming! Help me out by checking first.

## Bug Description
```shell
./Handy_0.5.5_amd64_linux.appimage
Cannot get default EGL display: EGL_SUCCESS
Cannot create EGL context: invalid display (last error: EGL_SUCCESS)
[1]    23588 segmentation fault (core dumped)  ./Handy_0.5.5_amd64_linux.appimage
```

## System Information
**App Version:**
Handy_0.5.5_amd64_linux.appimage

**Operating System:**
Linux cachyos-x64 6.17.6-2-cachyos

**CPU:**
Intel(R) Core(TM) Ultra 9 275HX

**GPU:**
NVIDIA Corporation GB206M [GeForce RTX 5060 Max-Q / Mobile] (rev a1)

cjpais:
Hard to say for sure but would guess this might be similar to #33, likely a CI issue

cjpais:
@jesses2025smith there was a fix in the most recent version that might help, would you mind giving it a try? otherwise I will probably end up closing the issue

jesses2025smith:
@cjpais 
```
./Handy_0.6.8_amd64.AppImage
Cannot get default EGL display: EGL_SUCCESS
Cannot create EGL context: invalid display (last error: EGL_SUCCESS)
[1]    435044 segmentation fault (core dumped)  ./Handy_0.6.8_amd64.AppImage

```

cjpais:
@jesses2025smith thank you for the confirmation! Will keep it open

jesses2025smith:
```
./Handy_0.6.10_amd64.AppImage
Cannot get default EGL display: EGL_SUCCESS
Cannot create EGL context: invalid display (last error: EGL_SUCCESS)
[1]    21272 segmentation fault (core dumped)  ./Handy_0.6.10_amd64.AppImage
```

jesses2025smith:
I cloned the source code and executed `cargo tauri dev` locally without any problems. I believe the issue is with the versions of the compilation dependencies.
```shell
vulkan-headers 1:1.4.335.0-1
vulkan-icd-loader 1.4.335.0-1.1
vulkan-intel 1:25.3.2-2
[...]

PR #789, opened 2026-02-11
fix: Apply DMABUF workaround on all Linux systems

suuuehgi:
The previous implementation only applied the workaround on X11 sessions, leaving Wayland users affected by WebkitGTK DMABUF crashes. This affects various GPU configurations (at least Intel and NVIDIA) on both X11 and Wayland.

This change applies the workaround unconditionally on Linux, ensuring the app doesn't crash on affected systems while having minimal performance impact.

Related:

- https://github.com/tauri-apps/tauri/issues/9394
- https://github.com/cjpais/Handy/issues/308

### DIFF
--- a/src-tauri/src/main.rs
+++ b/src-tauri/src/main.rs
@@ -4,12 +4,9 @@
 fn main() {
     #[cfg(target_os = "linux")]
     {
-        if std::path::Path::new("/dev/dri").exists()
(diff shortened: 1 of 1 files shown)


cjpais:
Thanks for this. I'm just gonna yolo merge it because I believe it's the appropriate change, and we will just have to find out what happens for people. But I suspect it will resolve more issues than it hurts.

PR #796, opened 2026-02-12
feat(i18n): add Traditional Chinese (zh-TW) locale

parkghost:
## Summary
- Add complete Traditional Chinese (zh-TW) translation for the entire UI
- Fix locale detection to preserve full locale codes (e.g., `zh-TW` instead of truncating to `zh`)
- Update language matching to try exact match first, then fall back to prefix match
- Register `zh-TW` in language metadata with proper priority ordering


cjpais:
you are awesome, I'm in Taiwan right now and was thinking this needed to happen!

PR #770, opened 2026-02-10
feat: add hidden OCR template context for post-processing

evrenesat:
## Summary

This PR introduces an experimental OCR context path for post-processing prompts across supported desktop platforms.

When the selected post-processing prompt contains `${OCR}` or `${ocr}`, Handy captures text from the frontmost app context (OCR) and injects it into the prompt template before sending to the post-processing model.

This is intentionally a hidden/experimental feature for now:
- No UI changes
- No new settings UI
- No public frontend command/API changes

Note: If guided, I can make follow-up changes needed for merge readiness (including UI exposure if desired).

## Intent

The goal is to improve post-processing quality by adding optional visual context from the active window/screen, especially for fixing terminology and typo-like dictation misunderstandings.

I’m using the following addition to the default prompt and getting better results than non-OCR post-processing:

```text
Relevant context (OCR’ed from active window, use for fixing typos and terminology):
${OCR}
```

I intentionally kept scope minimal and hidden to validate usefulness first, without increasing UI complexity or feature surface.

## Architecture / Flow

- Existing backend/frontend architecture remains unchanged (Rust backend in `src-tauri`, React/TS frontend in `src`).
- Prompt template expansion in `src-tauri/src/actions.rs` now supports:
  - `${output}` (existing transcript variable)
  - `${OCR}` / `${ocr}` (new OCR context variable)
- OCR is attempted only when:
  1. OCR variable exists in the selected prompt
  2. Existing `experimental_enabled` setting is true
- OCR text is truncated to 8000 chars to limit prompt growth.
- On OCR failure, missing permission, or unsupported conditions, OCR resolves to an empty string and processing continues safely.

## Platform Notes

<details>
<summary><strong>macOS</strong></summary>

- Added:
  - `src-tauri/swift/macos_ocr.swift`
  - `src-tauri/swift/macos_ocr_bridge.h`
  - `src-tauri/src/macos_ocr.rs`
- Updated:
  - `src-tauri/build.rs` (bridge build/link wiring)
  - `src-tauri/src/lib.rs` (module registration, macOS-gated)
  - `src-tauri/src/actions.rs` (template expansion + OCR integration + tests)
- Frameworks linked:
  - Foundation
  - AppKit
  - CoreGraphics
  - Vision
- Permission behavior:
  - Screen recording permission requested on demand
  - Request-at-most-once per app run
  - Graceful fallback to empty OCR context if denied/unavailable

</details>

<details>
<summary><strong>Windows</strong></summary>

[...]

PR #808, opened 2026-02-12
 feat(windows): add hidden OCR context capture from frontmost window for post-processing

evrenesat:
## Summary
  This PR adds Windows support for the hidden OCR context template variable (`${OCR}` / `${ocr}`) used in post-processing prompts.(Windows counter-part of #770)

  The captured text is read from the current foreground window and injected only into the prompt payload.

  ## Why
  This improves prompt quality for post-processing while keeping the app local-first and extensible:
  - Better context-aware rewriting/correction prompts
  - No cloud dependency added
  - Clear platform-specific implementation that is easy to fork/extend

  ## What changed
  - Added Windows OCR capture module:
    - `src-tauri/src/windows_ocr.rs`
  - Wired Windows OCR into prompt-building flow:
    - `src-tauri/src/actions.rs`
  - Registered Windows OCR module:
    - `src-tauri/src/lib.rs`
  - Added required Windows API feature flags:
    - `src-tauri/Cargo.toml`

  ## Behavior
  - OCR context is only used if prompt contains `${OCR}` or `${ocr}`.
  - If OCR capture fails, processing continues with empty OCR context.
  - On unsupported platforms for this implementation path, OCR context resolves to empty string.

  ## Testing
  Tested on Windows:
  1. Configure a post-processing prompt containing `${OCR}`.
  2. Start transcription with post-processing enabled.
  3. Verified prompt receives OCR text from frontmost window.
  4. Verified fallback to empty OCR context when capture fails.
  5. Verified normal transcription flow is unchanged.

Note: I have built and manually tested this on a Windows 11 ARM guest running on MacOS. Even with help of Codex, it took hours to build on arm64. I tested that post-processing works both with and without OCR prompts. In my limited testing, with OCR, model consistently managed to use correct symbols in dictation result (e.g. createUpdaterArtifacts instead of "create-updater-artifacts" or "Create Updater Artifacts" or "create updater artifacts").

  ## Scope / Non-goals
  - This PR is Windows-only OCR.
  - No local machine build-workaround changes are included in this PR. (hopefully)
  - No updater/signing config changes included.

  ## Breaking changes
  None.

  ## AI Assistance Disclosure
  - AI used: **Yes**
  - Tools used: **GPT-5 Codex / ChatGPT**
  - Extent: **Assisted with implementation scaffolding, conflict resolution, and debugging workflow; final code reviewed and validated manually.**

Note: I'm planning to try to create another PR for Linux as well, using Tesseract.

### DIFF
--- a/src-tauri/Cargo.toml
+++ b/src-tauri/Cargo.toml
@@ -87,10 +87,18 @@ tauri-plugin-updater = "2.9.0"
 
 [target.'cfg(windows)'.dependencies]
 windows = { version = "0.61.3", features = [
+  "Foundation",
+  "Graphics_Imaging",
+  "Media_Ocr",
+  "Security_Cryptography",
+  "Storage_Streams",
   "Win32_Media_Audio_Endpoints",
+  "Win32_Storage_Xps",
+  "Win32_System_Com",
   "Win32_System_Com_StructuredStorage",
   "Win32_System_Variant",
   "Win32_Foundation",
+  "Win32_Graphics_Gdi",
   "Win32_UI_WindowsAndMessaging",
 ] }
 
--- a/src-tauri/src/actions.rs
+++ b/src-tauri/src/actions.rs
@@ -10,6 +10,8 @@ use crate::tray::{change_tray_icon, TrayIconState};
 use crate::utils::{
     self, show_processing_overlay, show_recording_overlay, show_transcribing_overlay,
 };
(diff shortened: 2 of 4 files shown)


cjpais:
@evrenesat would you mind merging this into the same PR as #770? Would rather have the feature come in all at once when I decide to merge it. Closing this PR for now, I'm trying to slim down the number of active PR's so I can have some sanity and manage maintenance.

PR #809, opened 2026-02-13
fix(linux): uppercase letters dropped in Direct paste mode (auto/enigo)

### LINKED ISSUE #429: First character missing when using Direct paste method on GNOME/Wayland
When using Handy on GNOME/Wayland with the "Direct" paste method, the first character of the transcription is consistently missing.

## Environment
- OS: Arch Linux
- Desktop: GNOME on Wayland
- Handy version: v0.6.5
- Transcription model: Parakeet V3

## Steps to reproduce
1. Configure Handy with Paste Method: Direct
2. Use a global shortcut (via GNOME custom keybinding sending `SIGUSR2`) to start/stop recording
3. Speak something
4. The transcribed text is typed, but the first character is missing

## Examples
- Said: "Hello, how are you?" → Typed: "ello, how are you?"
- Said: "The quick brown fox" → Typed: "he quick brown fox"

## What we tried
- Changing the keyboard shortcut (from `Ctrl+Space` to `²` key) → same issue
- Adding a delay before sending `SIGUSR2` → no effect (the issue is after transcription, not before)
- Installing `wtype` → doesn't work on GNOME (Mutter doesn't support the virtual keyboard protocol)
- Installing `dotool` → works on GNOME, but Handy uses `enigo` internally, not dotool

## Analysis
Looking at the logs, Handy uses the `enigo` Rust library via X11/XTest:
```
[enigo::platform][DEBUG] x11 connection established
[enigo::platform][DEBUG] fast text entry not available. Trying to enter individual letters now
```

The first keypress seems to be "eaten", possibly due to timing issues with the key release from the trigger shortcut.

## Possible solutions
1. Add a small configurable delay (e.g., 50-100ms) before typing the first character
2. Add support for `dotool` as an alternative backend (works on GNOME/Wayland via uinput)
3. Investigate why the first character is lost with enigo

## Workaround
Using "Clipboard" paste method works, but requires manually pasting with `Ctrl+Shift+V`.

NateWeav:
On Linux, Direct paste mode drops all uppercase letters when set to auto mode on KDE Wayland. The root cause:

The enigo fallback (used in Auto mode on KDE Wayland when no native tool is available) dynamically remaps keycodes for uppercase letters via x11rb. On XWayland, the keymap change doesn't propagate fast enough, so the key press fires before the remap takes effect.

Fixed the enigo fallback by explicitly using Shift + lowercase letter for ASCII uppercase, which uses keycodes already in the base keymap and avoids dynamic remapping.

## Before Submitting This PR

## Human Written Description

Shortly after installing Handy I noticed that when using direct paste mode on CachyOS (KDE Plasma Wayland), all uppercase letters were being dropped from transcription output. For example, "Hello World" would come out as "ello orld". Figured I'd clone the repo and give opus a go at fixing the issue, and it one-shotted it!

## Related Issues/Discussions

Fixes #429  

## Community Feedback

Fixes direct paste using enigo on Wayland systems. Many users were having the same issue, see #429.

## Testing

- [x] Tested on Plasma 6 (Wayland) with direct paste enabled.
- [ ] Tested on GNOME
- [ ] Tested on X11

## AI Assistance

**If AI was used:**

- Tools used: Claude Code with Opus 4.6
- How extensively: AI identified the bug given the description of the problem and wrote the code to fix it. I hand reviewed the code and tested.

### DIFF
--- a/src-tauri/src/input.rs
+++ b/src-tauri/src/input.rs
@@ -115,9 +115,46 @@ pub fn send_paste_shift_insert(enigo: &mut Enigo) -> Result<(), String> {
 /// Pastes text directly using the enigo text method.
 /// This tries to use system input methods if possible, otherwise simulates keystrokes one by one.
 pub fn paste_text_direct(enigo: &mut Enigo, text: &str) -> Result<(), String> {
-    enigo
-        .text(text)
-        .map_err(|e| format!("Failed to send text directly: {}", e))?;
+    // On Linux, enigo uses x11rb which dynamically remaps keycodes for characters
+    // not in the base keymap (like uppercase letters). On XWayland (KDE Plasma etc.),
+    // the keymap change doesn't propagate fast enough, causing uppercase letters to
+    // be dropped. Work around this by explicitly using Shift + lowercase letter for
+    // ASCII uppercase, which avoids keymap remapping entirely.
+    #[cfg(target_os = "linux")]
+    {
+        for c in text.chars() {
+            if c.is_ascii_uppercase() {
+                // Press Shift, type the lowercase letter, release Shift.
+                // This uses keycodes already in the base keymap, avoiding
+                // the dynamic remap that fails on XWayland.
+                enigo
+                    .key(Key::Shift, enigo::Direction::Press)
+                    .map_err(|e| format!("Failed to press Shift: {}", e))?;
+                std::thread::sleep(std::time::Duration::from_millis(5));
+                enigo
+                    .key(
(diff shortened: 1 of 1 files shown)


cjpais:
I'm sure this works, but it fundamentally changes what direct is.. Which is enigo's .text method. It feels kind of like a hack and it's best we just try to fix the bug in enigo itself. @NateWeav if you want to fork enigo and give that a shot I would love to see an upstream fix! It will help even more people!!!

PR #820, opened 2026-02-14
fix: prevent app lockup when toggle shortcut is pressed rapidly

alexpooley:
The microphone pipeline is purely singleton — when a toggle action (start/stop recording) is already in progress, there is no point queuing additional ones. Rapid presses caused threads to pile up waiting on the state mutex while the first press held it during blocking audio device I/O, freezing the app.

Add an AtomicBool guard that rejects concurrent toggle presses while an action is in-flight.

## Before Submitting This PR

## Human Written Description

Sometimes when I push to talk the UI doesn't appear immediately. I press again because I think I missed the button, etc. Eventually I've pressed a few times and the app has locked up. I have to kill via the Activity Monitor.

## Related Issues/Discussions

No existing issue found — this is a new bug report via PR.

## Testing

Could not fully test the dev build due to macOS TCC refusing accessibility permissions for ad-hoc signed debug binaries. The fix is 6 lines and correct by inspection — the `AtomicBool` CAS guard is a standard pattern for preventing re-entrant calls to a singleton resource.

## AI Assistance

**If AI was used:**

- Tools used: Claude Code
- How extensively: AI investigated the root cause (traced the full shortcut → audio manager mutex chain), identified the deadlock pattern, and wrote the fix. Human identified the bug from real usage and directed the approach.

### DIFF
--- a/src-tauri/src/shortcut/handler.rs
+++ b/src-tauri/src/shortcut/handler.rs
@@ -4,6 +4,7 @@
 //! used by both the Tauri and handy-keys implementations.
 
 use log::warn;
+use std::sync::atomic::{AtomicBool, Ordering};
 use std::sync::Arc;
 use tauri::{AppHandle, Manager};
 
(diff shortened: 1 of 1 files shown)


cjpais:
Closing in favor of #824, thank you for taking the time to submit the PR

Issue #707, opened 2026-02-03
Modifier keys get stuck, requires logout to fix


jpoindexter:
[handy.log](https://github.com/user-attachments/files/25053126/handy.log)

Not an expert here, but I think I found a bug that appeared after last update or so. This just started happening yesterday, no problems until then.

I was working on a Tauri app and my keyboard shortcuts stopped working - had to hold Cmd just to type normally. Spent a while debugging thinking it was a Tauri issue.

When I closed the app I was working on, same thing kept happening. Then I tried changing the modifier keys in Handy settings and the same issue happened while working in VSCode too.

Only way to fix it was to log out and log back in.

Found this with Claude Code while debugging what I thought was a Tauri plugin bug: https://github.com/tauri-apps/plugins-workspace/issues/3245

**Environment:**
- MacBook Pro M4 Pro
- macOS Tahoe 26.3

cjpais:
Thanks for the report and log, looks like you are using CMD + FN is that right? I will try and replicate

jpoindexter:
> Thanks for the report and log, looks like you are using CMD + FN is that right? I will try and replicate

correct

cjpais:
@jpoindexter, could you let me know if this is fixed on 0.7.7 for you? There's a related fix, but I don't know if it will help here or not

jpoindexter:
This bug has not appeared since the latest update seems like it was fixed. Also i should mentioned i have also updated my os to osx 26.4 Beta (25E5207k) as well. I attached my log for your review if needed.

Cheers,
great product!

[handy.log](https://github.com/user-attachments/files/25470954/handy.log)

cjpais:
Ty

Issue #64, opened 2025-08-02
Multiple Instances Allowed on Windows


patrickiel:
On Windows, it's currently possible to launch multiple instances of the application. This can lead to unexpected behavior and should be prevented.

Ideally, attempting to start a second instance should bring the existing one to focus instead.

PR #824, opened 2026-02-14
fix: transcription lock-up race condition & add small debounce

johnpyp:
## Before Submitting This PR

  - There have been a couple attempts at this ([1](https://github.com/cjpais/Handy/pull/820), [2](https://github.com/cjpais/Handy/pull/644)), but this is a more readable/simpler implementation that covers more edge cases, both the signal handler + shortcut entrypoint.

### Part 1 - Race condition fix

This PR started by using an atomic state machine, but it still felt pretty brittle. I realized the correct way to approach this was just by using a "coordinator" (aka actor pattern) which gets simple "notifications" from the rest of the system, and handles the incoming commands sequentially, naturally brokering access to the transcription state.

**Note**: This fix doesn't rely on the 30ms debounce, the debounce is just a UX thing as described below.

Other attempts to fix this only used an AtomicBool, which seems good on the surface, but doesn't truly fix the race condition because it either 1) still causes a lockup in some edge cases (conservative lock) or 2) has multiple pipelines processing at the same time, potentially pasting text interleaved and whatnot. If you only guard the keypress, then you can have multiple push to talk / keypresses accidentally queue up multiple transcriptions.

Basically, there's a myriad of edge case race conditions that get categorically eliminated by using an actor here :)

### Part 2 - Debounce for UX

Second, this PR adds a 30ms debounce which just helps avoid sound effect spam and needless repeated processing (but again, this is NOT necessary for the correctness guarantee)

## Related Issues/Discussions

Fixes https://github.com/cjpais/Handy/issues/64
Fixes https://github.com/cjpais/Handy/issues/707
Fixes https://github.com/cjpais/Handy/issues/462 (closed, but not truly fixed)

Related attempts: 
- https://github.com/cjpais/Handy/pull/820 *probably still be useful, dealing with audio race conditions and recovery
- https://github.com/cjpais/Handy/pull/644

Previous related PR which attempted to mitigate this but caused the lock up accidentally:
- https://github.com/cjpais/Handy/pull/560

## Community Feedback

Issues linked above.

## Testing

Tested on linux + wayland using SIGUSR2.

I have not tested this on macos, or using push to talk keybinds in the app. If someone else is able to test in those environments that'd be great, or we can merge and see how it works out. I don't think this would cause issues.

## AI Assistance

**If AI was used:**

- Tools used: Claude Code
- How extensively: I used Claude Code to explore the issue and verify the race condition, then instructed it on precisely how to fix the issue, and verified the fix. I stand by the fix.

### DIFF
--- a/src-tauri/src/actions.rs
+++ b/src-tauri/src/actions.rs
@@ -435,7 +435,6 @@ impl ShortcutAction for TranscribeAction {
                 utils::hide_recording_overlay(&ah);
                 change_tray_icon(&ah, TrayIconState::Idle);
             }
-
         });
 
         debug!(


johnpyp:
Note: refactored to use an actor to be even more robust to future race conditions and be a lot easier to understand conceptually. The original state machine commit is still here.

cjpais:
from the quick once over this seems like probably the best change, I'll take a further look

edit: I tested this on MacOS and it seems to work great for me

PR #835, opened 2026-02-16
feat: overlay streaming text display with copy/close buttons

divnjl2:
Add streaming text display, copy/close buttons, and done state to recording overlay. See commit message for details.


cjpais:
Please provide a screenshot and fill out the PR form. You need to also provide details here itself. Otherwise your PR will be closed. I don't have time for PR's where the human being behind them is not involved.

PR #837, opened 2026-02-17
support moonshine v2

cjpais:


fikrikarim:
Since Moonshine is a streaming model, I wonder if there's going to be streaming support?

cjpais:
eventually yes, I am thinking about what the API surface should look like and how to handle native streaming models, and potentially what abstraction exists for 'simulated streaming'

fikrikarim:
This is awesome. Looking forward to it. Thank you!

PR #850, opened 2026-02-18
feat: add audio speed control for faster transcription

marlonsena:
## Summary
Add an audio speed control setting that allows users to speed up audio before transcription. This enables users with slower hardware to use larger, more accurate models without sacrificing speed.

## Context
Many users, especially in regions where high-end hardware is expensive, struggle with slow transcription times. Additionally, language learners who speak at a slow pace (like non-native speakers practicing English, like me) often have slower speech patterns. 

This feature speeds up the audio before it reaches the transcription engine, reducing processing time while maintaining reasonable accuracy. 
So user don't need to choose a smaller model, it can choose a better one and the optimization is on input (user recording).

  ## Use cases
  - Users with slower CPUs/GPUs who want to use larger models
  - Language learners who speak at a natural pace
  - Anyone who wants faster transcription
  
  ## What to test before you use it
  - [x] Test with different speed settings (1.0x - 2.0x). 
      - In my use case 1.7x works perfectly like 1x.

  ## Notes
  - Default is 1.0x (no speed change)
  - Higher speeds can reduce accuracy

### DIFF
--- a/src-tauri/src/lib.rs
+++ b/src-tauri/src/lib.rs
@@ -292,6 +292,7 @@ pub fn run(cli_args: CliArgs) {
         shortcut::change_mute_while_recording_setting,
         shortcut::change_append_trailing_space_setting,
         shortcut::change_app_language_setting,
+        shortcut::change_audio_speed_setting,
         shortcut::change_update_checks_setting,
         shortcut::change_keyboard_implementation_setting,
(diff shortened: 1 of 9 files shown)


cjpais:
Sorry I'm not accepting new features at this time.

Issue #439, opened 2025-12-11
[BUG] `Paste Method > Direct` does not use the correct keyboard layout.


d4h0:
## Bug Description

As the title says, `Paste Method > Direct` does not use the correct keyboard layout.

To test Handy, I said "Eins, zwei, drei" ("One, two, three" in German).

The History shows "Eins, zwei, drei" and if I switch `Paste Method` to `Clipboard`, "Eins, zwei, drei" is pasted.

`Paste Method > Direct` results in "Eins, ywei, drei." ("zwei" vs. "ywei").

The characters 'z' and 'y' are switched on the QWERTY (e.g., US) and QWERTZ (e.g., Germany) keyboard layouts, therefore I guess `Paste Method > Direct` does not use the correct keyboard layout.

All users that don't use QWERTY probably will have this issue.

I asked Gemini how to fix the issue, and it gave me the following:

---

This is likely a "Scancode vs. Unicode" issue in the Rust backend. The application appears to be simulating key presses based on physical US key locations (Scancodes) rather than sending the actual characters.

**The Fix:**
In the Rust code handling the "Direct" paste method (likely using the `enigo` crate), switch from simulating individual keys to using the text injection method.

If using `enigo`, change the implementation to use `.text()`:

```rust
// ❌ Likely current implementation (maps to physical keys on US layout):
// enigo.key_click(Key::Layout('z')); 

// ✅ Proposed fix (sends the string as Unicode, respecting layout):
enigo.text(transcribed_text);
```

---

## System Information

**App Version:**

v0.6.6

**Operating System:**

Arch Linux (rolling release, no version)

**CPU:**

Not relevant, I believe.

**GPU:**

Not relevant, I believe.

cjpais:
I think largely this is an enigo issue. We already use `.text()`, that's what the direct is under the hood. Unfortunately I'm not a Linux user, so I can't easily test fixes. Plus this is probably dependent on compositor and DE. We will need enigo upstream to fix this.

d4h0:
There seems to be a related issue in the enigo repo: https://github.com/enigo-rs/enigo/issues/429

pentamassiv:
Enigo has that issue only with the xdo feature. It used to be the default, but now it uses x11rb by default, where that should work. Can you please share some logs so I can check what's going on?

Also, I noticed that you create the Enigo struct each time you want to paste something. I recommend creating it only once, because it does a lot of work creating a connection to the compositor that you keep repeating.

[...]

PR #868, opened 2026-02-20
fix: widened intl keyboard support

CreativeHouseDotOrg:

- ❌ I have searched [existing issues](https://github.com/cjpais/Handy/issues) and [pull requests](https://github.com/cjpais/Handy/pulls) (including closed ones) to ensure this isn't a duplicate

I am aware that this has been discussed multiple times, but I want to share my solution which I found worked in my case.

https://github.com/cjpais/Handy/issues?q=is%3Aissue%20state%3Aopen%20keyboard

https://github.com/cjpais/Handy/issues/439

- ✅ I have read [CONTRIBUTING.md](https://github.com/cjpais/Handy/blob/main/CONTRIBUTING.md)

- ✅ I have explained in the description below why this should be reconsidered
- ❌ I have gathered community feedback (link to discussion below)

## Human Written Description

- I noticed that the keyboard support wasn't accurate. When assigning a key it got assigned to another key. Supposedly because of international keyboard layout shifting

- I think it will matter because it will make the key assignment more accurate even if the key that is being displayed is not exactly the one that is being mapped. At least the button which has been assigned is also the one that works.

## IMPORTANT NOTE:

The effective shortcut key can be correct at runtime even if the UI label appears as a different symbol/key name in some layouts. This information can also be found in the Readme.md.

## Related Issues/Discussions

Fixes # ❌
Discussion: ❌

## Community Feedback

❌

## Testing

I have tested this in: 

**Development using**

    bun run tauri dev

**Production build using**

    bun run tauri build

❌

## AI Assistance

- ❌No AI was used in this PR
- ✅ AI was used (please describe below)

**If AI was used:**

- Tools used: Github Copilot
- How extensively: **Wrote all of the code.**

### DIFF
--- a/README.md
+++ b/README.md
@@ -136,6 +136,10 @@ For reliable text input on Linux, install the appropriate tool for your display 
 
 Without these tools, Handy falls back to enigo which may have limited compatibility, especially on Wayland.
 
+### Keyboard Shortcut Display Notes
+
+The effective shortcut key can be correct at runtime even if the UI label appears as a different symbol/key name in some layouts.
+
 **Other Notes:**
 
 - **Runtime library dependency (`libgtk-layer-shell.so.0`)**:


cjpais:
It looks like you're making a lot of changes to [Handy Keys](https://github.com/handy-computer/handy-keys), and I believe that you may need to submit the changes directly to that repo instead. I would love to fix this in handy, but we largely need to fix this in the upstream repo, so please submit the changes there, so they can be reviewed.

PR #872, opened 2026-02-21
Bump macOS minimum version to 10.15 (Catalina)

brycedrennan:
## Summary
- Bumps `minimumSystemVersion` from 10.13 (High Sierra) to 10.15 (Catalina)

## Motivation

This is a prerequisite for upgrading transcribe-rs, which bumps whisper-rs to 0.15.1. The newer whisper.cpp (v1.7.4) uses `std::filesystem::path` in `ggml-backend-reg.cpp`, which requires macOS 10.15+.

macOS 10.13 (High Sierra) has been EOL since November 2020, and Apple Silicon Macs all ship with 10.15+, so this shouldn't affect any active users.

Related: https://github.com/cjpais/transcribe-rs/pull/37

## Test plan
- [x] Full `cargo tauri build` succeeds on macOS with this change
- [ ] Verify app launches on macOS 10.15+

### DIFF
--- a/src-tauri/tauri.conf.json
+++ b/src-tauri/tauri.conf.json
@@ -51,7 +51,7 @@
     "macOS": {
       "files": {},
       "hardenedRuntime": true,
-      "minimumSystemVersion": "10.13",
+      "minimumSystemVersion": "10.15",
       "signingIdentity": "-",
       "entitlements": "Entitlements.plist"
     },


cjpais:
@brycedrennan you might underestimate how many people use handy (I assume 500k+ based on my website metrics) hahah. This will affect active Intel Mac users (which the number is above 0, and probably below 10,000) as far as I'm aware. 

It will entirely stop them from being able to have new releases. 

I may drop support at some point, but would love some further discussion on this.

Possibly we need to feature flag things off? So we can have both 0.13.2 and 0.15.1 support?

brycedrennan:
Yes good point. I defer to your expertise.  

I was just switching from SuperWhisper, and was trying to figure out why Handy wasn't as fast, and this is where I ended up. 

I've already compiled a new version of Handy that's going faster for me, so I'm good either way.

cjpais:
@brycedrennan makes sense and understandable

Working on figuring out the best way to accelerate everything with good distribution methodology is something I'll be thinking about a lot in the coming months. Hopefully we should be able to speed up inference on all platforms ideally. Especially when using parakeet or the other models too

cjpais:
This will be superseded by #1041 I was able to fix the CI for whisper-rs finally. Whisper is >2x speedup on my Mac as a result. We will be dropping support for 10.13 and migrating to 10.15

We will see if everything still works when we get the build out and tested by people

Issue #133, opened 2025-09-27
Feature request: option to specify recordings folder


gm-333:
Took me a while to find, where recordings are stored.
Please let me specify where they are stored and possibly also delete single ones/clean complete history from within the app. Otherwise i have to do it from folder. I basically always want to delete after i know, that a transcript worked.

Thank you

cjpais:
Delete is coming in the next version. Will not be allowing a specified directory.

Aravinth-Earth:
@cjpais 
As a user, i feel like having an option to set custom path for audio recording codul be usefuel for other meta processing of my audio recordings.
But i am curious, on the reason behind this design decision, Coudl you share some insights on that ?

cjpais:
Mainly I don't have time to make a nice UX for it. I will accept PRs on this @Aravinth-Earth, I just need the UX to be nicely considered and potentially a few design options to pick from

Aravinth-Earth:
Thanks for the quick reply :). The PR, that i can try, will revert soon with the PR :)

cjpais:
Please just think out the UI/UX and provide screenshots. I have limited time to review, so please make it easy for me :)

Aravinth-Earth:
@cjpais please check the PR #874 
Had added screenshots, coudln't added diff varients, let me know of ur thoguths

PR #874, opened 2026-02-21
Add custom recordings directory management and UI integration

Aravinth-Earth:
## Before Submitting This PR

## Human Written Description

As we save voice inputs as recordings, insted of default app data folder, i prefer to save them in a folder where i can easily  moniter/mange, as it's of a PII info my machine collectes / stores

## Related Issues/Discussions
- https://github.com/cjpais/Handy/issues/133#issuecomment-3938329286

Changes gist #

- Implement `set_recordings_directory` command to manage custom recordings path.
- Introduce `RecordingsDirectory` component for user interface to select and manage the recordings directory.
- Update `HistoryManager` to resolve recordings directory from settings dynamically.
- Modify settings structure to include `recordings_custom_dir`.
- Enhance error handling and user feedback for directory operations.

## Discussion & Community Feedback

Nothign Yet

## Testing

Tested belwo scenariso & found no issues
1. Settings → Advanced → History — new "Recordings Folder" row appears
2. Toggle ON → pick a custom folder →  Record something, check the custom folder has the .wav 
4. Toggle ON with move_existing → old recordings migrate, toast shows count 
5. Toggle OFF → confirmation dialog appears, cancel does nothing
6. Toggle OFF → confirm → recordings move back to default, toast
7. History page → Open Recordings Folder → opens custom folder
8. Restart app with custom folder set → still uses custom folder

1. New Setting Option Added
New Setting Option Added

2. When Custom Folder selected - that path shown in UI & recordings auto moved to new path
When Custom Folder selected - that path shown in UI   recordings auto moved to new path

3. When Disabling custom path, a user confirmation is asked
When Disabling custom path, a user confirmation is asked

4. On succesful revert of recordings path, all recording auto moved to default folder
On succesful revert of recordings path, all recording auto moved to default fodler

## AI Assistance

**If AI was used:**

- Tools used: Codex & Claude Code
- How extensively: All the implementaiton, Build Generation, Seconday Testing


cjpais:
Maybe we can move this to the about pane? Similar to how we have the paths for App Data Dir and also the Debug Dir? We can add a new one for models, and then a 'change..' button or something similar which lets a user select a folder?

Aravinth-Earth:
Yes I coudl move this & have all paths to be shown in same way, along with associated action buttons in same place

1. [ Current ] App Data Directory - with `Open` 
2. [ Current ] Log Directory - with `Open` 
3. [ New ] Recordings Directory - with `Change` 
5. [ New ] Models Directory - with `Change` 

For both New Optiosn, how about theses cases ?
    1. Once changed from default having `Change` and `Reset` adds vlaue or just `Change` only always ]
    2. Do we need to have `Open` as well ?

cjpais:
You bring up a good point for models too. I wonder if we can just change the entire app data directory, instead of having so many options for where things can be. Just simplify. 

Maybe we can switch to icons and see how that feels for open and edit? Edit pops open a dialog to choose the location and open just opens as it does now?

Aravinth-Earth:
1. Entire app directry as a single config, feels bit confusing (at least for me), When i see the folder contents [ ..\AppData\Roaming\com.pais.handy ], i see 3 types of contents, hence having separate config points feels right
    1. Internal App Data
        1. 🧠 Configuration files and similar 
        2. the `settings_store.json` comes under this category 
        3. If logical we can add log files here or keep in curent place itself ]
    2. Personal Data
        1. 🙂 Contains recordings, history DB, etc. 
        2. Users may place this wherever they want: cloud sync folder, big volume, encrypted storage, etc. 
        3. I prefer keeping this in a could sync folder
    3. Models & Artifacts Folder
        1. ⚙️ Contains downloaded third-party models and related artifacts 
        2. These files are large and performance sensitive, so separating them allows better space/performance control.
2. On the Icons, yes that is a good change, we could have dynamic 2/3 action icons in same line, with hover text
        1. Open
        2. Edit
        3. Revert  [ Comes only when default changed ]
        image

cjpais:
I don't see the changes, please push them so I can try

Aravinth-Earth:
i was just trying locally, Here you go, this brings 

1. the Recording Directory display to About
2. and introduced Models Directory

That new 3 layers are yet to be implemented, for now just a thoguh to get your view

image

cjpais:
I liked the simpler view better. Also there is not consistency when using icons and not which looks... weird.

I do have some concern that changing directories will cause issues for the app data however... hmm. This will need to be tested

Aravinth-Earth:
i am tryign that 3 config path option, [ in belwo approach ]

- Internal App Data → settings + logs (one path)
- Personal Data → recordings + history.db (one configurable path)  
- Models → models (one configurable path)

this feels good, any thoughts ?

And functioanlly yes, this is introducing change sin 2 additioanl paths, so need deep testing

cjpais:
I think maybe only have models be a separate dir now. It at least logically makes sense because they are quite large.

Recordings could get large too but im not that concerned about it?

Aravinth-Earth:
for me, having recordings & history.db to be configurabel, in a cloud folder separatly is vlaue add, so i can sync & reprocess later

So the below works ?

1. models as it is under 'Models' [ configurable ]
2. recordings + history.db  - either both in same folder or 2 sub folders under Personal Data as in Recordings & DB [ configurable ]
3. settings & logs in current path itslef [ Leave as it is ]

cjpais:
no, I understand the value add but it looks too messy to my eye so I don't want all that configuration yet. only model dir

Aravinth-Earth:
So this 4 paths shown already is fine and no further changes right ?

image

cjpais:
no, remove recordings being able to change

Aravinth-Earth:
1. recordings editablility removal, goes back to defualt app logic only - got it
2. icons for Models dir action, fine ? if yes, Change Open to icon for App Data Dir & logs too ?

cjpais:
1. thanks
2. I think lets do Icons everywhere, changing open to the the folder or whatever. we just need matching ui

PR #509, opened 2026-01-02
feat: Implement OpenAI style local API server for audio transcription

Yorick-Ryu:
## Before Submitting This PR

**Please confirm you have done the following:**

- [X] I have searched [existing issues](https://github.com/cjpais/Handy/issues) and [pull requests](https://github.com/cjpais/Handy/pulls) (including closed ones) to ensure this isn't a duplicate
- [X] I have read [CONTRIBUTING.md](../CONTRIBUTING.md)

**If this is a feature or change that was previously closed/rejected:**

- [ ] I have explained in the description below why this should be reconsidered
- [ ] I have gathered community feedback (link to discussion below)

## Human Written Description

I implemented a local STT API that follows the OpenAI Whisper format. Currently, the Whisper model is only accessible within Handy; however, many users want to leverage this functionality for external tasks like subtitle transcription without loading multiple model instances. This change exposes the speech-to-text capability as a standardized service, allowing users to do more with limited system memory.

## Related Issues/Discussions

Fixes # None
Discussion: https://github.com/cjpais/Handy/discussions/241

## Community Feedback

https://github.com/cjpais/Handy/discussions/241

## Testing

Environment:

- Tested on: macOS 26.2 (Apple Silicon M1 Pro)
- Status: Functional on macOS. Need help testing on Windows and Linux platforms to ensure consistent behavior.

Test Cases:

- Features: Tested by calling the API using `curl` and [Demo: convert MP3 to SRT](https://qm.rick216.cn/?lang=en&localenabled=true&localport=5500)
- On-demand Loading: Verified via curl that calling the `/v1/audio/transcriptions` endpoint correctly triggers the model loading process in the background.
- Waiting Mechanism: Confirmed the API response waits until the model is fully loaded before processing the transcription, preventing "Model not loaded" errors.
- Verified Limitations: Tested various audio formats and confirmed only MP3 currently works reliably; documented this behavior and added a "welcome PRs" note in `LOCAL_API.md` to guide future contributors.

## Screenshots/Videos (if applicable)

image2

image1

PR #886, opened 2026-02-24
feat: add OpenAI-compatible local server transcription

in0vik:
**Regarding the Feature Freeze:** I'm aware the project is in a feature freeze. I believe this PR is worth considering despite that because it's entirely opt-in (zero behaviour change for existing users), adds no new crate dependencies, and directly addresses a recurring ask from the community (see Discussions below). Happy to wait for the freeze to lift if that's preferred.

## Human Written Description

Big fan, tried dozens of alternatives, always come back to this one.
I run faster-whisper on a separate GPU box, so I added support for any OpenAI-compatible /audio/transcriptions endpoint as a transcription backend. Works with local servers (faster-whisper, whisper.cpp, LocalAI, fully private) and cloud providers like Groq/OpenAI as a bonus. Built it with Claude since I don't really know Rust 😅
Kept it minimal: no new crate deps, fully opt-in, single collapsible card in settings. Thanks for the great project!

## Note on PR #509

PR #509 exposes Handy as an HTTP server so other apps can call it. This PR makes Handy a client that calls your existing Whisper server. Complementary directions, not duplicates.

## Related Issues/Discussions

- [Discussion #374](https://github.com/cjpais/Handy/discussions/374) — users asking for whisper.cpp support; this PR enables connecting to a whisper.cpp server
- [Discussion #241](https://github.com/cjpais/Handy/discussions/241) — community interest in API-level integration

## Community Feedback

Discussions #374 and #241 show community interest in connecting Handy to external Whisper infrastructure. This PR is the client-side answer to that.

## What's included

**Backend (Rust):**
- `EngineType::Cloud` + `LoadedEngine::Cloud` — plugs into existing engine dispatch with no pipeline changes
- WAV encoding of recorded audio (16 kHz mono, 16-bit) for API upload
- `call_cloud_api()` — multipart POST to `/audio/transcriptions`, OpenAI format
- 3-attempt retry with delays (300 ms, 800 ms)
- Credentials re-read on every call (no stale config after settings change)
- `save_pending_transcription()` + `retranscribe_history_entry` — if all retries fail, audio is saved and user can retry from History
- Extra params textarea — arbitrary JSON fields forwarded to the endpoint
- `block_in_place` to safely bridge sync ↔ async runtimes

**Frontend (React/TypeScript):**
- `CloudTranscriptionCard` — collapsible accordion in Models settings
- Live Test button (checks endpoint connectivity, shows ✓/✗)
- API key is **optional** — works with auth-free local servers (whisper.cpp)
- All strings use i18n (`en/translation.json`)

**History:**
- `cloud_pending` DB column + migration - failed transcriptions show a Retry button

## Testing

Tested on macOS 15 (Apple Silicon M1):
- **faster-whisper-server** on localhost - end-to-end ✓
- **Groq Whisper API** - ✓, ~300 ms round-trip
- **OpenAI Whisper API** - ✓
- Server down → audio saved as pending → Retry from History ✓
- Settings changed mid-session → new credentials used immediately ✓
- 36 existing Rust tests pass, ESLint clean, `cargo clippy` no new warnings

image

## AI Assistance

- [x] Claude was used in this PR

**If AI was used:**

- Tools used: Claude Code
- How extensively: Pair-programming-— I drove the design and testing; Claude helped with implementation, code review, and writing this PR description. All code was reviewed and tested by me.

### DIFF
--- a/src/bindings.ts
+++ b/src/bindings.ts
@@ -339,6 +339,46 @@ async changeShowTrayIconSetting(enabled: boolean) : Promise<Result<null, string>
     else return { status: "error", error: e  as any };
 }
 },
+async changeCloudTranscriptionBaseUrl(baseUrl: string) : Promise<Result<null, string>> {
+    try {
+    return { status: "ok", data: await TAURI_INVOKE("change_cloud_transcription_base_url", { baseUrl }) };
+} catch (e) {
+    if(e instanceof Error) throw e;
+    else return { status: "error", error: e  as any };
+}
+},
+async changeCloudTranscriptionApiKey(apiKey: string) : Promise<Result<null, string>> {
+    try {
+    return { status: "ok", data: await TAURI_INVOKE("change_cloud_transcription_api_key", { apiKey }) };
+} catch (e) {
+    if(e instanceof Error) throw e;
+    else return { status: "error", error: e  as any };
+}
+},
+async changeCloudTranscriptionModel(model: string) : Promise<Result<null, string>> {
+    try {
+    return { status: "ok", data: await TAURI_INVOKE("change_cloud_transcription_model", { model }) };
+} catch (e) {
+    if(e instanceof Error) throw e;
+    else return { status: "error", error: e  as any };
+}
+},
+async testCloudTranscriptionConnection() : Promise<Result<null, string>> {
+    try {
+    return { status: "ok", data: await TAURI_INVOKE("test_cloud_transcription_connection") };
+} catch (e) {
+    if(e instanceof Error) throw e;
+    else return { status: "error", error: e  as any };
+}
+},
+async changeCloudTranscriptionExtraParams(params: string) : Promise<Result<null, string>> {
+    try {
+    return { status: "ok", data: await TAURI_INVOKE("change_cloud_transcription_extra_params", { params }) };
+} catch (e) {
+    if(e instanceof Error) throw e;
+    else return { status: "error", error: e  as any };
+}
+},
 /**
  * Start key recording mode
  */
@@ -712,6 +752,14 @@ async updateRecordingRetentionPeriod(period: string) : Promise<Result<null, stri
     else return { status: "error", error: e  as any };
 }
 },
+async retranscribeHistoryEntry(id: number) : Promise<Result<null, string>> {
+    try {
+    return { status: "ok", data: await TAURI_INVOKE("retranscribe_history_entry", { id }) };
+} catch (e) {
+    if(e instanceof Error) throw e;
+    else return { status: "error", error: e  as any };
+}
+},
 /**
  * Checks if the Mac is a laptop by detecting battery presence
  * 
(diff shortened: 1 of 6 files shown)


cjpais:
We will support being a server, but not using remote (or other local) providers. 

I believe local transcription is possible across most computers in the past 10 years, and I aim to support that as fully as possible. Even fairly low power machines (read rpi or worse) can use local transcription engines at faster than realtime speed with the right optimizations. That is part of the goal of the project and there will be no half measures to use remote. This is a local only transcription app and it will stay that way. For model support and speed, we should improve that ourselves and make it better. Those are the constraints the app is working with. I would love help from anyone who wants to make local AI inference better

I've written on this extensively in the past in other issues and many other closed PRs. To reiterate, this will not come in. We will focus on making local AI inference better and be even better than remote.

I will say this UI is the best reason for it yet however.

in0vik:
> We will support being a server, but not using remote (or other local) providers.
> 
> I believe local transcription is possible across most computers in the past 10 years, and I aim to support that as fully as possible. Even fairly low power machines (read rpi or worse) can use local transcription engines at faster than realtime speed with the right optimizations. That is part of the goal of the project and there will be no half measures to use remote. This is a local only transcription app and it will stay that way. For model support and speed, we should improve that ourselves and make it better. Those are the constraints the app is working with. I would love help from anyone who wants to make local AI inference better
> 
> I've written on this extensively in the past in other issues and many other closed PRs. To reiterate, this will not come in. We will focus on making local AI inference better and be even better than remote.
> 
> I will say this UI is the best reason for it yet however.

Hey, I really appreciate the work you're doing here and I fully respect the local-first philosophy - it's a strong and  resonates.  That said, I'd love to push back a little on one specific point, because I think there's a logical tension worth discussing.

The app already supports remote LLM calls for text post-processing. From a privacy standpoint, that's arguably *more* sensitive than remote transcription - you're sending the actual semantic content of someone's conversation to an external server (which most transcription providers don't retain anyway). So the privacy boundary has already been crossed, just at a different layer.

Example
My use case is multilingual transcription ru/en. Parakeet is great for English, but for other languages the quality drops noticeably. Keeping Whisper v3 large in memory to handle multilingual workloads costs 1–2GB RAM on a machine that might be doing other things. For users in that situation, a fast cloud transcription API would actually be the *more* pragmatic local-first choice - freeing up local resources for everything else.

I'm not arguing against local inference as the default or even the primary path - I think it should stay that way. But given that remote LLM is already an option in the app, I'd genuinely love to understand what makes remote transcription categorically different in your view.

Happy to be wrong here - just trying to understand the reasoning

cjpais:
> The app already supports remote LLM calls for text post-processing. From a privacy standpoint, that's arguably more sensitive than remote transcription - you're sending the actual semantic content of someone's conversation to an external server (which most transcription providers don't retain anyway). So the privacy boundary has already been crossed, just at a different layer.


Basically it's vibes. And more or less this is a project built on my opinions. And my opinion is that local transcription is good enough, and local LLM is not good enough. 

It is an alpha feature for a reason. Using an LLM to post process transcription enables a much broader range of possibilities than transcription does fundamentally. Not to mention people who fork the application can take advantage of this without having to do extra work. The class of problems that LLMs and STT models solve is very, very different. Transcription is largely a solved problem locally and LLM is not. I don't like it either, but it is overwhelmingly useful enough that it's important to have I feel. Maybe it's a necessary evil for the time being.

I think the percentage of people who can reasonably use a local LLM on their machine is significantly smaller than those who can use a local STT model. Until there are good enough small models to handle a wide variety of post-processing use cases, I will not be shipping local models and likely the post-processing feature will stay in alpha until then. I would love for post processing to become the default path.

At some point I may want to fine tune a small model, but this means we also have to get a much better understanding of what use cases people are using post processing for. 

I mean I think we can fine-tune a model today probably at a fairly small size and ship it as well. And at some point, likely that will be shipped, but its scope in terms of its prompt acceptance is probably much smaller, and maybe it's just tuned for a few prompts. If anyone wants to help out with this problem, certainly I'd love some help on it. For the most part, it's probably just a synthetic dataset generation task + fine tuning a model around 1B in size or smaller. I would potentially ship that. Especially if the model could be natively trained/finetuned/inferenced in 4bit or 8bit precision. 

> My use case is multilingual transcription ru/en. Parakeet is great for English, but for other languages the quality drops noticeably. Keeping Whisper v3 large in memory to handle multilingual workloads costs 1–2GB RAM on a machine that might be doing other things. For users in that situation, a fast cloud transcription API would actually be the more pragmatic local-first choice - freeing up local resources for everything else.

You can have the program offload the memory for the model when it's not being used, and the cost to spin it up is fairly low because it's done in parallel with when you start recording, so unless you're doing a lot of sentences which are just a few seconds long, you should never feel the latency penalty of loading the model. The model on my computer is never kept in memory. I may eventually make this the default. I just don't really know all of other users experiences, but I have no issues loading the model into memory from disk every single time I do a transcription.

And also to say I do appreciate you trying to contribute and wanting to contribute. And if you want to do the most to contribute today, it would be to help try to solve some of the stability issues so we can start focusing on the bigger picture stuff. If you want to start taking on some of these bigger picture tasks of potentially fine-tuning a model or generating a data set for fine-tuning a model, that's also great. And heck, I mean figuring out community places to have these discussions be more open rather than just in some random GitHub PR. This is all stuff that I don't know anything about and I really need help with. This project is far bigger than me and I'm more or less maintaining it alone. I need help. And that's largely why there is a feature freeze. I fundamentally cannot accept new features right now without solving core issues first. I have a lot of ideas from all of the feedback of where to take Handy next, but I think it's most important to solve the core issues first.

in0vik:
Have you considered opening a Discussion specifically looking for people who want to take on the ~1B fine-tuning? Seems like there's enough community interest and it's concrete enough that someone might actually pick it up.

Also worth adding a recruitment note in your README - a lot of potential contributors never make it to Discussions.

cjpais:
@in0vik if you could help me with a PR or opening a discussion for me it would be massive. I can't single handedly steward every part

PR #913, opened 2026-02-28
feat: add GigaAM v3 for Russian speech recognition

pantafive:
Adds GigaAM v3 e2e_ctc engine — Russian speech recognition with punctuation, Latin characters and digits. Uses int8 quantized ONNX model (225 MB), BPE tokenizer with 257 subword tokens.

The model is currently downloaded from HuggingFace (istupakov/gigaam-v3-onnx). It needs to be mirrored to blob.handy.computer to be consistent with other models.

### DIFF
--- a/src-tauri/src/managers/model.rs
+++ b/src-tauri/src/managers/model.rs
@@ -407,7 +407,7 @@ impl ModelManager {
                 id: "gigaam-v3-e2e-ctc".to_string(),
                 name: "GigaAM v3".to_string(),
                 description: "Russian speech recognition. Fast and accurate.".to_string(),
-                filename: "gigaam-v3-e2e-ctc-int8.onnx".to_string(),
+                filename: "giga-am-v3.int8.onnx".to_string(),
                 url: Some("https://blob.handy.computer/giga-am-v3.int8.onnx".to_string()),
                 size_mb: 225,
                 is_downloaded: false,


cjpais:
Can you add gigaam to transcribe rs first and then I will pull that in

pantafive:
Added GigaAM as a proper engine in transcribe-rs: cjpais/transcribe-rs#45

Once that's merged and published, I'll update this PR to use the crate feature instead of the standalone module.

cjpais:
Thank you so much @pantafive, amazing to see how much support we have for this in just one day! I've released `transcribe-rs` 0.2.7 with support :) I also have uploaded to handy blob site at https://blob.handy.computer/giga-am-v3.int8.onnx

Only comment I have is maybe in the description section just making sure it's very clear it's for Russian. Also if you have opinion on testing, would be helpful to have your opinion there too. Is it the best for Russian speech you've tested?

pantafive:
Thanks for the release and the CDN upload!

Updated the PR — now uses `transcribe-rs` 0.2.7 with the `gigaam` feature, the standalone module is removed. Model URL points to `blob.handy.computer`. Tested locally, everything works.

Regarding the description — could you clarify which description you'd like updated? The model description in the app already says "Russian speech recognition", but happy to adjust wherever you think it needs to be clearer.

As for testing — GigaAM v3 is the best Russian speech model I've tested. It outperforms Whisper-large-v3 on Russian benchmarks (9.2% vs 25.1% avg WER) and handles punctuation natively.

cjpais:
Thank you! Mainly I was thinking something a bit stronger for the description, like "Best model for Russian speakers or similar"

pantafive:
I think "best" might be risky in a description — it's subjective, and things move fast in this space, so it could become misleading quickly. "Russian speech recognition. Fast and accurate." states what it does without overpromising. But it's your project — happy to go with whatever you think works best!

eboyko:
"Best model for Russian speakers. Great for bilingual Russian/English use — especially developers mixing both languages."

cjpais:
I'm good with whatever and will defer to you since I don't speak Russian haha. Things do sure move fast

gtubolcev:
Not a streaming model or am I wrong?
Great work anyway, thank you!
It would be great to have some real streaming model. Like the last vosk 0.54 https://huggingface.co/alphacep/vosk-model-ru

PR #930, opened 2026-03-01
Add transcription hook

AlexanderYastrebov:
## Before Submitting This PR

**Please confirm you have done the following:**

- [x] I have searched [existing issues](https://github.com/cjpais/Handy/issues) and [pull requests](https://github.com/cjpais/Handy/pulls) (including closed ones) to ensure this isn't a duplicate
- [x] I have read [CONTRIBUTING.md](https://github.com/cjpais/Handy/blob/main/CONTRIBUTING.md)

**If this is a feature or change that was previously closed/rejected:**

- [x] I have explained in the description below why this should be reconsidered
- [x] I have gathered community feedback (link to discussion below)

## Human Written Description

Add support for transcription hook - an executable script in app's data directory.

If `hooks/transcription` file exists, Handy runs it passing transcription text via stdin and
uses script stdout as a transcription result.

This approach is a flexible extension point for advanced users
(which nowadays means with access to coding LLM) akin to git hooks.

Here are some possible scenarios:
* simple transcription modifications
* a pipeline involving LLM processing, language detection and translation
* custom paste method (as Handy does nothing if transcription is empty)
* conditional processing based on the active application waiting for the input
* quote clipboard content and append transcription

## Related Issues/Discussions

See related:
* https://github.com/cjpais/Handy/discussions/168
* https://github.com/cjpais/Handy/discussions/162
* https://github.com/cjpais/Handy/discussions/916
* https://github.com/cjpais/Handy/discussions/911
* https://github.com/cjpais/Handy/discussions/834
* https://github.com/cjpais/Handy/discussions/847
* https://github.com/cjpais/Handy/discussions/833
* https://github.com/cjpais/Handy/discussions/662
* https://github.com/cjpais/Handy/discussions/601
* https://github.com/cjpais/Handy/discussions/335
* https://github.com/cjpais/Handy/pull/739
* https://github.com/cjpais/Handy/pull/638
* https://github.com/cjpais/Handy/pull/455
* https://github.com/cjpais/Handy/issues/157

## Testing

Build and ran Handy with both example scripts.

## AI Assistance

- [ ] No AI was used in this PR
- [x] AI was used (please describe below)

**If AI was used:**

- Tools used: Copilot with Claude Sonnet 4.5
- How extensively: Handle Command stdin in Rust and compose hook example that uses Apple Script.

PR #943, opened 2026-03-02
Remove filler words that conflict with European languages

### LINKED ISSUE #941: [BUG] Automatic filler word removal includes a legitimate word in portuguese
## Bug Description

It seems that the automatic filler word removal is removing the portuguese word "um", which is a legitimate word and often used, because it's both an indefinite article and the number one in the text form. Please see [this discussion](https://github.com/cjpais/Handy/discussions/940) for more details.

## System Information

**App Version:**

0.7.9

**Operating System:**

Windows 11

**CPU:**

AMD Ryzen 7 250

**GPU:**

AMD RAdeon 780M

## Logs

AlexanderYastrebov:
## Before Submitting This PR

## Human Written Description

Removed three filler words from the transcription filter that are actual words in European languages:

- 'um' - Portuguese/Spanish indefinite article meaning 'a/an' (masculine) Example: Portuguese 'Isto é um teste' (This is a test) Removing this was breaking Portuguese transcriptions

- 'ha' - Spanish/Italian/Norwegian/Swedish auxiliary verb meaning 'has/have' Example: Spanish 'Él ha comido' (He has eaten) This is a very common verb form in Romance and Scandinavian languages

- 'eh' - Italian interjection and Canadian English discourse marker While less critical, this can appear in legitimate Italian speech

The remaining filler words are primarily English
vocalized hesitations that don't conflict with common words in other European languages.

Updated tests to use 'uhm' instead of 'um' where needed.

## Related Issues/Discussions

Fixes #941
Discussion: https://github.com/cjpais/Handy/discussions/940

## Testing

Tested using google translate with parakeet model as described in https://github.com/cjpais/Handy/discussions/940 for `um` and `ha`.

## AI Assistance

**If AI was used:**

- Tools used: Copilot+Claude Sonnet 4.5
- How extensively: asked about filler words:
> Please summarize which of these filler words exist in European languages and therefore are not safe to remove.

### DIFF
--- a/src-tauri/src/audio_toolkit/text.rs
+++ b/src-tauri/src/audio_toolkit/text.rs
@@ -196,8 +196,7 @@ fn extract_punctuation(word: &str) -> (&str, &str) {
 
 /// Filler words to remove from transcriptions
 const FILLER_WORDS: &[&str] = &[
-    "uh", "um", "uhm", "umm", "uhh", "uhhh", "ah", "eh", "hmm", "hm", "mmm", "mm", "mh", "ha",
-    "ehh",
+    "uh", "uhm", "umm", "uhh", "uhhh", "ah", "hmm", "hm", "mmm", "mm", "mh", "ehh",
 ];
 
 static MULTI_SPACE_PATTERN: Lazy<Regex> = Lazy::new(|| Regex::new(r"\s{2,}").unwrap());
@@ -327,21 +326,21 @@ mod tests {
 
     #[test]
     fn test_filter_filler_words() {
-        let text = "So um I was thinking uh about this";
+        let text = "So uhm I was thinking uh about this";
         let result = filter_transcription_output(text);
         assert_eq!(result, "So I was thinking about this");
     }
 
     #[test]
     fn test_filter_filler_words_case_insensitive() {
-        let text = "UM this is UH a test";
+        let text = "UHM this is UH a test";
         let result = filter_transcription_output(text);
         assert_eq!(result, "this is a test");
     }
 
     #[test]
     fn test_filter_filler_words_with_punctuation() {
-        let text = "Well, um, I think, uh. that's right";
+        let text = "Well, uhm, I think, uh. that's right";
         let result = filter_transcription_output(text);
         assert_eq!(result, "Well, I think, that's right");
     }
@@ -362,7 +361,7 @@ mod tests {
 
     #[test]
     fn test_filter_combined() {
-        let text = "  Um, so I was, uh, thinking about this  ";
+        let text = "  Uhm, so I was, uh, thinking about this  ";
         let result = filter_transcription_output(text);
         assert_eq!(result, "so I was, thinking about this");
     }


cjpais:
I don't think this is a good solution. We probably should at minimum ship this as a setting and allow people to change the list as this will significantly impact English users

AlexanderYastrebov:
BTW `um` is also used in German, see https://de.wikipedia.org/wiki/Um-zu-Satz

IMO crude removal of words without language context is wrong.

cjpais:
I agree, but what I am saying is this should be a customizable list. Or to have language detection generally done and pass that along to a filter list to apply different ones for different languages. At least this should be possible with Whisper

I don't think it's a fix to create a regression for other users. We can only improve both, not one step forward one step back.

AlexanderYastrebov:
It does not look like #589 provided any statistics why this specific set of words was removed. I think  the problem magnitude also depends on the model used and might not be an issue at all on the powerful models like parakeet v3 (note that I am not arguing to rip off stutter clean up).
Maybe this can be accepted optimistically and reverted in case english speakers report a regression.

Speaking of new features, please consider adding #930 - its a very powerful feature that likely requires less code than customizable cleanup list which is trivially implemented like:
```python
#!/usr/bin/env python3
import sys
import re

text = sys.stdin.read()

FILLER_WORDS = (
    "uh", "um", "uhm", "umm", "uhh", "uhhh", "ah", "eh", "hmm", "hm", "mmm", "mm", "mh", "ha",
    "ehh",
)

# Remove filler words (case-insensitive)
for word in FILLER_WORDS:
    # Match filler word with word boundaries, optionally followed by comma or period
    pattern = re.compile(r'\b' + re.escape(word) + r'\b[,.]?', re.IGNORECASE)
    text = pattern.sub('', text)

# Clean up multiple spaces to single space
text = re.sub(r'\s{2,}', ' ', text)

sys.stdout.write(text)

```

cjpais:
@AlexanderYastrebov new features are not really being accepted at the moment.

What you are suggesting is a power user feature. Not something for the average person. Handy targets the most average grandma as the primary user. 

As an English speaker I know there will be regressions based on the usage of Parakeet I have and it's not going to be pulled in. I've given options for you for the way forward for this PR. Regressions are unacceptable. We can discuss ways forward but we cannot discuss what is effectively a hack. We must think about the proper way to fix. I will not entertain hacks or regressions.

AlexanderYastrebov:
> What you are suggesting is a power user feature. Not something for the average person. Handy targets the most average grandma as the primary user.

I'd argue its not the case anymore - with everyone using ChatGPT it is not a high bar to compose one-off script to transform stdin to stdout. Ship a stub hook like
```python
#!/usr/bin/env python3
"""
This script reads text from standard input (stdin), allows for 
text transformations, and writes the result to standard output (stdout).

Ask ChatGPT or Claude to add code here to change transcription like you wish.
"""

import sys

text = sys.stdin.read()

sys.stdout.write(text)
```
and you'll be surprised how creative people are.

cjpais:
@AlexanderYastrebov grandmas are not using ChatGPT for an application like this. People are creative but you're still talking about 1%. I have a lot of normal people as friends that would never do this. This is not up for discussion. Your feature will be considered, but like everyone else's. New features are not accepted at this time. And if you really want a feature you will listen to my feedback. Arguing with me and the little time I have to maintain this application is not a good strategy. If you really want this in, help me work on a Beta Channel so we can push experimental features out to people without creating a massive set of issues in mainline.

It is also a departure from the discussion on this PR, which I've given potential ways forward

rguerreiro:
I'm ok with the filler words being a customizable list that the user can change as he\she wishes. Either way it really needs to be implemented because, as Handy stands right now, it's unusable to me in portuguese.

AlexanderYastrebov:
Okay, closing this for now.

Issue #436, opened 2025-12-11
[BUG] Handy crashing Windows 11


duskmosss-creator:
## Before You Submit

**Please search [existing issues](https://github.com/cjpais/Handy/issues) to avoid duplicates.** Your bug may already be reported! Right now it's just me maintaining this project so many issues can be overwhelming! Help me out by checking first.

## Bug Description

anytime i try and activate handy with the model loaded and drivers up to date on windows 11 pro and a good laptop It crashes i tried a different shortcut than ctrl+space still crashed. a different model and changing every setting did nothing to prevent crashing. I tried unistalling then reinstalling from github instead of the website still crashed. 

## System Information

**App Version:**

Handy_0.6.5_x64-setup.exe

**Operating System:**

windows 11 pro 

**CPU:**
CPU

	AMD Ryzen AI 9 HX 370 w/ Radeon 890M 

**GPU:** 

nvidia 4060 laptop with studio driver 

^^ could it be i am limiting my clock speed too low and need to give it higher speed let me check; no it still crashed

cjpais:
did you try whisper or parakeet? I need more specific details of what particular models you are trying to load.

lcorsog:
For me, it crashes when I try to change the model from Parakeet to Whisper after the download. I’m also having the same issue when trying to activate it using the hotkey: the pop-up opens, but the application closes and nothing happens.

cjpais:
@lcorsog there are many known issue with whisper, it can be quite unstable on certain sys configurations. I'm working to provide another option which should be more stable.

lcorsog:
I think the crashes were happening because I was using the x64 version on a Windows on ARM machine. I’ve now downloaded version 0.6.7 for ARM64, but I’m having trouble opening the application. The error I’m getting is 0xc000007b.

cjpais:
well, whisper for arm definitely is broken hahah and is a known issue also

duskmosss-creator:
sorry about that i tried both models it seems to be an issue that is of the activation so anytime that i do the transcription comand it quits the program i did an update to the most recent version today 12/12/25

duskmosss-creator:
i have tried whisper small and parakeet v3

duskmosss-creator:
i can give you a screen recording of what happens but it just closes the program with no dialog or warnings and there are no more instances in task manager whenever i try to press the command to transcribe

cjpais:
[...]

PR #945, opened 2026-03-02
Handle microphone init failure without aborting

ferologics:
## Summary
This PR prevents a hard crash when microphone stream initialization fails (for example, when Windows microphone privacy blocks access).

## What changed
- `src-tauri/src/audio_toolkit/audio/recorder.rs`
  - Replaced panic paths in `AudioRecorder::open` (`expect`/`unwrap`/`panic`) with explicit error propagation.
  - Added a startup handshake channel so `open()` only succeeds after worker initialization succeeds.
  - Converted known permission-denied errors into a clearer user/actionable message.
- `src-tauri/src/actions.rs`
  - When recording fails to start, immediately restore UI state by hiding the recording overlay and setting tray icon back to idle.

## Why
In release builds (`panic = "abort"`), panic paths during recorder startup can terminate the process. This keeps startup failure as a normal error path instead.

Refs #436

### DIFF
--- a/src/i18n/locales/ar/translation.json
+++ b/src/i18n/locales/ar/translation.json
@@ -540,7 +540,8 @@
     "dismiss": "تجاهل"
   },
   "errors": {
-    "loadDirectory": "خطأ في تحميل المجلد: {{error}}"
+    "loadDirectory": "خطأ في تحميل المجلد: {{error}}",
+    "recordingFailed": "Failed to start recording: {{error}}"
   },
   "appLanguage": {
     "title": "لغة التطبيق",
--- a/src/i18n/locales/cs/translation.json
+++ b/src/i18n/locales/cs/translation.json
@@ -540,7 +540,8 @@
     "dismiss": "Zavřít"
   },
   "errors": {
-    "loadDirectory": "Chyba při načítání adresáře: {{error}}"
+    "loadDirectory": "Chyba při načítání adresáře: {{error}}",
+    "recordingFailed": "Failed to start recording: {{error}}"
   },
   "appLanguage": {
     "title": "Jazyk aplikace",
--- a/src/i18n/locales/de/translation.json
+++ b/src/i18n/locales/de/translation.json
@@ -540,7 +540,8 @@
     "dismiss": "Schließen"
   },
   "errors": {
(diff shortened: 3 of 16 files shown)


cjpais:
@ferologics have you tested this yourself to verify?

Just want to make sure. Also do you know if the UI visibly says anything about this failure case?

ferologics:
@cjpais i wasn't able to make handy run locally from source ([ref](https://discord.com/channels/1391636091853078608/1391636092624568382/1478168674627092600)) so i couldn't test. 

I came across the crash due to having disabled microphone globally via settings.

Tried to make this change isolated to that, not touching any UI. Onboarding / in-app hints about disabled mic could made be better tho.

Right now when the crash happens it briefly shows the recording UI and then it disappears, no info in the app / anywhere about it.

ferologics:
re: the microphone access denied bit -- it comes directly from my local agent. i had it check the crash log, look at current system settings configuration and patch that particular case in handy.

cjpais:
Got it, I am going to push some changes for the UI I think. Let's get your build env fixed

ferologics:
Tested locally now.

Repro:
- launched Handy from source on Windows
- disabled microphone access in Settings -> Privacy & security -> Microphone
- pressed the transcribe hotkey repeatedly

Result:
- Handy no longer crashes
- the app stays alive/running
- repeated hotkey presses are handled safely
- logs show the normalized microphone access denied error path firing as expected

So the crash-path fix is working on my machine now.

Current UX is much better than before: the recording UI briefly appears and then resets instead of aborting the app, and the PR's UI follow-up now propagates the error to the frontend.

I think there is still room for polish in a follow-up (for example a more explicit microphone-blocked message/icon), but the main bug here appears fixed.

ferologics:
image

ferologics:
I would have expected the error to surface closer to the recording indicator / overlay, so I initially missed it. That said, I think this is fine for this PR: the crash is fixed, the error is handled, and there is now a visible path for surfacing it. A more contextual microphone-blocked warning could be a separate follow-up if we want to polish it later.

cjpais:
Thank you for this!

I suspect it may also fix: https://github.com/cjpais/Handy/issues/475

PR #952, opened 2026-03-03
Feature/headless transcribe cli (for use in OpenClaw)

mane81:
## Before Submitting This PR

## Human Written Description

I built this for my personal local automation setup with OpenClaw.  
My main use case is receiving voice messages (Telegram/WhatsApp) and having OpenClaw transcribe them reliably via a deterministic CLI call, instead of UI/hotkey interaction.  
This also helps avoid running multiple separate STT stacks in parallel and keeps my workflow simpler while still using Handy’s local model selection.

## Related Issues/Discussions

Fixes #N/A  
Discussion: N/A (happy to open one if required before merge)

## Community Feedback

No dedicated discussion thread yet.  
If preferred, I can open one first and collect feedback there before merge.

## Testing

Tested locally on macOS (Apple Silicon):

- `cargo check -q` passes
- `cargo run -- --transcribe-file /path/to/audio.wav --format text` returns transcript to stdout
- `cargo run -- --transcribe-file /path/to/audio.wav --format json` returns JSON output
- Verified default behavior uses selected model from `settings_store.json`
- Verified `--model-id` override works

N/A (CLI-only change)

## AI Assistance

**If AI was used:**
- Tools used: OpenClaw assistant (**Codex 5.3**)
- How extensively: Medium — AI helped with implementation/refactoring and wording; behavior and tests were validated by me.

### DIFF
--- a/src-tauri/src/cli.rs
+++ b/src-tauri/src/cli.rs
@@ -15,6 +15,10 @@ pub struct CliArgs {
     #[arg(long, default_value = "text", value_parser = ["text", "json"])]
     pub format: String,
 
+    /// Optional model id override (defaults to selected model in Handy settings)
+    #[arg(long)]
+    pub model_id: Option<String>,
+
     /// Disable the system tray icon
     #[arg(long)]
     pub no_tray: bool,
--- a/src-tauri/src/lib.rs
+++ b/src-tauri/src/lib.rs
@@ -45,6 +45,7 @@ use tauri_plugin_log::{Builder as LogBuilder, RotationStrategy, Target, TargetKi
 use crate::settings::get_settings;
 use anyhow::{anyhow, Context, Result};
 use transcribe_rs::engines::parakeet::{ParakeetEngine, ParakeetInferenceParams, ParakeetModelParams};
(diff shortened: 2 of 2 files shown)


AlexanderYastrebov:
Check out https://k2-fsa.github.io/sherpa/onnx/index.html which is more suitable for automation and has a lot of tools and features.
See examples from https://k2-fsa.github.io/sherpa/onnx/pretrained_models/offline-transducer/nemo-transducer-models.html?highlight=parakeet specifically.

cjpais:
Yeah, largely this kind of belongs in its own program rather than handy. It's something we might support at some point, but right now it's really not a super high priority feature. And to be honest, a claw could probably vibe code itself one of these. And quite frankly, there's probably at least a hundred open source CLI tools that do this.

Like yeah, I also will probably be writing one soon, but it's more a test binary for `transcribe-rs` than it is meant to be something super production grade. So for now I'm gonna close this for handy because I don't think we are going to support it in the super near term. Though I'm not opposed to it, largely speaking, it's just that we really need to focus on stability for the moment. And I really need contributors to submit stability fixes rather than features.

Issue #48, opened 2025-07-28
White menu bar icon


marksteve:
<img width="992" height="72" alt="Image" src="https://github.com/user-attachments/assets/44811fda-825d-4ec0-816c-225d419f5e2a" />

cjpais:
How did you change the color to black? Does this just happen with different backgrounds for contrast? _(edit: yes, I set my background to white and it very much makes the icon disappear)_. Thanks for pointing this out @marksteve 

Also noting this is definitely an issue on Windows as well, but just curious so I can test on macOS

@vladstudio If you get a chance and have time, would you mind adding black icon .pngs and calling `setIconAsTemplate` ([see this](https://docs.rs/tauri/latest/tauri/tray/struct.TrayIcon.html#method.set_icon_as_template)) on macOS? And if not no worries I can get to it. We can probably make the default black, and I think this may resolve this issue and make it more visible on Windows. Linux we may need white still, but not sure, before I pull whatever changes in I will make sure to test on all platforms.

cjpais:
@marksteve this will be resolved when 0.3.2 is released, I am planning to pull in a few other changes and then I will release it

marksteve:
Sorry for the late response but thanks for fixing this! Much appreciated! ✋

PR #957, opened 2026-03-04
model: Add Qwen3-ASR batch transcription engine.

andrewleech:
### Summary

Qwen3-ASR is a multilingual speech recognition model (12 languages) with two size variants (0.6B, 1.7B). I'm adding it as a new batch transcription engine backed by [transcribe-rs PR #48](https://github.com/cjpais/transcribe-rs/pull/48).

Qwen3 models are distributed as individual ONNX files (encoder, decoder, embeddings, tokenizer, config) rather than a single tar.gz archive. The existing download infrastructure only handled single-file tar.gz downloads, so the model manager now supports multi-file downloads with per-file SHA-256 verification, resume, and cancellation. The shared download logic was extracted into a `download_single_file` helper used by both the single-file and multi-file paths.

```mermaid
flowchart LR
    A[download_model] --> B{files empty?}
    B -->|yes| C[download_single_file<br/>tar.gz path]
    B -->|no| D[download_model_files]
    D --> E[verify existing files<br/>SHA-256 + size]
    E --> F[download_single_file<br/>per missing file]
    F --> G[remove .incomplete marker]
```

The second commit (`chore: temporarily pin transcribe-rs to git branch`) should be dropped once the upstream transcribe-rs PR merges and a crates.io release includes qwen3 support.

### Testing

- `cargo check` clean, 35 unit tests pass
- Frontend lint clean
- Not yet runtime-tested (model files not yet uploaded to HuggingFace)

The model registry entries are verbose (~160 lines each for the 9-file manifests with SHA-256 hashes). An alternative would be loading manifests from external JSON, but that adds a fetch dependency to model discovery. Keeping them inline is consistent with existing model entries and avoids the extra complexity.

### DIFF
--- a/src-tauri/Cargo.toml
+++ b/src-tauri/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "handy"
-version = "0.8.0"
+version = "0.7.11"
 description = "Handy"
 authors = ["cjpais"]
 edition = "2021"
@@ -68,8 +68,7 @@ chrono = "0.4"
 rusqlite = { version = "0.37", features = ["bundled"] }
 tar = "0.4.44"
 flate2 = "1.0"
-sha2 = "0.10"
-transcribe-rs = { version = "0.3.2", features = ["whisper-cpp", "onnx"] }
+transcribe-rs = { git = "https://github.com/andrewleech/transcribe-rs", branch = "feat/qwen3-batch", features = ["whisper-cpp", "onnx", "qwen3"] }
 handy-keys = "0.2.4"
 ferrous-opencc = "0.2.3"
 clap = { version = "4", features = ["derive"] }
@@ -88,7 +87,7 @@ tauri-plugin-single-instance = "2.3.2"
 tauri-plugin-updater = "2.10.0"
 
 [target.'cfg(windows)'.dependencies]
-transcribe-rs = { version = "0.3.2", features = ["whisper-vulkan", "ort-directml"] }
+transcribe-rs = { git = "https://github.com/andrewleech/transcribe-rs", branch = "feat/qwen3-batch", features = ["ort-directml", "qwen3"] }
 windows = { version = "0.61.3", features = [
   "Win32_Media_Audio_Endpoints",
   "Win32_System_Com_StructuredStorage",
@@ -100,12 +99,10 @@ winreg = "0.55"
 
 [target.'cfg(target_os = "macos")'.dependencies]
 tauri-nspanel = { git = "https://github.com/ahkohd/tauri-nspanel", branch = "v2.1" }
-transcribe-rs = { version = "0.3.2", features = ["whisper-metal"] }
 
 [target.'cfg(target_os = "linux")'.dependencies]
 gtk-layer-shell = { version = "0.8", features = ["v0_6"] }
 gtk = "0.18"
-transcribe-rs = { version = "0.3.2", features = ["whisper-vulkan"] }
 
 [patch.crates-io]
 tauri-runtime = { git = "https://github.com/cjpais/tauri.git", branch = "handy-2.10.2" }
--- a/src-tauri/Cargo.toml
+++ b/src-tauri/Cargo.toml
@@ -69,7 +69,7 @@ rusqlite = { version = "0.37", features = ["bundled"] }
 tar = "0.4.44"
 flate2 = "1.0"
 sha2 = "0.10"
-transcribe-rs = { version = "0.3.8", features = ["whisper-cpp", "onnx"] }
+transcribe-rs = { git = "https://github.com/andrewleech/transcribe-rs", branch = "feat/qwen3-batch", features = ["whisper-cpp", "onnx", "qwen3"] }
 handy-keys = "0.2.4"
 ferrous-opencc = "0.2.3"
 clap = { version = "4", features = ["derive"] }
@@ -88,7 +88,7 @@ tauri-plugin-single-instance = "2.3.2"
 tauri-plugin-updater = "2.10.0"
 
 [target.'cfg(windows)'.dependencies]
-transcribe-rs = { version = "0.3.3", features = ["whisper-vulkan", "ort-directml"] }
+transcribe-rs = { git = "https://github.com/andrewleech/transcribe-rs", branch = "feat/qwen3-batch", features = ["ort-directml", "qwen3"] }
 windows = { version = "0.61.3", features = [
   "Win32_Media_Audio_Endpoints",
   "Win32_System_Com_StructuredStorage",
@@ -100,12 +100,10 @@ winreg = "0.55"
 
 [target.'cfg(target_os = "macos")'.dependencies]
 tauri-nspanel = { git = "https://github.com/ahkohd/tauri-nspanel", branch = "v2.1" }
-transcribe-rs = { version = "0.3.3", features = ["whisper-metal"] }
 
 [target.'cfg(target_os = "linux")'.dependencies]
 gtk-layer-shell = { version = "0.8", features = ["v0_6"] }
 gtk = "0.18"
-transcribe-rs = { version = "0.3.3", features = ["whisper-vulkan"] }
 
 [patch.crates-io]
 tauri-runtime = { git = "https://github.com/cjpais/tauri.git", branch = "handy-2.10.2" }
--- a/src-tauri/src/managers/model.rs
+++ b/src-tauri/src/managers/model.rs
@@ -26,6 +26,7 @@ pub enum EngineType {
     SenseVoice,
     GigaAM,
     Canary,
+    Qwen3,
 }
 
 #[derive(Debug, Clone, Serialize, Deserialize, Type)]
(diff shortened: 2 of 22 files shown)


cjpais:
We probably should do a tar if we are bringing this in rather than a new methodology.

We can do verification but it should be a separate PR

andrewleech:
> We probably should do a tar if we are bringing this in rather than a new methodology.

Do you mean tar the individual model files into a single download? 

Would you want to keep it as a tar on disk and unpack on the fly during load, or just unpack at the end of download?

cjpais:
Look at how the other ones are done

src-tauri/Cargo.toml
```diff
@@ -68,7 +68,7 @@ chrono = "0.4"
 rusqlite = { version = "0.37", features = ["bundled"] }
 tar = "0.4.44"
 flate2 = "1.0"
-transcribe-rs = { version = "0.2.8", features = ["whisper", "parakeet", "moonshine", "sense_voice", "gigaam"] }
+transcribe-rs = { git = "https://github.com/andrewleech/transcribe-rs", branch = "feat/qwen3-batch", features = ["whisper", "parakeet", "moonshine", "sense_voice", "gigaam", "qwen3"] }
```
VirenMohindra: this pins the entire project to a personal fork branch. if `andrewleech/transcribe-rs` `feat/qwen3-batch` goes away, handy can't build. this should wait for upstream [transcribe-rs PR](https://github.com/cjpais/transcribe-rs/pull/48) to merge and ship on crates.io before merging.

src-tauri/src/managers/model.rs
```diff
@@ -424,6 +425,61 @@ impl ModelManager {
...
+                description: "Multilingual batch transcription with good accuracy-efficiency trade-off.".to_string(),
+                filename: "qwen3-asr-0.6b".to_string(),
+                url: Some("https://huggingface.co/andrewleech/qwen3-asr-0.6b-onnx/resolve/main/qwen3-asr-0.6b.tar.gz".to_string()),
+                size_mb: 2750,
+                is_downloaded: false,
+                is_downloading: false,
+                partial_size: 0,
+                is_directory: true,
+                engine_type: EngineType::Qwen3,
+                accuracy_score: 0.75,
```
VirenMohindra, line 641: these huggingface URLs point to personal repos (`andrewleech/qwen3-asr-*-onnx`). are the model files uploaded yet? the PR description says "not yet runtime-tested (model files not yet uploaded to HuggingFace)" — nobody can verify this works until they are. the 1.7B model is 7.3 GB, worth runtime testing before merge.

src-tauri/src/managers/model.rs
```diff
@@ -692,7 +748,7 @@ impl ModelManager {
                     is_recommended: false,
                     supported_languages: vec![],
                     is_custom: true,
-                },
+                    },
```
VirenMohindra: unrelated whitespace change — should be dropped from this PR.

src-tauri/src/managers/transcription.rs
```diff
@@ -549,6 +569,9 @@ impl TranscriptionManager {
                         LoadedEngine::GigaAM(gigaam_engine) => gigaam_engine
                             .transcribe_samples(audio, None)
                             .map_err(|e| anyhow::anyhow!("GigaAM transcription failed: {}", e)),
```
VirenMohindra, line 629: `transcribe_samples(audio, None)` — no inference params despite qwen3 supporting 12 languages. fine for an initial PR but worth a follow-up to wire up language selection like whisper/sensevoice do. users selecting e.g. `zh-Hans` in settings won't get that passed through to the engine.

VirenMohindra (commented):
nice work on the engine integration — follows the existing patterns cleanly. left a few inline comments on things to address.

src-tauri/Cargo.toml
andrewleech: Yes this is all still WIP (should have made this a draft pr sorry - If/when my library PR there is finished & merged I'll fix this.

src-tauri/src/managers/model.rs
andrewleech: I've been fixing / testing / improving the models in various formats over the last few days, the current uploads are hopefully pretty well release ready - but I'm just reviewing my transcribe-rs changes more thoroughly which use these before I push that up again, then I'll update this PR to match

andrewleech:
**Updated to address review feedback and rebase onto current main.**

Changes since the original PR:

- **Switched from multi-file downloads to tar.gz archives** as requested — Qwen3 models now download as a single `.tar.gz` and extract to a directory, matching the existing Parakeet/Moonshine/SenseVoice pattern. The `ModelFile` struct, per-file SHA-256 verification, `.incomplete` markers, and `download_model_files()` method are all removed. The `sha2` crate dependency is also dropped.

- **Rebased onto current main** including the transcribe-rs 0.3.1 migration (`SpeechModel` trait, `onnx::*` module paths), Canary engine, DirectML acceleration, and other upstream changes. The Qwen3 integration now uses the new API (`Qwen3Model::load(&path, &Quantization)` + `model.transcribe(&audio, &options)`).

- **Expanded language support** from 12 to 55 languages, matching the engine's actual `KNOWN_LANGUAGES` list cross-referenced against the frontend language constants.

- **Translated model descriptions** in all 17 locale files (previously English-only placeholders in non-English locales).

- **Added HTTP `read_timeout(30s)`** to the download client and documented lock ordering in `finish_downloading`.

- **Int4 quantized variants** are supported in the transcription code path (quantization derived from model ID suffix) but the int4 model entries are not yet registered — they'll be added once the int4 tar.gz archives finish uploading to HuggingFace.

- **FP32 and Int4 model archives**:
  - https://huggingface.co/andrewleech/qwen3-asr-0.6b-onnx (uploaded)
  - https://huggingface.co/andrewleech/qwen3-asr-1.7b-onnx (upload in progress)

The DROP commit (`pin transcribe-rs to git branch`) should be dropped when https://github.com/cjpais/transcribe-rs/pull/48 is merged.

VirenMohindra:
great job. I'll take a look at this holistically and test e2e once https://github.com/cjpais/transcribe-rs/pull/48 lands. feel free to ping me directly!

src-tauri/Cargo.toml
andrewleech: Agreed — this is in a separate `DROP:` commit at the tip specifically so it can be dropped cleanly when [transcribe-rs PR #48](https://github.com/cjpais/transcribe-rs/pull/48) merges. The commit message makes this explicit.

src-tauri/src/managers/model.rs
andrewleech: Models are now uploaded and download URLs are live — both FP32 and Int4 variants for 0.6B and 1.7B. Runtime tested on Windows with the 0.6B Int4 model.

src-tauri/src/managers/model.rs
andrewleech: Resolved in subsequent rebases — the current diff has no stray whitespace changes.

src-tauri/src/managers/transcription.rs
andrewleech: Qwen3-ASR auto-detects language from the audio — there's no language token in the prompt template (unlike Whisper). The transcribe-rs Qwen3 engine only exposes max_tokens, with no language field. supports_language_selection: false is set correctly to prevent the UI from showing a language selector that would have no effect. If the engine adds language hints in the future, this can be wired up then.

andrewleech: Good call — this is feasible as a follow-up. I investigated the Qwen3-ASR prompt structure and there is a mechanism for language hinting.

**How the model handles language:**

The model outputs `language {Name}<asr_text>{transcription}` — it auto-detects the language, emits a language tag, then the `<asr_text>` special token (ID 151704), then the transcription. The `strip_language_prefix` function in transcribe-rs removes this prefix before returning the result.

**How language forcing would work:**

Instead of letting the model generate the language tag, you prefill the assistant turn with the language prefix tokens. The prompt currently ends with:

```
<|im_start|>assistant\n
```

To force e.g. Chinese, you'd extend it to:

```
<|im_start|>assistant\n language Chinese<asr_text>
```

The token IDs are: `[11528, <lang_name_token>, 151704]` where 11528 = "language", the language name is a single BPE token (e.g. 6364=English, 8453=Chinese, 10769=Japanese), and 151704 = `<asr_text>`.

The decoder then starts generating from after the prefilled tokens, skipping auto-detection entirely and producing transcription in the specified language.

**What would need to change (transcribe-rs):**

- Add a language-name-to-token-ID mapping (~57 entries matching `KNOWN_LANGUAGES`)
- Add `language: Option<String>` to `Qwen3Params`
- Extend `build_prompt_ids` to append `[11528, lang_token, 151704]` after the assistant prefix when language is set
- Adjust `strip_language_prefix` to not strip when language was forced (it's already been consumed by the prefill)

**What would need to change (Handy):**

- Pass `options.language` through to the Qwen3 engine (currently uses `TranscribeOptions::default()` since `supports_language_selection: false`)
- Set `supports_language_selection: true` on the Qwen3 model entries
- The frontend language selector would then appear for Qwen3 models, and the selected language would flow through `TranscribeOptions.language` → `Qwen3Params.language` → prefilled prompt tokens

I'll track this as a follow-up to the transcribe-rs Qwen3 engine.

cjpais:
Closing this, as it will come in as part of a separate set of changes I am working on behind the scenes, should have very good Qwen3 perf

andrewleech:
Looking forward to what you've got cooking - I'd still like to see better performance then I could get with ort... are you integrating llama.cpp directly by any chance, I see some qwen asr support has been started there ;-)

If your new method needs some testing on high core count amd or integrated amd GPU I'm happy to help, also happy to let you work in peace!

cjpais:
@andrewleech I've been building transcribe.cpp (hopefully releasing in a couple weeks) which implements all the models (and more) currently supported in transcribe-rs, but in ggml. Written from the ground up, not a fork of whisper.cpp or llama.cpp but built on the same ggml foundation. Vulkan and CPU performance seem to be quite good :) I bet there's quite a lot of performance still on the table as well. Also WER's are very good.

If you're curious to try I can give early access, the repo is private right now just so I can work a bit in the quiet. It's getting very close now though

## Qwen 3 ASR 0.6b

### WER

| Quantization | Size | WER (LibriSpeech test-clean) |
|---|---|---|
| BF16 | 1.46 GB | 2.12% |
| F16 | 1.47 GB | 2.12% |
| Q8_0 | 811 MB | 2.11% |
| Q6_K | 658 MB | 2.11% |
| Q5_K_M | 615 MB | 2.21% |
| Q4_K_M | 562 MB | 2.26% |

### Apple M4 Max

| Backend | Sample       |          Q8_0 |        Q4_K_M |
| ------- | ------------ | ------------: | ------------: |
| Metal   | jfk (11.0s)  |  155 ms (71×)  |  142 ms (77×)  |
| Metal   | dots (35.3s) |  597 ms (59×)  |  527 ms (67×)  |
| CPU     | jfk (11.0s)  |  660 ms (17×)  |  588 ms (19×)  |
| CPU     | dots (35.3s) | 2.26 s (16×)   | 2.10 s (17×)   |

macOS 26.3.1, transcribe.cpp `0c88a71`.

### AMD Ryzen 7 4750U Pro

| Backend | Sample       |          Q8_0 |        Q4_K_M |
| ------- | ------------ | ------------: | ------------: |
| Vulkan  | jfk (11.0s)  | 1.27 s (8.7×)  | 1.08 s (10.1×) |
| Vulkan  | dots (35.3s) | 4.87 s (7.3×)  | 3.99 s (8.9×)  |
| CPU     | jfk (11.0s)  | 2.37 s (4.6×)  | 1.92 s (5.7×)  |
| CPU     | dots (35.3s) | 8.61 s (4.1×)  | 7.34 s (4.8×)  |

Fedora 43, transcribe.cpp `3d16f74`. Vulkan device: `AMD Radeon
Graphics (RADV RENOIR)`.

andrewleech:
Sounds fantastic! Did you quantize the models yourself? I tested a qwen3 asr Q4_K_M from HF (mradermacher) quickly the other day with llama.cpp and the quality was broken, text from standard JFK was more wrong than right.

cjpais:
Yeah I quantized them and verified my implementation end to end for numerical accuracy and WER performance. I've done the same for all the models I've implemented, takes a long time but I think it's much nicer to be able to validate we are able to match the expected performance of the model released on HF

I think Q8/Q6 will ultimately be shipped in Handy. Q4 looses too much on most models for modest gains

There's a lot of sub par implementations out there. Not sure why llama.cpp is broken but certainly transcribe.cpp should have a working version

PR #49, opened 2025-07-28
Cancel actions in menu

vladstudio:
Users can now stop recording or cancel transcription directly from the system tray menu instead of having to wait for them to finish. Useful when you need to quickly cancel an operation, e.g. having it recording for hours accidentally :-).

PR #958, opened 2026-03-04
gpu: Add runtime GPU execution provider selection.

andrewleech:
### Summary

ORT-based transcription engines (Parakeet, Moonshine, SenseVoice) can use GPU acceleration via ONNX Runtime execution providers, but until now the build had no way to select one. This adds compile-time Cargo feature flags (`gpu-directml`, `gpu-cuda`, `gpu-coreml`, `webgpu`) that gate the available providers, plus a runtime settings dropdown that lets the user switch between them without restarting.

Backed by [transcribe-rs PR #49](https://github.com/cjpais/transcribe-rs/pull/49) which adds `GpuProvider` enum, `set_gpu_provider()`, `available_providers()`.

```mermaid
flowchart LR
    A[Startup] --> B{Persisted provider<br/>available in build?}
    B -->|yes| C[set_gpu_provider]
    B -->|no| D[Reset to auto<br/>+ warn]
    D --> C
    E[User changes setting] --> F{Model loaded?}
    F -->|ORT engine| G[Unload + reload<br/>with new EP]
    F -->|Whisper| H[Skip reload<br/>whisper.cpp ignores EP]
    F -->|Transcription in flight| I[Reject + revert]
```

The dropdown only appears when a GPU EP beyond `auto`+`cpu` is compiled in, so default builds see no UI change.

The second commit (`chore: temporarily pin transcribe-rs to git branch`) should be dropped once the upstream transcribe-rs PR merges and a crates.io release includes GPU provider support.

#### Benchmarks

Tested on AMD Ryzen AI 9 365 (Zen 5) with integrated AMD Radeon 860M iGPU. The autoregressive decoding loop in current ORT-based engines doesn't benefit from GPU parallelism — iGPU memory bandwidth is shared with CPU and per-token overhead dominates.

**DirectML** (Qwen3-ASR decoder only — encoder hits a DML Conv2d NaN bug and must stay on CPU):

| Metric | Result |
|---|---|
| 1.7B decoder | ~7% speedup |
| 0.6B decoder | ~9% slowdown (GPU dispatch overhead > benefit) |

**WebGPU** (Qwen3-ASR encoder + decoder, native Windows, D3D12 backend):

| Model | Audio | CPU RTF | CPU time | WebGPU RTF | WebGPU time | Delta |
|---|---|---|---|---|---|---|
| 0.6B | 11s | 2.54x | 4.33s | 2.42x | 4.55s | -5% |
| 0.6B | 30s | 1.45x | 20.37s | 1.38x | 21.45s | -5% |
| 1.7B | 11s | 1.31x | 8.38s | 1.16x | 9.47s | -12% |
| 1.7B | 30s | 0.98x | 30.18s | 0.66x | 45.16s | -50% |

On this hardware, GPU acceleration doesn't help. The infrastructure is in place for users with discrete GPUs where the cost/benefit may be different, and the setting defaults to `auto` so there's no regression for everyone else.

### Testing

- `cargo check` clean, 33 unit tests pass
- Frontend lint clean
- Runtime-tested on Windows with `--features gpu-directml` (setting persists, model reloads, stale provider resets correctly on next launch)

Whisper models (whisper.cpp backend) don't use ORT, so GPU provider changes are intentionally skipped for them — the reload would be a no-op. An alternative would be hiding the setting entirely when only Whisper models are available, but the current approach is simpler and the dropdown already hides itself when no GPU EP is compiled in.

### DIFF
--- a/src-tauri/Cargo.lock
+++ b/src-tauri/Cargo.lock
@@ -1444,7 +1444,7 @@ version = "0.5.2"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "330c60081dcc4c72131f8eb70510f1ac07223e5d4163db481a04a0befcffa412"
 dependencies = [
- "libloading 0.8.9",
+ "libloading 0.7.4",
 ]
 
 [[package]]
@@ -4100,6 +4100,7 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "e2aba9f5c7c479925205799216e7e5d07cc1d4fa76ea8058c60a9a30f6a4e890"
 dependencies = [
  "flate2",
+ "glob",
  "pkg-config",
  "sha2",
  "tar",
@@ -4125,7 +4126,7 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "7d8fae84b431384b68627d0f9b3b1245fcf9f46f6c0e3dc902e9dce64edd1967"
 dependencies = [
  "libc",
- "windows-sys 0.61.2",
+ "windows-sys 0.45.0",
 ]
 
 [[package]]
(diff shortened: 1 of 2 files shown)


cjpais:
Okay, I'm open to this, but I really need you to do some work on the CI/CD too. I get it works for you at build time but this cannot be pulled in if it cannot be distributed across platforms properly.

andrewleech:
To be fair it doesn't work for me, but that's an AMD laptop issue (even though they brand it as an AI laptop) ;-) 

Good point regardless, I'm sorry for throwing this up without tests, my bad. All the testing scripts were too local looking to include in the commit, I overlooked that was all I had. 

I'll look into what can be covered in a GitHub runner, if nothing else the program flow/logic.

cjpais:
im going to close this, but we will add the option, I just want to start it from scratch

PR #590, opened 2026-01-15
feat(history): add search functionality

rdadm0:
## Before Submitting This PR

**Please confirm you have done the following:**

- [x] I have searched [existing issues](https://github.com/cjpais/Handy/issues) and [pull requests](https://github.com/cjpais/Handy/pulls) (including closed ones) to ensure this isn't a duplicate
- [x] I have read [CONTRIBUTING.md](../CONTRIBUTING.md)

## Human Written Description

I wanted to be able to search through my transcription history to find past recordings. With many transcriptions building up over time, it becomes hard to find specific ones without scrolling through everything.

## Related Issues/Discussions

N/A - new feature suggestion

## Testing

- Tested search filtering with various queries
- Verified Cmd+F keyboard shortcut focuses the search input
- Confirmed clear button (X) resets search
- Checked "no results" message displays correctly

## Screenshots/Videos (if applicable)

N/A

## AI Assistance

- [ ] No AI was used in this PR
- [x] AI was used (please describe below)

**If AI was used:**

- Tools used: Claude Code (Claude Opus 4.5)
- How extensively: AI wrote the implementation code; human provided direction and testing

PR #975, opened 2026-03-06
feat(history): add search functionality to history tab

AniruddhaChattopadhyay:
## Before Submitting This PR

## Human Written Description

I found myself scrolling through my entire transcription history trying to find a specific past recording. With auto-delete disabled, the list grows quickly and there's no way to locate anything without reading through every entry. A simple search bar at the top of the History tab would make this much more usable.

A previous attempt at this (#590) was self-closed because it broke audio playback. My implementation takes a simpler approach — it only filters the already-loaded list client-side without touching audio loading, so that issue doesn't apply here.

## Related Issues/Discussions

- Previous attempt: #590 (self-closed by author due to audio playback bug)
- Discussion: #974

## Community Feedback

Discussion opened at https://github.com/cjpais/Handy/discussions/974

The author of the previous PR #590 noted: *"search option would be really useful if implemented well"*. This implementation avoids the audio playback issue that caused #590 to be closed.

## Testing

- Verified search filters entries by transcription text (case-insensitive)
- Verified search also matches against post-processed text
- Confirmed "No matching transcriptions found" message appears when search has no matches
- Confirmed audio playback still works correctly on filtered entries
- Confirmed clearing the search input restores the full list
- Verified ESLint passes cleanly

## AI Assistance

**If AI was used:**

- Tools used: Claude Code
- How extensively: AI helped with implementation; human provided direction, testing, and review

## Implementation Notes

**Why client-side filtering:** The default history limit is 5 entries. Even users who increase this or disable auto-delete are unlikely to have thousands of entries. Filtering the already-loaded array is instant and avoids unnecessary backend complexity (no new Tauri commands or database query changes needed).

**Changes:**
- `src/components/settings/history/HistorySettings.tsx` — Added search input with icon, search state, and filtering logic
- `src/i18n/locales/en/translation.json` — Added `search` and `noResults` translation keys

### DIFF
--- a/src/components/settings/history/HistorySettings.tsx
+++ b/src/components/settings/history/HistorySettings.tsx
@@ -2,7 +2,7 @@ import React, { useState, useEffect, useCallback } from "react";
 import { useTranslation } from "react-i18next";
 import { AudioPlayer } from "../../ui/AudioPlayer";
 import { Button } from "../../ui/Button";
-import { Copy, Star, Check, Trash2, FolderOpen } from "lucide-react";
+import { Copy, Star, Check, Trash2, FolderOpen, Search } from "lucide-react";
 import { convertFileSrc } from "@tauri-apps/api/core";
 import { listen } from "@tauri-apps/api/event";
 import { readFile } from "@tauri-apps/plugin-fs";
@@ -36,6 +36,7 @@ export const HistorySettings: React.FC = () => {
   const osType = useOsType();
   const [historyEntries, setHistoryEntries] = useState<HistoryEntry[]>([]);
   const [loading, setLoading] = useState(true);
+  const [searchQuery, setSearchQuery] = useState("");
 
   const loadHistoryEntries = useCallback(async () => {
     try {
@@ -182,6 +183,17 @@ export const HistorySettings: React.FC = () => {
     );
   }
 
+  const filteredEntries = searchQuery
+    ? historyEntries.filter((entry) => {
+        const query = searchQuery.toLowerCase();
+        return (
+          entry.transcription_text.toLowerCase().includes(query) ||
+          (entry.post_processed_text &&
+            entry.post_processed_text.toLowerCase().includes(query))
+        );
+      })
+    : historyEntries;
+
   return (
     <div className="max-w-3xl w-full mx-auto space-y-6">
       <div className="space-y-2">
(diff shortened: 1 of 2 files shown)


cjpais:
Feature requests are not being accepted at this time.

You also didnt provide screenshots or videos, so I cannot even attempt to review without pulling the code myself. Make it easier for me.

AniruddhaChattopadhyay:
My apologies, Here is a video of this functionality working. I'm a huge fan of the product and appreciate all the work that you're doing. 

https://github.com/user-attachments/assets/e9dad7b7-1372-428d-be20-d77341786f85

cjpais:
Can you make the UI more consistent. Like make sure you're using the same corner radius. Also it looks like it doesn't span all the way to the same width of the history boxes which is not very good looking imo. Spend some time and make the visual design consistent.

AniruddhaChattopadhyay:
Sure. Let me work on making the design consistent and revise the PR. Will make sure to add screenshots and videos this time 😊 

Appreciate you willing to consider this.

AniruddhaChattopadhyay:
@cjpais thanks for the feedback! I've reworked the search input, here's what I did:
                                                                                             
Moved the search bar inside the card instead of having it float separately above it. That fixes the width issue since it's now part of the same container. Also updated the border radius, border colors, and focus styles to match whats already there. Sharing the before and after picture here.

Before-After

I basically followed the same pattern that's used in LanguageSelector.tsx and ModelsSettings.tsx , both of those have search inputs sitting inside their parent container with a border separator, so figured it made sense to do the same thing here to keep things consistent.

For the focus state, I went with the pink background fill (like in Input.tsx and Textarea.tsx) rather than just a border highlight, since that seems to be the more common pattern across the app. But I'm open to going either way on this — happy to switch to border-only if you think that fits better here.

Here is a snapshot of how just the border highlight looks instead of background fill. 
image

Let me know if there's anything else you'd like me to adjust!

Finally the video of the entire thing. 


https://github.com/user-attachments/assets/e43288e8-9858-4453-ad99-e9ea0ac290c8

AniruddhaChattopadhyay:
@cjpais just a gentle nudge on this:) happy to incorporate any more changes if needed. 

I can totally understand if this is not a priority at the moment.

PR #988, opened 2026-03-10
fix: makes postprocessing key actually optional

### LINKED ISSUE #987: [BUG] "Optional" post-processing hotkey (#355) is not actually optional
## Before You Submit

**Please search [existing issues](https://github.com/cjpais/Handy/issues) to avoid duplicates.** Your bug may already be reported! Right now it's just me maintaining this project so many issues can be overwhelming! Help me out by checking first.

## Bug Description

The "optional" post-processing hotkey introduced in #355 that is enabled when turning on post-processing in the Experimental settings is not actually optional. There is no way to unset it, and it is the only way to trigger postprocessing. This effectively means that to have postprocessing always enabled, the user would have to set their "main" hotkey to some random binding, and set the "postprocessing hotkey" to the hotkey they actually want to use.

## System Information

**App Version:** 0.7.9

**Operating System:** macOS 26

**CPU:** Apple M1 Pro

**GPU:** Apple M1 Pro

snarkyerica:
## Before Submitting This PR

## Human Written Description

I filed issue #987 as the dedicated post-processing key that was implemented in #355 was not actually optional as stated in the original PR. This PR fixes that behavior, making the post-processing key truly optional (the default binding is unset) and enabling post-processing on the main hotkey whenever the dedicated key is unset.

## Related Issues/Discussions

Fixes #987, related to #355

## Testing

Changes build, and have been tested locally by me, seems to work okay, the behavior is now as expected and there's no obvious regressions or crashes.

## AI Assistance

**If AI was used:**

- Tools used: Codex
- How extensively: Drafted the initial implementation plan, helped locate the issue in the codebase, implemented the Rust side, generated commit messages.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,7 +1,7 @@
 {
   "name": "handy-app",
   "private": true,
-  "version": "0.7.9",
+  "version": "0.7.10",
   "type": "module",
   "scripts": {
     "dev": "vite",
--- a/src-tauri/Cargo.lock
+++ b/src-tauri/Cargo.lock
@@ -2416,7 +2416,7 @@ dependencies = [
 
 [[package]]
 name = "handy"
-version = "0.7.9"
+version = "0.7.10"
 dependencies = [
  "anyhow",
  "chrono",
--- a/src-tauri/Cargo.toml
+++ b/src-tauri/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "handy"
-version = "0.7.9"
+version = "0.7.10"
 description = "Handy"
 authors = ["cjpais"]
 edition = "2021"
--- a/src-tauri/src/actions.rs
+++ b/src-tauri/src/actions.rs
@@ -17,8 +17,8 @@ use once_cell::sync::Lazy;
 use std::collections::HashMap;
 use std::sync::Arc;
 use std::time::Instant;
-use tauri::AppHandle;
 use tauri::Manager;
+use tauri::{AppHandle, Emitter};
 
 /// Drop guard that notifies the [`TranscriptionCoordinator`] when the
 /// transcription pipeline finishes — whether it completes normally or panics.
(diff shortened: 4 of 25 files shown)


cjpais:
Let me think about this. I think it's probably a good change. I removed the optionality on purpose because I felt it was easier to reason about. 

You know you have one shortcut for regular and one shortcut for post processing. 

I'm not sure the best UX here generally. As it gets complicated otherwise. Mainly how is someone to know that the behavior is optional?  Post processing is in alpha for a few reasons and I'm partially willing to accept imperfection, but I think trying to get as clear of a UX is the primary goal. For example when out of alpha I would expect this shortcut to be in general settings. Possibly removing the second shortcut entirely, and just a switch to enable or disable the feature, as it should be good enough on average to always have on. That's the goal we are reaching for long term

femaphore:
Yeah, I mean, the main issue I had is that the "optional" wording suggests that the keybind is optional for an "always postprocess" behavior.

The problem I found is that well, if I want postprocessing enabled on all transcriptions, I would have to set the "main" keybind to something I would never use (right now, in the release version, I have it set to Shift+Ctrl+Opt+Cmd+Space -- it's a bit obnoxious), and then set the "postprocess always" keybind to the keybind I actually intend to use (Opt+Cmd+Space, for instance), as either keybind cannot be unset.

It feels kind of unclean to have one keybind that must always be set that wouldn't get used in that scenario.

cjpais:
I understand, but things move fast and I changed the code since that PR I believe

I see what you mean, but I don't think this is that huge of a deal. At least to me, I just have two bindings set (`fn` for normal and `opt+space` for post process). It kind of just works, and if I wanted different I just have to swap them. Largely I feel this is good enough, and most of this app is just built on my preferences.

It's generally not clean having multiple bindings I feel, and would rather try to collapse to one shortcut generally. I understand the frustration and also fully encourage forks :) If you have ideas on how to support forks better or how we can pull in certain features and stuff happy to hear

femaphore:
Alright, I'll keep the fork up and carry the branch forward as things get merged until the shortcuts get collapsed later on, it's been essentially the only issue I've had, so there's not really much I'd pull in to a fork otherwise. Thanks for hearing me out! It's very appreciated.

cjpais:
No worries, I think this is a super valid discussion and it's not really clear fully the best approach

I think it honestly could use wider discussion in the community and generally how to do UX better across the app

PR #991, opened 2026-03-10
Add Windows microphone permission onboarding

ferologics:
## Summary

Adds a Windows microphone-permission onboarding flow.

### What this does
- checks Windows microphone privacy state during onboarding / startup
- routes Windows users to the permissions screen when microphone access is blocked
- opens `Privacy & security -> Microphone` from the onboarding button
- forces the main window visible on startup when Windows microphone permissions are denied for a returning user, even if `start_hidden` is enabled
- cleans up Rust/build warning noise touched during implementation

## Context

This is a follow-up to the Windows mic startup crash fix that was merged in:
- #945

## Local verification

Tested locally on Windows:
- with microphone privacy toggles disabled, Handy shows the permissions screen
- the settings button opens Windows microphone privacy settings
- re-enabling the toggles while Handy is open unblocks the flow into the main app

## Checks

- `bun run lint`
- `bun run build`
- `cargo check`

### DIFF
--- a/src-tauri/build.rs
+++ b/src-tauri/build.rs
@@ -84,7 +84,11 @@ fn generate_tray_translations() {
 
     fs::write(Path::new(&out_dir).join("tray_translations.rs"), out).unwrap();
 
-    let _ = (translations.len(), fields.len());
+    println!(
+        "cargo:warning=Generated tray translations: {} languages, {} fields",
+        translations.len(),
+        fields.len()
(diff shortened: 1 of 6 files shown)


cjpais:
I will test this at the same time as #985 and likely pull both in, thank you!

cjpais:
Tested this earlier and it builds clean, merged

PR #1001, opened 2026-03-11
feat: rewrite selected text with voice instruction

Piping:
## What this adds
- New shortcut action `rewrite_selected_with_voice_instruction`.
- On key press: capture currently selected text (copy/snapshot).
- On key release: transcribe the spoken instruction, rewrite the snapshot via configurable LLM provider/model/prompt, and replace the selection.
- Failure UX: does not modify text and emits a toast error event (`rewrite-selected-error`).

## Configuration
- Adds a separate "Rewrite" LLM config surface (provider, API key, model, prompts) independent from post-processing.
- Adds a third default shortcut (macOS: `option+command+space`).
- Prevents duplicate shortcut bindings across actions.

## Notes
- Selection snapshot uses a clipboard sentinel + copy shortcut and restores the prior clipboard value.

## Testing
- `bun run build` passes locally.
- I couldn’t run the dev binary due to an AMFI policy requiring CMS-signed binaries functionality should be testable on a normal dev macOS environment.


cjpais:
You didn't do anything that I asked in the PR form. And you also should have read that I'm not accepting new features right now, especially ones that have not been requested by the community.

Issue #921, opened 2026-03-01
[BUG] Clipboard is being overwritten


rishiskhare:
## Bug Description

The existing contents of the clipboard are erased after I transcribe. They are not restored to what they were originally.

## System Information

I use an Apple M1 MacBook with the latest version of Handy installed.

cjpais:
What setting are you using for the clipboard handling in advanced? This might be expected if you've changed the setting.

rishiskhare:
Here are my settings: 

<img width="517" height="188" alt="Image" src="https://github.com/user-attachments/assets/7d45c2d8-e1d6-4b31-a688-aaa6e3998294" />

cjpais:
Can you provide a log and does this happen every time?

rishiskhare:
Sure, could you share how to access the logs for clipboard to share?

cjpais:
Go to the about page..

rishiskhare:
More specifically, try copying an image first, then using Handy. The image will no longer be in the clipboard.

cjpais:
That is super helpful context haha, it works fine without images?

rishiskhare:
Yup, text works flawlessly.

NazmusSayad:
Hey, this is kind of a correct solution. We also have a similar [product](https://github.com/oiper/desktop), and we have faced the same issue. But the related PR seems to have lots of changes. It should be a small fix.

We have fixed this by this method which works fine in Mac and windows, Linux needs slightly more care:

1. Store previous clipboard content (can be image or text, if you are using `arboard` you can use HTML too).
2. Paste Target Text
3. Restore previous clipboard content

Here is the exact code snippet from our project:

```rust
pub struct ClipboardContent {
  pub text: Option<String>,
  pub image: Option<tauri::image::Image<'static>>,
}

pub fn get_clipboard_content<R: Runtime>(clipboard: &Clipboard<R>) -> ClipboardContent {
  let text = clipboard.read_text().ok();
  let image = clipboard.read_image().ok().map(|img| img.to_owned());
  ClipboardContent { text, image }
}

pub fn set_clipboard_content<R: Runtime>(
  clipboard: &Clipboard<R>,
  content: &ClipboardContent,
) -> Result<(), String> {
  if let Some(ref txt) = content.text {
    clipboard
      .write_text(txt)
      .map_err(|e| format!("Failed to write text to clipboard: {e}"))?;
  } else if let Some(ref img) = content.image {
    clipboard
      .write_image(img)
[...]

PR #1013, opened 2026-03-12
fix: preserve image clipboard content during paste

sheeki03:
## Before Submitting This PR

## Human Written Description

When you copy an image to the clipboard (e.g. a screenshot) and then use Handy to transcribe speech, the image gets silently destroyed. The clipboard save/restore logic in `paste_via_clipboard` only handled text — so any non-text content was replaced with an empty string after pasting.

This fix makes the clipboard snapshotting format-aware: it detects and preserves images in addition to text. For an empty clipboard or unsupported formats (files, HTML), it now clears the clipboard after pasting instead of leaving the transcript behind.

## Related Issues/Discussions

Partially addresses #921

The issue reports clipboard content being erased after transcribing. This fix covers the two most common clipboard formats (text and images). Other formats (files, HTML) cannot be preserved due to limitations in `arboard` (no `get_html()` API, no file list support), but they are now explicitly cleared rather than leaving the transcript in the clipboard.

## Community Feedback

Issue #921 has 8 comments with active back-and-forth confirming the bug, particularly with image clipboard content being lost.

## Testing

**Automated:**
- 3 new unit tests added covering text round-trip, image round-trip, and empty clipboard behavior
- Tests use `arboard` directly (the same crate backing `tauri-plugin-clipboard-manager`)
- Tests are `#[ignore]` by default since they require a system clipboard and must run serially to avoid SIGSEGV from concurrent pasteboard access on macOS
- Run with: `cargo test --lib -- clipboard::tests::clipboard --test-threads=1 --ignored`
- All 6 tests pass (3 existing + 3 new)

**Manual (macOS, M4 MacBook Air, dev build via `bun run tauri dev`):**
1. Image preservation: Copied screenshot to clipboard → triggered transcription → transcript pasted correctly → `Cmd+V` in Preview restored the original screenshot ✅
2. Text round-trip: Set clipboard to `original text` via `pbcopy` → transcribed → `pbpaste` returned `original text` ✅
3. Empty clipboard: Cleared clipboard → transcribed → `pbpaste` returned empty output (transcript not left behind) ✅

**Known limitation:**
- File and HTML clipboard content cannot be preserved (arboard limitation). After transcription, the clipboard is cleared instead.

N/A — clipboard behavior, not UI changes.

## AI Assistance

**If AI was used:**

- Tools used: Claude Code
- How extensively: Claude Code helped with implementation, code review, and test writing. The approach (format-aware clipboard snapshotting) and manual testing were done collaboratively.

### DIFF
--- a/src-tauri/src/clipboard.rs
+++ b/src-tauri/src/clipboard.rs
@@ -69,7 +69,7 @@ fn restore_clipboard(
         }
         SavedClipboard::Image(image) => {
             if let Err(e) = clipboard.write_image(image) {
-                warn!("Failed to restore image to clipboard: {}", e);
+                warn!("Failed to restore image to clipboard: {e}");
             }
         }
         SavedClipboard::Empty => {


cjpais:
I tried this and it does not work. At least for me on Mac OS. And testing with Figma. I copied an image from the internet and then used handy and then pasted into Figma and it just gave the URL. I try to paste it before I use handy, it worked just fine.

cjpais:
It also fails to build on other platforms.

cjpais:
Closing for now since creator hasn't responded.

PR #1015, opened 2026-03-12
fix: clean up corrupt partial downloads to prevent infinite extraction loop

### LINKED ISSUE #858: [BUG] Partial downloads prevent the model from ever being downloaded again
Installed the app on a brand new pc for the first time and attempted to download the recommended Parakeet V3 model. The download progress seemed to be fine until it reached the extracting step which was taking an abnormally long time, I ended up killing the app and then attempted to re-download the model again. This refused to work as it would seem to trigger and move it to the top but it would then reappear in the model list. Tried a couple other models which were able to be downloaded just fine.

On further inspection, there seemed to be a Partial Download in the AppData folder, deleting this then allowed me to download the model again.

Image

In conclusion, I **believe the problem to be two-fold**, it seems to hang at the extracting step for some models _(seems to be the parakeet ones from my testing)_, a frustrated user would then attempt to restart the app which would then leave it in a broken state.

Image

My suggestion would be to scan for incomplete models/unrecognized files and indicate accordingly in the list (red bg with exclamation) and allow the user to delete these to restart the process or on startup automatically purge these which presumably should be non `.bin` files. Additionally, investigate why the extracting process would run upwards 10 minutes on a successful download.

0.7.6, Windows 10, i7 8700k, GTX 4090

cc @VirenMohindra

sheeki03:
## Before Submitting This PR

## Human Written Description

When a directory-based model download completes but extraction fails (e.g. corrupt archive, interrupted extraction), the `.partial` tar.gz file is left behind. On the next download attempt the app resumes from EOF, thinks the download is complete, and retries extraction on the same broken archive — stuck in an infinite loop with no way to re-download.

This fix ensures the corrupt `.partial` is deleted when extraction fails so the next attempt starts a fresh download. It also distinguishes between archive corruption (delete `.partial`) and environmental failures like permissions or disk space (preserve `.partial` since the archive itself may be valid). The frontend is also fixed to clear stale download/progress state when extraction fails.

## Related Issues/Discussions

Fixes #858

The maintainer suggested: "checksum and if the checksum does not match, delete the file and re-download." This PR addresses the core stuck-download loop without checksums (no canonical hash source is available). Checksum validation can be added as a follow-up if model hashes are published.

## Community Feedback

Issue #858 has 2 comments including the maintainer confirming the bug and suggesting the fix direction.

## Testing

**Automated:**
- 1 new unit test (`test_extraction_failure_cleans_up_partial_file`) verifying that a corrupt `.partial` is deleted after extraction failure and the `.extracting` directory is cleaned up
- All 4 model manager tests pass: `cargo test --lib managers::model::tests`

**Manual (macOS, M4 MacBook Air, dev build via `bun tauri dev`):**

1. **Corrupt archive cleanup (core bug):**
   - Deleted extracted Moonshine V2 Tiny model directory
   - Created a 31MB corrupt `.partial` file (random bytes, not a valid tar.gz)
   - Restarted app, model showed as not downloaded
   - Clicked Download, app resumed from EOF, attempted extraction, extraction failed
   - `.partial` was deleted, backend state reset, UI cleared
   - Clicked Download again, fresh download started from scratch, completed successfully

2. **Resume regression (must not break):**
   - Started downloading Moonshine V2 Tiny, cancelled mid-download at ~20MB
   - Restarted app, `.partial` preserved at 20MB
   - Clicked Download, resumed from byte 20159808 (not from 0%), completed successfully

**Backend log evidence:**
```
[INFO] Resuming download of model moonshine-tiny-streaming-en from byte 32505856
[INFO] Extracting archive for directory-based model: moonshine-tiny-streaming-en
[INFO] Starting fresh download of model moonshine-tiny-streaming-en
[INFO] Successfully extracted archive for model: moonshine-tiny-streaming-en
```

N/A — download/extraction behavior, not UI changes.

## AI Assistance

**If AI was used:**

- Tools used: Claude Code
- How extensively: Claude Code helped with implementation, code review, and test writing. The fix approach (scoped file handles for Windows compat, separate cleanup paths for corrupt archives vs environmental failures) and manual testing were done collaboratively.

### DIFF
--- a/src-tauri/src/managers/model.rs
+++ b/src-tauri/src/managers/model.rs
@@ -957,7 +957,7 @@ impl ModelManager {
             // transient environment issue (permissions, disk space), not a corrupt
             // archive, so we preserve .partial for retry.
             if let Err(e) = fs::create_dir_all(&temp_extract_dir) {
-                let error_msg = format!("Failed to create extraction directory: {}", e);
+                let error_msg = format!("Failed to create extraction directory: {e}");
                 {
                     let mut extracting = self.extracting_models.lock().unwrap();
                     extracting.remove(model_id);
@@ -990,7 +990,7 @@ impl ModelManager {
                     Ok(f) => f,
                     Err(e) => {
                         // File::open failure is environmental, preserve .partial
-                        let error_msg = format!("Failed to open archive: {}", e);
+                        let error_msg = format!("Failed to open archive: {e}");
                         let _ = fs::remove_dir_all(&temp_extract_dir);
                         {
                             let mut extracting = self.extracting_models.lock().unwrap();


cjpais:
I'm not convinced this is the way forward. If you as a human want to review the code and tell me I'm wrong let me know.

I can easily (trivially) compute the hashes. This to me seems like a band-aid solution which just adds more code and possibility for faults. The code is already a giant mess in this department, and I think it's almost better to just fundamentally rethink all of it, or rip it out.

sheeki03:
Gotcha...I avoided checksums, but I can just download the files from blob.handy.computer and compute them myself.
That would catch corruption before extraction even starts, which is simpler overall

cjpais:
Yeah right now I think maybe going the simpler way might be best. 

Certainly there's a way to handle partial downloads, I'm just a bit wondering how much it matters to handle this. More or less I think it would mean revisiting this code overall and seeing if there's a simpler solution. It seems like things are kind of just in a busted state and I would rather make a proper fix, or move to checksums.

Most of the models are a few hundred MB and not multi gig files

VirenMohindra (commented):
the fix is directionally correct — deleting `.partial` on extraction failure is the right call. left a few notes inline.

src-tauri/src/managers/model.rs
```diff
@@ -953,33 +953,104 @@ impl ModelManager {
...
+                    }
+                }
+                {
+                    let mut flags = self.cancel_flags.lock().unwrap();
+                    flags.remove(model_id);
+                }
+                let _ = self.app_handle.emit(
+                    "model-extraction-failed",
+                    &serde_json::json!({
+                        "model_id": model_id,
```
VirenMohindra, line 978: this cleanup block (unlock extracting, unlock models, unlock cancel_flags, emit event, return error) is repeated 3 times in this PR — here, in the `create_dir_all` failure path, and in the `File::open` failure path. that's ~25 lines × 3. a helper like `fn fail_extraction(&self, model_id: &str, error_msg: &str, delete_partial: bool)` would cut this to ~10 lines total and make the intent clearer.

src-tauri/src/managers/model.rs
```diff
@@ -953,33 +953,104 @@ impl ModelManager {
...
                 {
                     let mut extracting = self.extracting_models.lock().unwrap();
                     extracting.remove(model_id);
                 }
+                {
+                    let mut models = self.available_models.lock().unwrap();
+                    if let Some(model) = models.get_mut(model_id) {
+                        model.is_downloading = false;
+                        model.partial_size = 0;
+                    }
```
VirenMohindra, line 1039: this is the actual one-line fix for #858 — nice. on current main, this line is missing and the `.partial` file survives extraction failure, causing the infinite loop. everything else in the PR is hardening around this.

src-tauri/src/managers/model.rs
```diff
@@ -953,33 +953,104 @@ impl ModelManager {
...
-            let mut archive = Archive::new(tar);
+            // Unpack the archive in a scoped block so file handles are dropped
+            // before cleanup (required for Windows). If unpack fails the archive
+            // is corrupt — delete .partial so the next attempt downloads fresh.
+            let extract_result = {
+                let tar_gz = match File::open(&partial_path) {
+                    Ok(f) => f,
+                    Err(e) => {
+                        // File::open failure is environmental, preserve .partial
+                        let error_msg = format!("Failed to open archive: {e}");
```
VirenMohindra, line 993: good catch scoping the file handles for windows compat. worth keeping even in a slimmer version of this PR.

src-tauri/src/managers/model.rs
```diff
@@ -953,33 +953,104 @@ impl ModelManager {
...
+                        let _ = fs::remove_dir_all(&temp_extract_dir);
+                        {
+                            let mut extracting = self.extracting_models.lock().unwrap();
+                            extracting.remove(model_id);
+                        }
```
VirenMohindra, line 998: `File::open` failing on a `.partial` that we know exists is unusual — likely permissions or the file was deleted between the existence check and the open. preserving `.partial` here is fine but this is a very rare edge case. might not be worth 25 lines of dedicated handling vs just falling through to the same "delete and retry" path.

cjpais:
also @tanshkoul if you want to help out, testing this pr, or helping to improve/review/etc would be very helpful

cjpais:
closing in favor of #1095

PR #1024, opened 2026-03-12
Save recordings before transcription

ferologics:
## Summary

This PR fixes the data-loss path where a recording could disappear if transcription failed.

### Save recordings before transcription
- persist the WAV immediately after recording stops
- create/update history entries for `pending`, `completed`, and `failed` states
- preserve failed/pending entries in History instead of dropping them on cleanup
- store `transcription_error`
- allow retrying transcription from History using the saved WAV
- keep tray copy behavior limited to completed entries
- simplify History status copy so the badge/player/actions carry most of the state

## Testing
- `bun run build`
- `cargo check -j 1`
- manual verification of failed/pending History UI and retry flow using saved WAVs

### DIFF
--- a/src-tauri/src/actions.rs
+++ b/src-tauri/src/actions.rs
@@ -505,8 +505,8 @@ impl ShortcutAction for TranscribeAction {
                     let wav_path = hm.recordings_dir().join(&file_name);
                     let wav_path_for_verify = wav_path.clone();
                     let samples_for_wav = samples.clone();
-                    let wav_handle = tauri::async_runtime::spawn(async move {
-                        crate::audio_toolkit::save_wav_file(&wav_path, &samples_for_wav).await
+                    let wav_handle = tauri::async_runtime::spawn_blocking(move || {
+                        crate::audio_toolkit::save_wav_file(&wav_path, &samples_for_wav)
                     });
 
                     // Transcribe concurrently with WAV save
--- a/src-tauri/src/audio_toolkit/audio/utils.rs
+++ b/src-tauri/src/audio_toolkit/audio/utils.rs
@@ -28,7 +28,7 @@ pub fn verify_wav_file<P: AsRef<Path>>(file_path: P, expected_samples: usize) ->
 }
 
 /// Save audio samples as a WAV file
-pub async fn save_wav_file<P: AsRef<Path>>(file_path: P, samples: &[f32]) -> Result<()> {
+pub fn save_wav_file<P: AsRef<Path>>(file_path: P, samples: &[f32]) -> Result<()> {
     let spec = WavSpec {
         channels: 1,
         sample_rate: 16000,


cjpais:
Let's not do the ORT error fix here. 

We should be adding an option to transcribe-rs to do chunking. This is code which should be broadly applicable not just for Handy

It will also keep this PR more focused

ferologics:
Dropped the ORT split-retry commit from this PR to keep it focused on the durability/history fix. I kept local notes and a preserved branch for a follow-up chunking/chunked-transcription pass in transcribe-rs.

src/bindings.ts
```diff
@@ -745,6 +745,14 @@ async deleteHistoryEntry(id: number) : Promise<Result<null, string>> {
     else return { status: "error", error: e  as any };
 }
 },
+async retryHistoryEntryTranscription(id: number) : Promise<Result<null, string>> {
+    try {
+    return { status: "ok", data: await TAURI_INVOKE("retry_history_entry_transcription", { id }) };
```
VirenMohindra, line 758: needs a rerun of codegen + bun run format to fix

src-tauri/src/actions.rs
```diff
@@ -449,104 +495,93 @@ impl ShortcutAction for TranscribeAction {
...
                                 }
-                                // Hide the overlay after transcription is complete
-                                utils::hide_recording_overlay(&ah_clone);
-                                change_tray_icon(&ah_clone, TrayIconState::Idle);
-                            })
-                            .unwrap_or_else(|e| {
-                                error!("Failed to run paste on main thread: {:?}", e);
+                            }
+
+                            if processed.final_text.is_empty() {
```
VirenMohindra, line 568: when `processed.final_text.is_empty()`, the code silently hides the overlay and does nothing else but `complete_transcription()` was **already called** with an empty string, leaving the entry as `status=Completed`, `transcription_text=""`. In the UI this shows as a successful entry with no text and users can't distinguish "completed but empty" from "failed." we should either call `fail_transcription()` with a descriptive error or skip `complete_transcription()` entirely for empty results

src-tauri/src/managers/history.rs
```diff
@@ -317,23 +516,24 @@ impl HistoryManager {
     ) -> Result<()> {
         let conn = self.get_connection()?;
 
-        // Calculate cutoff timestamp (current time minus retention period)
         let now = Utc::now().timestamp();
         let cutoff_timestamp = match retention_period {
-            crate::settings::RecordingRetentionPeriod::Days3 => now - (3 * 24 * 60 * 60), // 3 days in seconds
-            crate::settings::RecordingRetentionPeriod::Weeks2 => now - (2 * 7 * 24 * 60 * 60), // 2 weeks in seconds
-            crate::settings::RecordingRetentionPeriod::Months3 => now - (3 * 30 * 24 * 60 * 60), // 3 months in seconds (approximate)
```
VirenMohindra, line 354: nit: would it be possible to avoid blowing this diff up and avoid removing comments? would be easier to review

cjpais:
I think I have some inflight changes for this PR

src-tauri/src/actions.rs
```diff
@@ -449,104 +494,127 @@ impl ShortcutAction for TranscribeAction {
...
-                            {
-                                final_text = converted_text;
+                if samples.is_empty() {
+                    debug!("Recording produced no audio samples; skipping persistence");
+                    utils::hide_recording_overlay(&ah);
+                    change_tray_icon(&ah, TrayIconState::Idle);
+                } else {
+                    // Save WAV concurrently with transcription
+                    let sample_count = samples.len();
+                    let file_name = format!("handy-{}.wav", chrono::Utc::now().timestamp());
```
VirenMohindra, line 504: lets use a uuid here instead. timestamp is a second class citizen and two recordings within the same second will overwrite each other

or even `chrono::Utc::now().timestamp_nanos_opt().unwrap_or_else(|| chrono::Utc::now().timestamp_micros())` but il lean towards uuids

src-tauri/src/actions.rs
```diff
@@ -449,104 +494,127 @@ impl ShortcutAction for TranscribeAction {
...
-                                        .find(|p| &p.id == prompt_id)
-                                    {
-                                        post_process_prompt = Some(prompt.prompt.clone());
-                                    }
+                            let processed =
+                                process_transcription_output(&ah, &transcription, post_process)
+                                    .await;
+
+                            // Save to history if WAV was saved and transcription is non-empty
+                            if wav_saved && !transcription.is_empty() {
```
VirenMohindra: orphaned wav files when there is silent audio and no transcription and then `cleanup_old_entries` won't touch it 

the error path below this correctly writes an empty DB row for retry and the empty-transcription path should do the same

src-tauri/src/actions.rs
```diff
@@ -449,104 +494,127 @@ impl ShortcutAction for TranscribeAction {
...
+
+                    // Transcribe concurrently with WAV save
+                    let transcription_time = Instant::now();
+                    let transcription_result = tm.transcribe(samples);
+
+                    // Await WAV save and verify
+                    let wav_saved = match wav_handle.await {
+                        Ok(Ok(())) => {
+                            // Verify: 44-byte header + num_samples * 2 bytes (16-bit)
+                            let expected_size = 44 + (sample_count * 2) as u64;
```
VirenMohindra: seems a bit fragile, rereading the header with `WavReader::open` and comparing sample counts is prob more robust

src-tauri/src/commands/history.rs
```diff
@@ -53,6 +57,49 @@ pub async fn delete_history_entry(
...
+    let samples = crate::audio_toolkit::read_wav_samples(&audio_path)
+        .map_err(|e| format!("Failed to load audio: {}", e))?;
+
+    if samples.is_empty() {
+        return Err("Recording has no audio samples".to_string());
+    }
+
+    transcription_manager.initiate_model_load();
+
+    match transcription_manager.transcribe(samples) {
```
VirenMohindra: lets wrap this with `tauri::async_runtime::spawn_blocking` if possible

this can run for 5 to 30s depending on model and hardware. calling it directly inside an async fn would block the thread for the entire time

src-tauri/src/commands/history.rs
```diff
@@ -53,6 +57,49 @@ pub async fn delete_history_entry(
...
+            let processed =
+                process_transcription_output(&app, &transcription, entry.post_process_requested)
+                    .await;
+            history_manager
+                .update_transcription(
+                    id,
+                    transcription,
+                    processed.post_processed_text,
+                    processed.post_process_prompt,
+                )
```
VirenMohindra: UI has no way to distinguish "retry succeeded with silence" from the original failed state and the user will retry again with the same result indefinitely

we should prob return an error when `transcription.is_empty()` so the fe can show a meaningful state

VirenMohindra (commented):
some misc UI fixes ordered in terms of prio~

1. the retry button shows on completed entries and should only appear when `!hasTranscription`
2. content area shows "Transcription failed" during retry and should instead swap to a loading state
3. delete btn stays enabled during retry, and will be a possible race condition if you delete while retranscribing

i think we should do a proper UI pass before this lands

cjpais:
I think retry should be for everything, mainly if you want to re-transcribe with another model or you picked the wrong language (happens to me all the time)

Otherwise other 2 I agree

VirenMohindra:
also please add screenshots to the PR

cjpais:
@VirenMohindra let me know what you think of the current set of changes. Not sure if the original author will jump in, but I think things are doing okay right now. Would love a second opinion

cjpais:
Screenshot 2026-03-21 at 1 18 12 PM

ferologics:
thx for getting this across the finish line @cjpais & @VirenMohindra !! 
now we just need to recover from failures better :P

PR #1021, opened 2026-03-12
fix(nix): replace manual hash management with bun2nix

xilec:
**Please confirm you have done the following:**

  - [x] I have searched [existing issues](https://github.com/cjpais/Handy/issues) and [pull
  requests](https://github.com/cjpais/Handy/pulls) (including closed ones) to ensure this isn't a duplicate
  - [x] I have read [CONTRIBUTING.md](https://github.com/cjpais/Handy/blob/main/CONTRIBUTING.md)

  **If this is a feature or change that was previously closed/rejected:**

  - [ ] I have explained in the description below why this should be reconsidered
  - [ ] I have gathered community feedback (link to discussion below)

  ## Human Written Description

  The Nix build (`nix build .#handy`) keeps breaking because the fixed-output derivation                                           
  hash for bun dependencies has to be manually updated every time `bun.lock` changes.
  The same problem exists for cargo git dependency hashes. This makes maintaining the Nix                                          
  package tedious and results in frequent hash-update PRs.

  This PR eliminates all manual hash management by switching to bun2nix (per-package
  fetchurl expressions) for bun deps and `allowBuiltinFetchGit` for cargo git deps.
  A postinstall hook auto-regenerates the Nix files whenever `bun.lock` changes, so
  non-Nix developers won't accidentally break the Nix build.

  Normally, when you update TS dependencies with `bun add ...` or
  `bun update ...`, Nix dependency files are auto-updated via the postinstall hook.
  A developer just needs to commit the updated files, same as with `Cargo.lock`.

  If `bun.lock` is updated manually (without bun), the CI job `nix build check` will catch it
  and show a red status with an error message explaining that you should run
  `bun scripts/check-nix-deps.ts` to update Nix dependencies.

  I believe this should eliminate 99% of problems with Nix dependencies.
  After this PR, manual fixes are only expected when `ferrous-opencc` is updated
  (it requires a special patch currently), or if you decide to publish the app in
  nixpkgs (which prohibits `allowBuiltinFetchGit` — but for standalone flake packages
  it's fine). In all other cases, Nix dependencies should be updated automatically
  via the hook or by running the script in the same PR where the bun dependency changed,
  without requiring a separate manual PR.
    
  **Note:** The `nix build check` CI job is currently set to `continue-on-error: false`,
  meaning it will block PRs with outdated Nix files. The fix is always straightforward
  (just run `bun install` or `bun scripts/check-nix-deps.ts`), but if you'd prefer
[...]

PR #1025, opened 2026-03-12
feat(nix): NixOS integration — ALSA_PLUGIN_DIR + nixosModule

zitongcharliedeng:
## Summary

Two changes to make Handy work out-of-the-box on NixOS:

**1. ALSA_PLUGIN_DIR in package wrapper** (`flake.nix`, +1 line)

On NixOS with PipeWire, CPAL opens the ALSA "default" device but gets 0 audio samples because PipeWire's ALSA plugin isn't on the default search path. Setting `ALSA_PLUGIN_DIR` in the gapps wrapper fixes this.

**2. nixosModule for system-level config** (`nix/module.nix`, new file)

Adds a `nixosModules.default` output so NixOS users can enable Handy with:

```nix
inputs.handy.url = "github:cjpais/Handy";

# In your NixOS config:
{ programs.handy.enable = true; }
```

The module handles system config that the package wrapper cannot:
- udev rule for `/dev/uinput` (rdev `grab()` needs it for virtual input devices)
- Adds the Handy package to `systemPackages`

Users still need to add themselves to the `input` group for evdev hotkey access.

## Test plan

- [x] `nix flake check --no-build` passes
- [x] Build and run on NixOS with PipeWire — verify audio capture works
- [x] Verify `programs.handy.enable = true` sets up udev rule correctly

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -174,6 +174,7 @@
             preFixup = ''
               gappsWrapperArgs+=(
                 --set WEBKIT_DISABLE_DMABUF_RENDERER 1
+                --set ALSA_PLUGIN_DIR "${pkgs.pipewire}/lib/alsa-lib:${pkgs.alsa-plugins}/lib/alsa-lib"
                 --prefix LD_LIBRARY_PATH : "${
                   lib.makeLibraryPath [
                     pkgs.vulkan-loader
@@ -195,6 +196,22 @@
           default = self.packages.${system}.handy;
         }
       );
+
+      # NixOS module for system-level integration (udev, input group)
+      nixosModules.default =
+        { lib, pkgs, ... }:
+        {
+          imports = [ ./nix/module.nix ];
+          programs.handy.package = lib.mkDefault self.packages.${pkgs.stdenv.hostPlatform.system}.handy;
+        };
+
+      # Home-manager module for per-user service
+      homeManagerModules.default =
+        { lib, pkgs, ... }:
+        {
+          imports = [ ./nix/hm-module.nix ];
+          services.handy.package = lib.mkDefault self.packages.${pkgs.stdenv.hostPlatform.system}.handy;
+        };
 
       # Development shell for building from source
       devShells = forAllSystems (
--- a/nix/hm-module.nix
+++ b/nix/hm-module.nix
@@ -1,0 +1,41 @@
+# Home-manager module for Handy speech-to-text
+#
+# Provides a systemd user service for autostart.
+# Usage: imports = [ handy.homeManagerModules.default ];
+#        services.handy.enable = true;
+{
+  config,
+  lib,
+  pkgs,
+  ...
+}:
+let
+  cfg = config.services.handy;
+in
+{
+  options.services.handy = {
+    enable = lib.mkEnableOption "Handy speech-to-text user service";
+
+    package = lib.mkOption {
+      type = lib.types.package;
+      defaultText = lib.literalExpression "handy.packages.\${system}.handy";
+      description = "The Handy package to use.";
+    };
+  };
+
+  config = lib.mkIf cfg.enable {
+    systemd.user.services.handy = {
+      Unit = {
+        Description = "Handy speech-to-text";
+        After = [ "graphical-session.target" ];
+        PartOf = [ "graphical-session.target" ];
+      };
+      Service = {
+        ExecStart = "${cfg.package}/bin/handy";
+        Restart = "on-failure";
+        RestartSec = 5;
+      };
+      Install.WantedBy = [ "graphical-session.target" ];
+    };
+  };
+}
--- a/nix/module.nix
+++ b/nix/module.nix
@@ -1,0 +1,47 @@
+# NixOS module for Handy speech-to-text
+#
+# Handles system-level configuration that the package wrapper cannot:
+#   - udev rule for /dev/uinput (rdev grab() needs it for virtual input)
+#
+# Note: users must add themselves to the "input" group for evdev hotkey access.
+#
+# Usage in your flake:
+#
+#   inputs.handy.url = "github:cjpais/Handy";
+#
+#   nixosConfigurations.myhost = nixpkgs.lib.nixosSystem {
+#     modules = [
+#       handy.nixosModules.default
+#       { programs.handy.enable = true; }
+#     ];
+#   };
+{
+  config,
+  lib,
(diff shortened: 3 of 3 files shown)


zitongcharliedeng:
## Note: Global hotkeys on Wayland

~~Previously noted that global shortcuts don't work on Wayland via Tauri/rdev's X11 path.~~

**Update:** HandyKeys (evdev backend) works natively on Wayland — no xdotool workaround needed. Set `keyboard_implementation: "handy_keys"` in settings and it captures hotkeys directly via `/dev/input/event*`, bypassing X11 entirely.

cjpais:
Just a heads up I don't believe handy keys works generally on Linux, but you can let me know. It's certainly not going to be the default without more testing

zitongcharliedeng:
Seems to work fine on my Niri on Wayland.

https://github.com/user-attachments/assets/defd2be6-5a60-4c54-8357-86f2c2440706

zitongcharliedeng:
Updated to put the key fix in handey-keys repo.

zitongcharliedeng:
I presume there is some auto bump in the nix flake for it to propagate

cjpais:
> Updated to put the key fix in handey-keys repo.

what is this fix?

zitongcharliedeng:
Just adding the missing Insert Key and preventing keyboard hotplugs from
disconnecting listeners.
https://github.com/handy-computer/handy-keys/pull/12

cjpais:
@y0usaf @xilec @CaptainSpof @kakapt please confirm this is a good change to have

@zitongcharliedeng handy keys#12 should be pulled in tomorrow, just waiting on comment from you there to get more background

kakapt:
> @y0usaf @xilec @CaptainSpof @kakapt please confirm this is a good change to have
> 
> @zitongcharliedeng handy keys#12 should be pulled in tomorrow, just waiting on comment from you there to get more background

I think setting `ALSA_PLUGIN_DIR` is redundant because by default if we have pipewire enabled then cpal will able to detect ALSA. The issue is the version mismatch between cpal and ALSA. I've played around this in pr [787](https://github.com/cjpais/Handy/pull/787) and also discussed with @cjpais on Discord earlier

kakapt:
For nixosModule, I think it is a good change but can we consider providing home manager module too?

flake.nix
```diff
@@ -51,7 +51,7 @@
 
             outputHashAlgo = "sha256";
             outputHashMode = "recursive";
-            outputHash = "sha256-84Aw9E2+fEZT+lIb9k1bodessoex+YFr0im2GMVAPnw=";
+            outputHash = "sha256-x8SP6k7eZwlPhOuQ876aVI+hFxRUiYRWrdCDdQh7jY4=";
```
xilec: @zitongcharliedeng PR #1021 was just merged, which replaced the manual `outputHash` approach with bun2nix. Your `flake.nix` is based on the old version, so you'll need to rebase on current `main`.

flake.nix
```diff
@@ -181,6 +181,7 @@
                     pkgs.onnxruntime
                   ]
                 }"
+                --set ALSA_PLUGIN_DIR "${pkgs.pipewire}/lib/alsa-lib"
```
xilec: Consider supporting both PipeWire and PulseAudio by including both plugin paths:

```suggestion
                --set ALSA_PLUGIN_DIR "${pkgs.pipewire}/lib/alsa-lib:${pkgs.alsa-plugins}/lib/alsa-lib"
```

ALSA searches all paths and ignores non-existent ones, so this covers PulseAudio users as well without any conditional logic.

zitongcharliedeng:
Updated the PR based on feedback:

**@kakapt re ALSA_PLUGIN_DIR** — On my NixOS setup, without ALSA_PLUGIN_DIR set, CPAL gets 0 audio samples. Setting it to the pipewire store path fixed it, consistent with what nixpkgs does for alsa-utils (#432786). But I haven't tested with your PR #787 changes — is the ALSA version mismatch fix sufficient to make CPAL find PipeWire without ALSA_PLUGIN_DIR? Happy to drop it if so.

**@kakapt re home-manager module** — Added `homeManagerModules.default` with systemd service, `.asoundrc` PipeWire routing, and first-install settings seed. One caveat: Handy's Tauri `app.store()` overrides `settings_store.json` on startup, so the module can only seed settings on first install — after that Handy's UI is the source of truth.

kakapt:
> Updated the PR based on feedback:
> 
> **@kakapt re ALSA_PLUGIN_DIR** — On my NixOS setup, without ALSA_PLUGIN_DIR set, CPAL gets 0 audio samples. Setting it to the pipewire store path fixed it, consistent with what nixpkgs does for alsa-utils (#432786). But I haven't tested with your PR #787 changes — is the ALSA version mismatch fix sufficient to make CPAL find PipeWire without ALSA_PLUGIN_DIR? Happy to drop it if so.

Yeah, I had the same issue before. I also tried adding `ALSA_PLUGIN_DIR`  but finally figured out that it is actually the cpal error.
@zitongcharliedeng You can see some of my debug logs on Handy discord: https://discord.com/channels/1391636091853078608/1471037063171281007

kakapt:
> Updated the PR based on feedback:
> 
> **@kakapt re home-manager module** — Added `homeManagerModules.default` with systemd service, `.asoundrc` PipeWire routing, and first-install settings seed. One caveat: Handy's Tauri `app.store()` overrides `settings_store.json` on startup, so the module can only seed settings on first install — after that Handy's UI is the source of truth.

Thanks for it! I'll test it soon and let you know!

xilec:
LGTM

cjpais:
everyone here please give confirmation and I will pull it in when there is consensus!

zitongcharliedeng:
Lgtm

kakapt:
@cjpais I notice that the doc does not mention the udev rules for `uinput` for general linux user

cjpais:
@kakapt if you want to submit a PR to them I would gladly pull in

cjpais:
@xilec will this cause a merge conflict with the work we are doing in #1041, and if I pull this in, can you fix that conflict

xilec:
@cjpais The changes don't overlap — #1025 touches `preFixup` and adds modules, while #1041 adds the overlay and an env var. Git should auto-merge without conflicts.

cjpais:
great thanks, just gonna wait for the build to finish here and will pull in

Issue #192, opened 2025-10-09
[Feature Request] Automatically pause/resume media playback when recording starts/ends


Maicon-Moreira:
Hi,  

First of all, thanks a lot for Handy — it’s an amazing app and I use it all the time on macOS.  

There’s one feature that I think would make the experience even better: when I start a recording (using the keyboard shortcut), Handy could automatically pause whatever media is currently playing, and then resume it once the recording ends. This could apply to music in Spotify or a video playing in a browser, for example. It would help avoid having background sound picked up during the recording and just make the process smoother overall.

Example flow:  
- Press shortcut to start Handy recording → music or video pauses  
- Recording ends → playback resumes

Since Handy runs on multiple platforms, I imagine this could be handled differently depending on the OS. For example, maybe by sending a system media play/pause event or using platform-specific media APIs. I don’t know what’s best technically, but the idea would just be to have this simple automation that pauses and resumes playback around the recording session.

As a stretch idea, it would also be cool if Handy could enable the AirPods transparency mode when recording starts, so the user can hear their own voice better. I don’t think this is possible through public APIs, but just mentioning it in case it ever becomes doable.

Thanks again for all your work on Handy — it’s a really well-made and useful app.

cjpais:
Thanks! I'm happy to add this in if someone comes and implements it. Seems like a nice QOL improvement. I probably won't have time for a while to attempt this

Maicon-Moreira:
Sounds good! I might take a look at it when I have some free time — seems like it could be a fun feature to try.

therealpygon:
@cjpais  Given the goal of keeping handy simple, it might be a better option for people to use the media play/pause key on their keyboard if they are listening to music, or they could simply make a macro to pause/unpause music when they activate handy instead of handy trying to branch out into controlling music playback. This could be handled rather easily with a simple autohotkey macro.

Edit: I say this because the alternative gets into handy trying to figure out whether or not music is playing (on each of the various platforms) and reliably muting or pausing it, all while avoiding all the bugs for users who weren't listening to music. It's definitely doable, and if you plan to chase it might suggestion is that the keyboard media control play/pause might not be a bad way to do it. Alternatively, you might add an optional toggle to do something like mute all sound output while recording rather than trying to play/pause peoples music.

cjpais:
@therealpygon I agree with you, I'm going to close this issue for now. I will review a PR if it solves this in a nice way.

grishy:
[...]

PR #1028, opened 2026-03-13
Add pause-while-recording across macOS, Windows, and Linux (tested on Ubuntu)

### LINKED ISSUE #998: [BUG] macOS 26.3 - 'Mute While Recording' doesn't work when using USB Audio Interface. Can handy have 'Pause While Recording' as an alternative?
'Mute While Recording' doesn't work when using USB Audio Interface.

I believe this is due to how macOS doesn't natively support a bunch of USB Audio Interfaces in terms of the ability to change their volume level using keyboard or software, volume changes must be done with them using the hardware knob.

I'm not sure how handy tries to achieve Mute While Recording but the output is not actually muted when I use any USB Audio Interface with my Mac.

## Alternative suggestion
An alternative solution to this entire issue is a different suggestion - handy could have an option to automatically pause any media playback while recording instead of just muting. This way any playback of any app on any OS just pauses instead of relying on trying to mute playback from different combinations of hardware and software. I know https://superwhisper.com/ does this and it's very handy. Playback also auto-resumes after recording too.

## System Information

**App Version:**
0.7.10

**Operating System:**
macOS 26.3

**CPU:**
M1 Pro

**GPU:**
M1 Pro

## Logs
Log says `[DEBUG] Mute applied` but it never actually muted.
[handy.log](https://github.com/user-attachments/files/25894290/handy.log)

apotenza92:
## Before Submitting This PR

## Human Written Description

I originally opened #998 after having an issue on macOS where mute while recording wasn't working correctly with my hardware setup (USB audio interface), so I decided to just try making pause while recording as an additional option that I prefer.

## Related Issues/Discussions

Fixes #998
Related to #192

This revisits the older pause/resume media request from #192, but the scope here is intentionally framed as a bug-fix / hardening pass rather than a net-new feature pitch. This PR makes pause-while-recording behave consistently across macOS, Windows, and Linux, improves lifecycle safety, and aligns the settings/UI/backend behavior.

## Community Feedback

The original motivation came from #998. I did not open a separate Discussion thread for this change, but I manually built and tested the resulting behavior on macOS, Windows, and Ubuntu to verify the feature works across all three supported desktop platforms. I primarily run macOS but used Parallels to test in Windows and Ubuntu VMs.

### Automated / local checks
- `cargo test --manifest-path src-tauri/Cargo.toml media_control -- --nocapture`
- `bun run build`
- `bun run format:check`
- `bun run lint`

### Manual cross-platform validation
- **macOS**
  - Built `Handy.app` and DMG locally
  - Launched the app bundle successfully
  - Verified pause-while-recording pauses active media playback
  - Verified resume-after-recording restores playback when enabled
- **Windows 11 ARM VM**
  - Built native `handy.exe`
  - Launched and verified the app runs in the VM
  - Verified pause/resume behavior works during recording
- **Ubuntu 24.04 ARM64 VM**
  - Built native `handy`
  - Launched and verified the app runs in the VM
  - Verified pause/resume behavior works during recording

### What changed
- Added platform-specific media control backends for macOS, Windows, and Linux
- Added lifecycle hardening to avoid duplicate pause/resume and stale in-flight pause results
- Updated recording start/stop ordering so pause happens before Handy's start sound and resume happens after the stop sound
- Synced UI/store/backend behavior for pause/resume settings
- Added unit coverage for media-control lifecycle cases

### Notes
- `cargo clippy --all-targets --all-features -- -D warnings` still reports existing repo-wide warnings unrelated to this PR, so I did not fold a broad clippy cleanup into this change

None for now.

## AI Assistance

**If AI was used:**

- Tools used: pi coding agent / AI coding assistant
- How extensively: used for repository discovery, implementation assistance, build debugging, cross-platform VM build/test setup, and PR drafting support


cjpais:
this does not compile on macos x86

https://github.com/cjpais/Handy/actions/runs/23041353318/job/66920158212

apotenza92:
Reproduced locally on macOS x86_64 target and pushed fix. Issue was MediaRemote bridge being compiled for the host arch instead of the target arch

apotenza92:
I pushed a cleanup / update based on some issues that came up during my own local testing:
 - fixed macOS resume-after-recording behaviour
 - fixed the macOS case where playback could start after recording even if nothing
 was playing initially
 - restored the macOS whisper bindings override so x86_64-apple-darwin builds
 compile again
 - cleaned out unrelated local/debug churn from the branch

 Re-tested locally on macOS:
 - playback case: pauses during recording and resumes after
 - no-media case: does not spuriously start playback
 - cargo check -q
 - cargo check —target x86_64-apple-darwin —no-default-features -q

apotenza92:
Hi would really appreciate this being reviewed again! Thanks for all the work.

apotenza92:
Hi. Is there anything else I can do to help get this merged?

woerndl:
I have the same issue. Since muting does not work when you're using a USB interface, this would really be a cool workaround.

cjpais:
Please build and test it! More confirmation I get the faster it can be pulled in

woerndl:
I couldn't build it at first because of https://github.com/cjpais/Handy/issues/1448 but now I'm running the PR version.

Compiled fine on M1. The pause trigger has a slight delay of roughly 500 ms, but in general it works for me. This is probably a reasonable fix and a smart addition to the mute. It solves my problem that I can't listen to music while working with Handy because my speakers are connected via USB. Actually, it might be even nicer than just being able to mute, since things like songs, podcasts, etc. continue at the same position when you transcribe for longer.

apotenza92:
Updated this PR branch to the newer Handy v0.8.3/stable-based implementation.

Summary:
- macOS pause-while-recording is faster and safer: removed the old startup delay, uses MediaRemote now-playing playback rate to confirm active playback, then sends the HID play/pause media key for reliable pause/resume.
- Fixed the no-media case where a stale playing signal could cause the media key to start stopped media.
- Removed the temporary latency experiment logging/env knobs before finalizing the branch.
- Windows GSMTC and Linux MPRIS behavior were live-tested and kept as-is because they were already fast and stable.

Validation:
- macOS Music: rebuilt app tested locally and working well.
- macOS no-media regression: added/performed an ignored live test confirming Handy does not start Music when nothing is playing.
- Windows 11 ARM + Edge media: live audible pause/resume confirmed.
- Ubuntu 24.04 ARM64 + Firefox MPRIS: 20/20 pause, 20/20 resume, 10/10 already-paused guard.
- `cargo test --lib`: passing.
- `cargo test --lib macos_live_pause_does_not_start_media_when_nothing_is_playing -- --ignored --nocapture`: passing.
- `cargo fmt --check` and `git diff --check`: passing.

newideas99:
Built and tested this branch today (c9c5a7a) since I have one of the USB interfaces from #998 where mute does nothing.

Setup: macOS 27.0 (M series), Native Instruments Traktor Kontrol S4 MK2 as the default output, Spotify playing locally.

Results, consistent across repeated cycles:
- Spotify paused about 0.6s after recording started
- playback resumed about 0.3s after recording stopped

One small suggestion from reading the code: if the Accessibility permission ever isn't granted, CGEventPost drops the key event silently and the log would still print "Paused media playback for recording". A log line that distinguishes "sent the key" from "saw the playback state change" could save someone debugging time down the road. Not a blocker, just an observability nit.

So this works on exactly the hardware class from #998, nice work. For what it's worth I have a complementary PR (#1503) that ducks audio to a configurable level instead of pausing, for non resumable stuff like live streams and calls. The two approaches can coexist.

apotenza92:
Hi, anything else I can do to get this incorporated? 😃

cjpais:
Right now no, I'm the bottleneck and have about 100 PRs to review and bug fixes are priority. This topic is particularly challenging as there's 10+ PRs which have submitted various solutions which are similar. I need to make a UX decision as well as see if it justifies the bugs which will invariably fall out of the changes. Whoever's changes get merged will likely need their further support

adrian-rpf:
I've just tested this PR (after updating and fixing conflicts) 
- Apple - Macbook M4 Pro - 24GB Memory - Nov 2024 - 
- Parakeet Unified EN 0.6B 

While it does add more settings to the `sound` section. 
image


This _pausing_ setup is also much quicker at pausing the audio or reducing the audio so you can start talking.

cjpais:
thanks for feedback @adrian-rpf

PR #1035, opened 2026-03-13
Using the whisper `initial_prompt` field for custom words

jmcampanini:
Whisper supports an `initial_prompt` parameter that biases the model's vocabulary during transcription. For Whisper models, we now join the user's custom words list into a comma-separated string and pass it as the initial_prompt. This guides the model to prefer those spellings during decoding rather than relying solely on fuzzy post-correction.

For non-Whisper engines (Parakeet, Moonshine, SenseVoice, GigaAM), the existing apply_custom_words post-correction continues to apply since those engines don't support prompt-based vocabulary biasing.

## Before Submitting This PR

## Human Written Description

From the discussion:

I've encountered an issue with custom vocabulary/custom words in Handy. The current bigram replacement mechanism can be overly aggressive - it sometimes replaces words that shouldn't be replaced. However, removing those entries from the custom words list means the desired words never appear at all.

I know that Whisper supports an "initial prompt" concept, which biases the model toward producing specific words without forcefully replacing output. While looking through the codebase, I noticed there appears to be a way to use custom words to generate an initial prompt, and I wanted to ask whether a pull request for this would be welcome.

## Related Issues/Discussions

Fixes #
Discussion: https://github.com/cjpais/Handy/discussions/1009

## Testing

No unit tests for this. I built and used this on my machine. The use case I was working on was "CLI" and "call" - that was fun to see in transcripts once I noticed it :)

## AI Assistance

**If AI was used:**

- Tools used: claude-code opus 4.6
- How extensively: I knew of the capability from prior research and came up with the idea of using the existing vocabulary as the prompt instead of replacement. I then used claude-code to see if the fix was small and implement it. I then reviewed it, but I admit I have not written rust before, so my human review may be insufficient.

### DIFF
--- a/src-tauri/src/managers/transcription.rs
+++ b/src-tauri/src/managers/transcription.rs
@@ -501,6 +501,11 @@ impl TranscriptionManager {
                             let params = WhisperInferenceParams {
                                 language: whisper_language,
                                 translate: settings.translate_to_english,
+                                initial_prompt: if settings.custom_words.is_empty() {
+                                    None
+                                } else {
+                                    Some(settings.custom_words.join(", "))
+                                },
                                 ..Default::default()
                             };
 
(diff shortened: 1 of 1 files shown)


cjpais:
This looks good to me. Thank you. Approved and merged.

Issue #96, opened 2025-08-27
[BUG] Changing the Handy shortcut from ctrl + space doesn't work.


hbenthow:
## Bug Description
Whenever I try to change the Handy shortcut from ctrl + space, it doesn't work. Some combinations like sheift + ctrl, immediately revert back to ctrl + space, while others like my preferred shortcut (scroll lock) disappear after restarting Handy (and don't work even before the restart, even though the GUI lists it as the shortcut).

## System Information
**App Version:**

**Operating System:**

**CPU:**

**GPU:**

cjpais:
Let me see if I can replicate on Windows. I'm gone for a few days but will try it. This issue has been reported before and I haven't been able to replicate for the fix

rarkins:
I just installed Handy 0.4.2 and couldn't even figure out how to define the shortcut. It wasn't until I came here and searched for Windows that I found that Control Space is the default. In fact it says that there's no shortcuts defined. However, I'm happy to say that Control Space does work. In fact, I typed out this issue using it.

<img width="1021" height="809" alt="Image" src="https://github.com/user-attachments/assets/a97c4a2e-0139-4f5f-80e6-bbb6134b0684" />

cjpais:
Oh thank you for this note, are you on Windows 10? I'll try to fix that bug with no shortcut defined, that's really not right

rarkins:
I am using Windows 11 Pro.

cjpais:
Thanks, this is a big issue, I'll try to replicate and fix.

therealpygon:
> Oh thank you for this note, are you on Windows 10? I'll try to fix that bug with no shortcut defined, that's really not right

I'm seeing the same thing ("No shortcuts configured") with no way to configure a shortcut. Windows 10 Pro

Edit: As an aside, I was able to get the transcription working using the control space shortcut. That portion actually does work really well, and I'm excited to keep using it.

cjpais:
Thanks for the additional report @therealpygon! Hopefully, within the next week, I'll be able to test Windows and fix this. I suspect that this is a bug with something, how I configured the settings, and I refactored it in the latest version and probably this is a bug as a result of that, so hopefully I'll be able to get that fixed at the very least.

wlwwt:
Similar problem I have on Apple Silicon. most text to speech apps on Apple use the Fn as hotkey. But I am not able to switch it to this one. It always gets back to the standard.

cjpais:
See #134.

[...]

PR #1039, opened 2026-03-14
upgrade to handy-keys 0.2.3

cjpais:


cjpais:
maybe help address #96

PR #1051, opened 2026-03-16
fix: auto-unload model after idle timeout to reduce memory

VirenMohindra:
## summary

handy holds the transcription model in memory indefinitely (default: `Never` unload). on a 16 GB machine, a parakeet model uses ~1 GB, even when the user hasn't transcribed in hours

this PR adds automatic model unloading after a configurable timeout, reducing idle memory from ~1.1 GB to ~80 MB

| before      | after |
| ----------- | ----------- |
| before      |  after       |

## changes

- **default timeout: `Never` -> `Min5`**: model auto-unloads after 5 minutes of inactivity
- **fix: watcher thread killed by `Drop` on clones**: `initiate_model_load()` clones `TranscriptionManager`, and when the clone drops, `Drop::drop` sets `shutdown_signal = true`, killing the watcher. fixed with an `Arc::strong_count` guard so only the last clone shuts down the watcher
- **fix: reset `last_activity` on model load**: without this, switching models after timeout elapsed would immediately unload the just-loaded model
- **unload logging upgraded to `info!()` level**: was `debug!()`, invisible at default log level
- **removed duplicate "unloaded" event**: `unload_model()` already emits it
- **error handling on unload**: `match` instead of `if let Ok` so failures are logged

## test results (macOS, parakeet v2, physical footprint = activity monitor)

| state | memory |
|-------|--------|
| baseline (no model) | 46 MB |
| model loaded | 1.1 GB |
| after auto-unload (5s timeout) | 80-88 MB |
| unload time | 33-44ms |

actual time to reach 88mb is ~30seconds

full lifecycle:
```
46 MB -> 1.1 GB -> 660 MB -> 330 MB -> 88 MB
(idle)  (loaded)  (unloading... OS reclaiming pages)
```

## test plan

- [x] start app, timeout=sec5, transcribe -> model loads, unloads after ~12s
- [x] transcribe again -> model reloads, unloads after ~7s
- [x] watcher survives both initiate_model_load clone/drop cycles
- [x] activity monitor confirms 80-88 MB after unload (not 700+ MB)
- [x] unload duration logged at info level

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,7 +1,7 @@
 {
   "name": "handy-app",
   "private": true,
-  "version": "0.7.10",
+  "version": "0.7.11",
   "type": "module",
   "scripts": {
     "dev": "vite",
--- a/src-tauri/Cargo.lock
+++ b/src-tauri/Cargo.lock
@@ -2416,7 +2416,7 @@ dependencies = [
 
 [[package]]
 name = "handy"
-version = "0.7.10"
+version = "0.7.11"
 dependencies = [
  "anyhow",
  "chrono",
--- a/src-tauri/Cargo.toml
+++ b/src-tauri/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "handy"
-version = "0.7.10"
+version = "0.7.11"
 description = "Handy"
 authors = ["cjpais"]
 edition = "2021"
--- a/src-tauri/tauri.conf.json
+++ b/src-tauri/tauri.conf.json
@@ -1,7 +1,7 @@
 {
   "$schema": "https://schema.tauri.app/config/2",
   "productName": "Handy",
-  "version": "0.7.10",
+  "version": "0.7.11",
   "identifier": "com.pais.handy",
   "build": {
     "beforeDevCommand": "bun run dev",


cjpais:
I was noticing on the latest build the timeout was actually not working at all. Have you experienced this?

At least in the case of changing models from the menubar.

VirenMohindra:
> I was noticing on the latest build the timeout was actually not working at all. Have you experienced this? 
> At least in the case of changing models from the menubar.

think i found the bug. `load_model()` wasn't updating `last_activity`, so when you switch models from the menubar, here's what happens~
1. you transcribe at T=0 -> `last_activity = T=0`
2. some time passes (longer than the timeout)
3. you switch models from the tray and `load_model()` loads the new model
4. **but last_activity is still T=0**
5. idle watcher checks on its next 10s tick: now - last_activity > timeout -> true -> immediately unloads the model you just loaded

_so the model loads, then gets unloaded on the very next watcher cycle. it looks like the timeout "doesn't work" but really the timer was just starting from the wrong point_

pushed a fix, `load_model()` now resets last_activity after **successfully loading**, so the idle timer starts fresh from the moment the model is ready. this applies to both tray model switches and the background `initiate_model_load()` path

cjpais:
the changes as they are don't work for me

I dont see prints in the console that model unloading is happening. I am also not seeing memory being freed, nor is the UI updating to show the model is unloaded.

I think the most critical paths to verify are:
1. start the app
2. set to never unload
3. trigger transcription start/stop
4. model doesnt unload
5. set to 5 sec 
  * after 5-10 sec should see print in console, and indicated in the UI that the model is unloaded (including the status bar)
6. set to never (model does not load and loads next time)
7. set to 2 minutes 
8. trigger transcription start/stop
9. set to 5 seconds 
10. model should unload

I think realistically we might need a small state machine or something... Or at least very clear logic what happens when transitioning between states. It may be worth making a diagram at least... Posting that here for verification and then having an LLM go implement and verify it ideally. would be great to have it in a testable unit.

VirenMohindra:
tested all 10 steps - everything works now. the main issue was a pre-existing bug in `Drop` that was killing the idle watcher thread before it could ever fire.

script here
[test-model-unload.sh](https://github.com/user-attachments/files/26042636/test-model-unload.sh)


**root cause:** `initiate_model_load()` clones `TranscriptionManager` and moves it into a spawned thread. when that thread finishes loading, the clone is dropped -> `Drop::drop` fires -> sets `shutdown_signal = true` -> watcher dies. so the watcher was alive for ~60s after startup and then permanently dead. no auto-unload ever happened

**fix:** added an `Arc::strong_count` guard in `Drop` - clones skip shutdown since other owners (tauri state, watcher thread) still exist. only the very last clone shuts down the watcher

**test results (physical footprint, matches activity monitor)~**

| state | memory |
|-------|--------|
| baseline (no model) | 46 MB |
| model loaded | 1.1 GB |
| after auto-unload (sec5) | 80-88 MB |
| unload time | 33-44ms |

full lifecycle:
```sh
46 MB -> 1.1 GB -> 660 MB -> 330 MB -> 88 MB
(idle)  (loaded)  (unloading... OS reclaiming pages)
```

two back-to-back transcriptions, both auto-unloaded, watcher survived the entire session~
```sh
04:17:23  Transcription #1
04:17:35  Model idle 12s > 5s -> unloaded (44ms)
04:18:58  Transcription #2
04:19:05  Model idle 7s > 5s -> unloaded (33ms)
```

also upgraded unload logging from `debug!()` to `info!()` so it's visible at default log level, and replaced `if let Ok` with `match` to log unload errors

src-tauri/src/managers/transcription.rs
```diff
@@ -771,7 +782,15 @@ pub fn get_available_accelerators() -> AvailableAccelerators {
...
-        debug!("Shutting down TranscriptionManager");
+        // Skip shutdown unless this is the very last clone. TranscriptionManager
+        // is cloned by initiate_model_load() and the watcher thread — those
+        // clones dropping must not kill the watcher. The watcher thread holds
+        // its own clone, so engine's strong_count is always >= 2 while the
+        // watcher is alive. When it reaches 1, only this instance remains
+        // and we can safely shut down.
+        if Arc::strong_count(&self.engine) > 1 {
+            return;
+        }
```
VirenMohindra, line 793: `Arc::strong_count` docs warn it shouldn't be used for synchronization and that the count can change between the check and the `shutdown_signal.store`. in practice this only runs during app teardown so it's _fine_ today, but adding this comment so future clone paths don't break this silently. alternatively, an explicit `AtomicUsize` refcount we control would be more robust, but YAGNI for now since i doubt we will have more clone paths

src-tauri/src/settings.rs
```diff
@@ -180,7 +180,7 @@ impl Default for KeyboardImplementation {
 
 impl Default for ModelUnloadTimeout {
     fn default() -> Self {
-        ModelUnloadTimeout::Never
+        ModelUnloadTimeout::Min5
```
VirenMohindra, line 183: this only affects fresh installs right? prob worth confirming existing users who never touched this setting won't suddenly get auto-unload behavior after upgrading

PR #1053, opened 2026-03-16
fix: add configurable clipboard restore delay to fix race condition

### LINKED ISSUE #502: [BUG] Pastes clipboard instead of spoken text
This feels so weird that I feel I must be imagining it, but some non-zero chunk of the time, Handy seems to paste the clipboard into the window rather than whatever I've just said. Sometimes it happens a few times in a row, so there's clearly some trigger, but I haven't worked out what it is. I think it might be intense processor use as that's been the case on at least one of those occasions, but I'm guessing.

This is on OSX 15.5 and Handy otherwise works beautifully.

BillionClaw:
## Problem

Fixes #502 - Under heavy system load, Handy sometimes pastes the clipboard content instead of the transcribed text.

## Root Cause

In the `paste_via_clipboard` function, there was a hardcoded 50ms delay after sending the paste keystroke before restoring the original clipboard content. On systems under intense processor load, this delay was too short - the target application hadn't yet read from the clipboard before it was restored to its original content.

## Solution

Added a new configurable setting `clipboard_restore_delay_ms` (default: 150ms) that controls the delay between sending the paste keystroke and restoring the clipboard.

### Changes:
- **settings.rs**: Added `clipboard_restore_delay_ms` field with default of 150ms
- **clipboard.rs**: Updated `paste_via_clipboard` to use the configurable delay instead of hardcoded 50ms

## Testing

This fix increases the default delay from 50ms to 150ms, providing more headroom for the paste operation to complete on loaded systems. Users experiencing this issue can further increase the delay via the setting if needed.

## Notes

- The setting is backward compatible (defaults to 150ms if not present in existing configs)
- No changes to the public API or user-facing behavior beyond fixing the race condition

### DIFF
--- a/src-tauri/src/clipboard.rs
+++ b/src-tauri/src/clipboard.rs
@@ -19,6 +19,7 @@ fn paste_via_clipboard(
     app_handle: &AppHandle,
     paste_method: &PasteMethod,
     paste_delay_ms: u64,
+    clipboard_restore_delay_ms: u64,
 ) -> Result<(), String> {
     let clipboard = app_handle.clipboard();
     let clipboard_content = clipboard.read_text().unwrap_or_default();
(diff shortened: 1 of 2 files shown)


cjpais:
Effectively the same as #1040, but using a background thread there

PR #1041, opened 2026-03-14
whisper-rs 0.16.0 and ort.rc12

cjpais:

PR #1057, opened 2026-03-16
fix(nix): fix build with ort-sys 2.0.0-rc.12

xilec:
## Summary

Fixes the Nix build broken by the migration to `ort-sys 2.0.0-rc.12` (commit b85cc03).

Two issues:

1. **Linking failure** — `ort-sys` rc.12 removed `pkg-config` support, so with `ORT_LIB_LOCATION` set it defaults to static linking. nixpkgs only provides shared `.so` libraries → build fails. Fixed by setting `ORT_PREFER_DYNAMIC_LINK=1` (same as already done in `build.yml` for macOS).

2. **Runtime crash** — `ort` rc.12 enables API v24 by default, but nixpkgs ships onnxruntime 1.23.2 (API v23). The binary builds but panics at runtime: `The requested API version [24] is not available`. Fixed by overriding onnxruntime with Microsoft's prebuilt 1.24.2 binaries via a Nix overlay + `autoPatchelfHook`.

The overlay is temporary — once [NixOS/nixpkgs#499389](https://github.com/NixOS/nixpkgs/pull/499389) (onnxruntime 1.23.2 → 1.24.3) is merged, it can be removed.

## Test plan

- [x] `nix build .#handy` completes successfully
- [x] Parakeet model loads and transcribes without API version errors
- [x] Whisper turbo model loads and transcribes via Vulkan

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -36,7 +36,42 @@
         let
           pkgs = import nixpkgs {
             inherit system;
-            overlays = [ bun2nix.overlays.default ];
+            overlays = [
+              bun2nix.overlays.default
+              # TODO: Remove this overlay once nixpkgs ships onnxruntime ≥ 1.24.
+              # Tracking PR: https://github.com/NixOS/nixpkgs/pull/499389
+              # ort-sys 2.0.0-rc.12 requires ONNX Runtime 1.24 (API v24);
+              # nixpkgs only ships 1.23.2, so use MS prebuilt binaries.
+              (final: prev: {
+                onnxruntime = let
+                  version = "1.24.2";
+                  platform = {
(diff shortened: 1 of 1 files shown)


cjpais:
@xilec overall this makes sense to me, curious though, would it be worth pre-compiling the binary to keep CI relatively fast? Right now I have pre-compiled onnx for ubuntu-22.04 and also 

It's possible we can also use the MS official distributions if they work for nix? https://github.com/microsoft/onnxruntime/releases/tag/v1.24.2

xilec:
@cjpais Good point! I've updated the PR — the overlay now uses Microsoft's official prebuilt binaries from GitHub Releases (`onnxruntime-linux-x64-1.24.2.tgz` / `onnxruntime-linux-aarch64-1.24.2.tgz`), patched with `autoPatchelfHook` for NixOS compatibility.

Tested locally — both Parakeet and Whisper turbo load and transcribe successfully with the prebuilt binaries.

xilec:
The fixes from this PR were included in PR #1041

PR #1070, opened 2026-03-16
feat: replace single microphone selector with prioritized device list

i-kozlov:
## Before Submitting This PR

## Human Written Description

I have a USB microphone connected through a docking station, but it is not set as the default device. Setting it as default is not very useful because connecting AirPods always overrides the default input device.

Because of this, I cannot use the "use system default input" setting. At the same time, I also cannot configure this USB microphone as the fixed input device for Handy, because the docking station may be disconnected, which would make the microphone unavailable.

What I would like instead is the ability to set this USB microphone as a priority device, with an automatic fallback to the MacBook built-in microphone if the USB microphone is not available.

## Related Issues/Discussions

Discussion: https://github.com/cjpais/Handy/discussions/1069

## Community Feedback

See discussion: https://github.com/cjpais/Handy/discussions/1069

## Testing

- Verified migration: existing `selected_microphone` setting is automatically moved to position 0 of the new `prioritized_microphones` list on first launch
- Tested priority fallback: with two devices in the list, disconnecting the first causes Handy to use the second
- Tested empty list: removing all prioritized devices falls back to system default (same behavior as before)
- Tested clamshell override (macOS): `clamshell_microphone` still takes priority over the list when lid is closed

## Screenshots/Videos

UI with two prioritized devices (first one active):

The settings page now shows two sections:
- **Prioritized Devices** — ordered list with Active badge, reorder (↑↓) and remove (—) controls
- **Available Devices** — all other detected devices with add (+) button
- The Active badge refreshes automatically when the settings window regains focus.
CleanShot 2026-03-22 at 10 57 32

## AI Assistance

**If AI was used:**

- Tools used: Claude Code
- How extensively: Used for implementation assistance. The Human Written Description above is my own.


cjpais:
ill check this out when I have time, I think its probably a good change

you didnt give any screenshots though... we need to have good and obvious UI for this, and it's where I am most hesitant to pull in. carefully think about the UI and UX, it needs to be very good to be pulled in.

i-kozlov:
updated to resolve conflicts

cjpais:
I don't think you reviewed my earlier comment...

i-kozlov:
my bad. i missed that.
just added screenshot example.
trying to keep UI minimalistic - just + and - to add/remove and arrows to prioritise.

cjpais:
The UI doesn't really fit nicely into the styling of the app. Please consider this. Also consider how other apps have done this

i-kozlov:
CleanShot 2026-03-22 at 12 10 35
does this one fit?
or is it at least looks better?

cjpais:
much better looking imo, ill have a think about this too. im wondering if the ordering is an advanced setting? im not quite sure what the behavior is for this. im not sure how many people will use this, or just certain people. like for myself I wont use it, but I do see the value

i-kozlov:
Thanks, i think i got the idea now:
My function targets to small auditory and should be opt-in that does not affect default behaviour.
So basically user goes to General->Microphone
In case of multiple microphones 
user goes Advanced->Input->Enable feature support.


Defaults should be
CleanShot 2026-03-22 at 15 47 50
CleanShot 2026-03-22 at 15 48 06

When function is enabled it should be like
CleanShot 2026-03-22 at 15 49 03
CleanShot 2026-03-22 at 15 49 26

cjpais:
@VirenMohindra curious for some opinions here

I more or less like this directionally @i-kozlov I want to think about it a bit more the exact flow

VirenMohindra:
yeah me too. i have a couple questions though

1. the PR description says `selected_microphone` gets seeded into the priority list on first launch, but there's no code doing this. users who enable the toggle will see an empty list. should auto-populate from their existing mic selection
2. no tests for the fallback chain, the priority logic in `audio.rs` has three early-return paths and is the core of the feature. needs at least~
   1. first device unavailable -> picks second,
   2. no priority device available -> falls back to `selected_microphone`,
   3. feature disabled → original behavior
3. update all translations, we only did the english one

should also fix these ones if possible~

- `window.addEventListener("focus", refreshAudioDevices)` registers even when multi mic is disabled. should be inside the `if (enabled)` guard.
- no input validation — `set_prioritized_microphones` accepts duplicates, empty strings, anything. filter those out please


and some other nits which are non-blocking~

- add button is hidden from keyboard users because of the `opacity-0 group-hover:opacity-100` pattern which means you can't see the (+) button when tabbing. pls add `focus:opacity-100`

and finally, rebase if you have a sec, will take a closer look!

PR #1071, opened 2026-03-16
chore: improve type safety of unload timeout options

ruszabarov:
- [x] I have searched [existing issues](https://github.com/cjpais/Handy/issues) and [pull requests](https://github.com/cjpais/Handy/pulls) (including closed ones) to ensure this isn't a duplicate
- [x] I have read [CONTRIBUTING.md](https://github.com/cjpais/Handy/blob/main/CONTRIBUTING.md)

## Human Written Description

As I was investigating the issue with unloading timeout (same as https://github.com/cjpais/Handy/pull/1051), I found some type safety problems in the settings menu. I originally thought that these type safety issues are the root cause of the models not being unloaded (and so did codex 5.4). They turned out not be the root cause of the issue but I think the component could benefit from improved type safety anyway.

I changed the serialization to `lowercase` to retain backward-compatibility with existing config.

## Related Issues/Discussions

This PR fixes the actual root cause: https://github.com/cjpais/Handy/pull/1051

## Testing

Tested by selecting different values from the dropdown. Works the same as before.

## AI Assistance

- [x] AI was used (please describe below)

Used codex 5.4 to update the `Dropdown` component to take a generic parameter `T` so that the options can be typed instead of `string`. Everything else was done by me.

PR #1079, opened 2026-03-17
dropdown type safety

ruszabarov:
## Before Submitting This PR

## Human Written Description

Improving type safety of dropdowns with a generic on `DropdownOption`, allowing option to be typed instead of plain `string`s.

## Related Issues/Discussions

https://github.com/cjpais/Handy/pull/1071#issuecomment-4074374308

## Testing

Tested all dropdowns for interactivity. Tested some of simple options like overlay position -- works correctly. Biggest risks are the Whisper acceleration option (wasn't able to find it in the settings) and the clipboard options since I had to change rust code for those.

## AI Assistance

Used codex 5.4 for `get_available_accelerators()` rust function and to add the generic parameter `T` to `DropdownOption` in typescript.

### DIFF
--- a/src-tauri/src/managers/transcription.rs
+++ b/src-tauri/src/managers/transcription.rs
@@ -765,21 +765,29 @@ fn cached_gpu_devices() -> &'static [GpuDeviceOption] {
 
 #[derive(Serialize, Clone, Debug, Type)]
 pub struct AvailableAccelerators {
-    pub whisper: Vec<String>,
-    pub ort: Vec<String>,
+    pub whisper: Vec<WhisperAcceleratorSetting>,
+    pub ort: Vec<OrtAcceleratorSetting>,
     pub gpu_devices: Vec<GpuDeviceOption>,
 }
 
 /// Return which accelerators are compiled into this build.
 pub fn get_available_accelerators() -> AvailableAccelerators {
-    use transcribe_rs::accel::OrtAccelerator;
-
-    let ort_options: Vec<String> = OrtAccelerator::available()
+    let ort_options: Vec<OrtAcceleratorSetting> = transcribe_rs::accel::OrtAccelerator::available()
         .into_iter()
-        .map(|a| a.to_string())
+        .map(|accelerator| match accelerator {
+            transcribe_rs::accel::OrtAccelerator::Auto => OrtAcceleratorSetting::Auto,
+            transcribe_rs::accel::OrtAccelerator::CpuOnly => OrtAcceleratorSetting::Cpu,
+            transcribe_rs::accel::OrtAccelerator::Cuda => OrtAcceleratorSetting::Cuda,
+            transcribe_rs::accel::OrtAccelerator::DirectMl => OrtAcceleratorSetting::DirectMl,
+            transcribe_rs::accel::OrtAccelerator::Rocm => OrtAcceleratorSetting::Rocm,
+        })
         .collect();
 
-    let whisper_options = vec!["auto".to_string(), "cpu".to_string(), "gpu".to_string()];
+    let whisper_options = vec![
+        WhisperAcceleratorSetting::Auto,
+        WhisperAcceleratorSetting::Cpu,
+        WhisperAcceleratorSetting::Gpu,
+    ];
 
     AvailableAccelerators {
         whisper: whisper_options,
--- a/src-tauri/src/managers/transcription.rs
+++ b/src-tauri/src/managers/transcription.rs
@@ -783,6 +783,9 @@ pub fn get_available_accelerators() -> AvailableAccelerators {
         })
         .collect();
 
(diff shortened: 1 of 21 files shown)


cjpais:
Whisper acceleration option is in experimental settings, must toggle it on in advanced

ruszabarov:
Tested on M1 Pro. Works correctly when switching between CPU and GPU.

ruszabarov:
Updated retention history serialization to use `lowercase`. Previously, a mix of styles was used: lowercase for `day3`, `week2`, etc and snake_case for `preserve_limit`. I had to add a `rename` to preserve backwards compatibility.

An alternative solution would be to use snake_case everywhere and add aliases for lowercase to retain backwards compatibility. Depends on the migration path you want to take.

cjpais:
#1085 will go in first then we will update this

I will take a look and test and decide the best path forward. Thanks for the cleanup in general, it's much appreciated, just needs a solid testing. Don't want to break things in the name of cleanup

VirenMohindra:
@ruszabarov we just merged #1085, good to rebase now, ping me when you need some testers!

cjpais:
@ruszabarov, could you do a test. mainly what I want to know, is does this change wipe out existing user settings

if you could, go to the current release build, change a bunch of settings and then run this. just making sure everything is backwards compatible. ideally we change as little as possible with regards to the current string names.

ruszabarov:
@cjpais I set the following settings in the release version, then launched this dev version and the settings were preserved. I used the following values
- Overlay Position: Bottom
- Unload Model: After 5 minutes
- Paste Method: Direct
- Clipboard Handing: Copy to Clipboard
- Auto Submit: Enter
- Auto-Delete Recordings: After 3 days
- Keyboard Implementation: Handy Keys
- Whisper Acceleration: GPU
(Microphone and Output Device were untouched)

Also checked the raw `settings_store.json` and the current release version stores values like `min5`, `days3`, and `preserve_limit`, so the new serialization should be compatible. 

Backwards compatibility is something I paid close attention to when making these changes. That is why I went with `lowercase` serialization instead of `snake_case`. This ensures backwards compatibility but breaks style (`snake_case` is used almost everywhere else).

ruszabarov:
@VirenMohindra I could use another set of eyes on the Whisper Accelerator and Ort Accelerator option 🙂

src/stores/settingsStore.ts
```diff
@@ -106,7 +107,9 @@ const settingUpdaters: {
         : (value as string),
     ),
   recording_retention_period: (value) =>
-    commands.updateRecordingRetentionPeriod(value as string),
+    commands.updateRecordingRetentionPeriod(value ?? "never"),
+  model_unload_timeout: (value) =>
+    commands.setModelUnloadTimeout(value ?? "min5"),
```
VirenMohindra: nit: is it possible to pull these out and share them? if we ever deviate from never and min5 as defaults, we should have a single source of truth

src-tauri/src/managers/transcription.rs
```diff
@@ -769,20 +769,28 @@ pub fn apply_accelerator_settings(app: &tauri::AppHandle) {
...
+            transcribe_rs::accel::OrtAccelerator::Rocm => OrtAcceleratorSetting::Rocm,
+        })
         .collect();
 
-    let whisper_options = vec!["auto".to_string(), "cpu".to_string(), "gpu".to_string()];
+    let whisper_options = vec![
+        WhisperAcceleratorSetting::Auto,
+        WhisperAcceleratorSetting::Cpu,
+        WhisperAcceleratorSetting::Gpu,
+    ];
```
VirenMohindra, line 793: nit: if `WhisperAcceleratorSetting` gains a new variant (like CoreMl), this list won't update and the compiler won't warn. worth either mirroring the ort pattern or adding a comment explaining why these three are always the full set

VirenMohindra (approved):
LGTM, nice work

src-tauri/src/managers/transcription.rs
ruszabarov: Added a comment

src/stores/settingsStore.ts
ruszabarov: I decided to `Exclude` all the `undefined` values from `AppSettings`. `undefined` is not a legal option for `value`. The behavior should stay effectively the same since we were doing `as boolean`, `as string`, `?? "never"`, etc before

ruszabarov:
@cjpais just rebased and fixed all the conflicts. Since we are building a combined list `Auto, [GPU devices...], CPU`, we can no longer enforce a strict type on that dropdown, so I had to downgrade it to just `string`. Please check and merge if you have a chance 🙂

cjpais:
I think this is fine to do, but at the end of the day, I'm going to close it for the time being and we can pick clean up later. I'm more concerned with other bugs at the moment

PR #1083, opened 2026-03-17
fix: fallback to next audio format if stream build fails

givizager-boop:
## Problem

On some devices (e.g. Synaptics audio controllers), `build_input_stream` fails with error `0x80070057` when using F32 format, even though the device reports it as supported. The old code picked the best format once and gave up on failure.

## Fix

Replace single-try logic with a retry loop that tries all supported formats in preference order (F32 → I16 → I32 → others). On failure, logs a warning and tries the next format.

- `get_preferred_config` → `get_configs_by_preference` returning sorted `Vec`
- Loop over configs, clone `sample_tx` for each attempt
- First successful `build_stream` wins

## Tested on

- Synaptics HD Audio Device (Windows) — previously crashed on F32, now falls back to I16 ✓

### DIFF
--- a/src-tauri/src/audio_toolkit/audio/recorder.rs
+++ b/src-tauri/src/audio_toolkit/audio/recorder.rs
@@ -78,60 +78,58 @@ impl AudioRecorder {
 
         let worker = std::thread::spawn(move || {
             let init_result = (|| -> Result<(cpal::Stream, u32), String> {
-                let config = AudioRecorder::get_preferred_config(&thread_device)
+                let configs = AudioRecorder::get_configs_by_preference(&thread_device)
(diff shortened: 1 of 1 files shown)


givizager-boop:
Hi\! Could you trigger the \ workflow for this PR? The fix addresses a real issue with Synaptics audio devices on Windows where \ fails silently on F32 format. I can't build locally because the project has a dependency issue between \ and \ — \ is forward-declared as opaque in the bundled whisper.cpp header, causing compile errors. A test build from CI would let me verify the fix works. Thanks\!

cjpais:
Closing in favor of #1084, we can discuss further there if you would like!

PR #1085, opened 2026-03-18
fix: prevent idle watcher from unloading model during recording

VirenMohindra:
## Before Submitting This PR

## Human Written Description

when `model_unload_timeout` is set to `Immediately`, the idle watcher thread unloads the model mid-recording. this kills transcription for any sentence longer than ~10 seconds. bisected to d1da935.

`Immediately` returns `Some(0)` from `to_seconds()`, so the idle watcher (which checks every 10s) sees `idle_ms > 0`, which is always true, and unloads the model on its next tick, even while the user is still speaking. the model gets pulled out from under the transcription pipeline.

the `Immediately` variant is already correctly handled by `maybe_unload_immediately()` which fires after each transcription completes. the idle watcher just needs to skip it.

## Related Issues/Discussions

regression introduced in d1da935 (#1051)

## Testing

- [x] bisected: confirmed transcription works on 789a439 (before d1da935), fails on d1da935 (main)
- [x] with fix: long transcription completes successfully with `Immediately` setting
- [x] `maybe_unload_immediately` still fires after transcription ("Immediately unloading model after transcription" in logs)
- [ ] verify other timeout values (Min2, Min5, etc.) still unload correctly after idle

## AI Assistance

**If AI was used:**

- Tools used: claude code
- How extensively: bisected the regression, investigated root cause, wrote the one-line fix

### DIFF
--- a/src-tauri/src/managers/transcription.rs
+++ b/src-tauri/src/managers/transcription.rs
@@ -232,8 +232,7 @@ impl TranscriptionManager {
 
     /// Reset the idle timer to now.
     fn touch_activity(&self) {
-        self.last_activity
-            .store(Self::now_ms(), Ordering::Relaxed);
+        self.last_activity.store(Self::now_ms(), Ordering::Relaxed);
     }
 
     /// Unloads the model immediately if the setting is enabled and the model is loaded


cjpais:
I think the fix was incomplete. I added some changes which should help.

Basically there was still a race that existed, if you have it set to 5 seconds, or 2 minutes, you could feasibly have the same issue.

Instead we need to see if we are recording currently. If we are, we cannot unload the model. Also to make sure we don't have another race, we reset the idle_ms time every time the watcher tick fires and we are recording. The debug setting has been changed to 15 seconds, above the 10 second tick rate of the watcher thread. This is so there is no chance of a race on that setting.

Please let me know how the code looks and what you think

VirenMohindra:
thanks for the fixes, guard was super helpful

> Please let me know how the code looks and what you think

looks good to me, perf is great. tested locally on for this branch and ran several transcriptions including longer sentences (~20s of speech) with immediately set. all completed correctly, immediately unloading model after transcription fires each time

memory still drops from ~700MB to ~84MB physical footprint after each unload. no mid-recording unloads, no errors

lets 🚢 🔥

cjpais:
Sweet thanks for the confirmation I also had no issues when testing :)

PR #1095, opened 2026-03-19
fix: sha256 verification to prevent corrupt partial download loop

### LINKED ISSUE #858: [BUG] Partial downloads prevent the model from ever being downloaded again

VirenMohindra:
## Before Submitting This PR

- [x] I have searched existing issues and pull requests to ensure this isn't a duplicate

## Human Written Description

the partial download bug (#858) has been causing infinite loops for a long time — corrupt .partial files survive extraction failures and get resumed on next launch, trapping users with no way out except manually hunting down and deleting the file. i traced the root cause to commit 86a1205 where cancel_download() only flips a flag but doesn't cancel the in-flight request, leaving partial files on disk in unknown states.

sha256 checksums catch corruption regardless of cause (interrupted download, bad resume, anything), and hardcoding them next to each model's url means the fix is transparent and easy to maintain. also deletes .partial on extraction failure as a belt-and-suspenders fix for cases where checksums aren't available (e.g. future custom model support).

## Related Issues/Discussions

Fixes #858
Related: #1015

## Community Feedback

#858 has been open since feb 2026 with multiple users confirming the bug. this is a complete fix rather than the cleanup-on-failure approach in #1015.

## Testing

- fresh model download: verifies sha256 post-download, proceeds normally
- corrupt .partial (simulated by truncating): hash mismatch → deletes partial → next launch starts fresh, no loop
- extraction failure: .partial deleted, next attempt re-downloads cleanly
- download failure: surfaces toast in both settings page and onboarding (previously silent in settings)

## AI Assistance

**If AI was used:**

- Tools used: Claude Code
- How extensively: pair programmed the full implementation — root cause investigation, library research, architecture decisions (hardcoded hashes vs manifest endpoint, sha256 vs etag), and code writing. all tradeoffs reviewed and approved by me.

### DIFF
--- a/src/components/model-selector/ModelStatusButton.tsx
+++ b/src/components/model-selector/ModelStatusButton.tsx
@@ -34,7 +34,7 @@ const ModelStatusButton: React.FC<ModelStatusButtonProps> = ({
       case "downloading":
         return "bg-logo-primary animate-pulse";
       case "verifying":
-        return "bg-blue-400 animate-pulse";
+        return "bg-orange-400 animate-pulse";
       case "extracting":
         return "bg-orange-400 animate-pulse";
       case "error":
--- a/src/i18n/locales/ar/translation.json
+++ b/src/i18n/locales/ar/translation.json
@@ -119,6 +119,8 @@
     "custom": "مخصص",
     "active": "نشط",
     "noModelsAvailable": "لا توجد نماذج متاحة",
+    "verifying": "جارٍ التحقق من {{modelName}}...",
+    "verifyingGeneric": "جارٍ التحقق...",
     "extracting": "...جاري استخراج {{modelName}}",
     "extractingMultiple": "...جاري استخراج {{count}} من النماذج",
     "extractingGeneric": "...جاري الاستخراج",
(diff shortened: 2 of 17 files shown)


VirenMohindra:
would love any feedback and thoughts @sheeki03

PR #1112, opened 2026-03-22
untested - migrate events to specta

cjpais:


VirenMohindra:
happy to test if we kick off a test build

PR #1117, opened 2026-03-22
feat: add sort-by option to models settings page

goellavish10:
## Before Submitting This PR

## Human Written Description

When browsing the Models settings page, there's no way to sort models, I was stuck with the default ordering. I wanted to quickly compare models by accuracy, speed, or size to pick the right one for my use case. This adds a sort dropdown next to the existing language filter, keeping the UI consistent while making model selection more intuitive.

## Related Issues/Discussions

No existing issue or discussion for this feature — searched both open and closed issues/PRs and found no prior work on model sorting.

## Community Feedback

This is a new contribution. Happy to start a discussion if maintainers would like community input before merging.

## Testing

- Ran `bun run tauri dev` and verified the sort dropdown renders correctly next to the language filter
- Tested all four sort options (Default, Accuracy, Speed, Size) — models reorder correctly in both "Downloaded" and "Available to Download" sections
- Tested ascending/descending toggle — direction reverses as expected
- Verified active model stays pinned first regardless of sort
- Verified sorting works correctly when combined with language filtering
- Ran `bun run lint` — 0 errors
- Ran `bun run format:check` (Prettier) — all files pass

## Screenshots/Videos

Sort feature — sorting by Accuracy (ascending)
Sort feature — sorting by Accuracy (ascending)

## AI Assistance

**If AI was used:**

- Tools used: Claude Code
- How extensively: Wrote most of the implementation code and translations. I reviewed and directed all changes


cjpais:
please improve the UI.

goellavish10:
## UI Improvements

Updated the sort and language filter dropdowns for better visual consistency:

- **Unified sort button**  replaced the split-button pattern (label + separate direction toggle) with a single cohesive button that shows direction inline
- **Dropdown indicators** selected items now show checkmarks or direction arrows for clearer feedback

See the updated screenshots above

cjpais:
I do appreciate this. I'm just going to close this for now because I'm not 100% sure what I want to do with the models page right now. At some point, I may want to kind of refactor it entirely. I think there's probably some changes that are going to make it in shortly in relation to streaming and non-streaming. I think sorting might be helpful as well since there's a lot of models. My main concern is that what we're sorting based on is like kind of subjective, and that's not really my favorite. So it's possible something like this will make it into a future release where maybe we have two different UIs, like a power user UI versus like a normal user because probably normal users don't really need to have such features, and we actually probably don't even want to tell them what specific model they are using under the hood

PR #1120, opened 2026-03-23
settings: Add ORT thread count setting with auto-tune benchmark.

andrewleech:
### Summary

ORT's default of "use all cores" isn't always the fastest for ONNX inference — on my Zen 5 machine, 6 threads gives 7.1x realtime vs ~6.7x at 16 threads for Qwen3 0.6B Int4. But finding the optimal count requires manual trial-and-error with no feedback.

This adds two things:

1. **Thread count setting** — a number input (0-32, where 0 = auto) in Advanced → Experimental Features that controls `transcribe_rs::accel::set_ort_intra_threads`. Changing it unloads the current model so the next transcription picks up the new session configuration.

2. **Auto-tune benchmark** — an "Auto" button next to the input that opens a modal dialog, transcribes real recordings from the user's history at different thread counts, and shows the results. It tests a coarse grid (1, 2, 4, 6, 8, 12, 16, 20, 24, 32), starting from mid-range for quick initial feedback. On completion it shows a table of all results with RTF (real-time factor) so the user can see the performance curve and apply the best value.

handy_8C0jgYuQDw
handy_bMKBkPQRsQ
handy_Ig4Kt31dpp

The benchmark:
- Selects recordings from history totalling ≥15 seconds (longest-first)
- Auto-loads the selected model if it was idle-unloaded
- Tests mid-range first, then max, then 1, then fills the remaining grid
- Supports cancellation between trials
- Uses RAII to guarantee the original thread count is restored on completion or panic
- Shows all trial results in a table with the best highlighted

Depends on [transcribe-rs feat/ort-thread-count](https://github.com/andrewleech/transcribe-rs/tree/feat/ort-thread-count) (PR forthcoming). The DROP commit pins to this branch and should be dropped when it merges upstream.

### Testing

- Tested on Windows (Zen 5 + AMD 860M iGPU) with Qwen3 ASR 0.6B Int4 and Parakeet V3
- Verified: setting persists across restarts, model reloads on change, benchmark completes and applies result
- Verified: cancel mid-benchmark restores previous state
- Verified: benchmark with no recordings shows error message

- The benchmark requires a full model reload per trial (~4-12s each depending on model size) because ORT session thread count is baked in at creation time. For a 10-point grid this means ~1-2 minutes total.
- The coarse grid [1,2,4,6,8,12,16,20,24,32] may miss the exact optimum but gets within ±2 threads of it.
- The benchmark locks the engine exclusively — the user can't record while it runs.

### DIFF
--- a/src-tauri/Cargo.lock
+++ b/src-tauri/Cargo.lock
@@ -99,9 +99,9 @@ dependencies = [
 
 [[package]]
 name = "anstream"
-version = "0.6.21"
+version = "1.0.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "43d5b281e737544384e969a5ccad3f1cdd24b48086a0fc1b2a5262a26b8f4f4a"
+checksum = "824a212faf96e9acacdbd09febd34438f8f711fb84e09a8916013cd7815ca28d"
 dependencies = [
  "anstyle",
  "anstyle-parse",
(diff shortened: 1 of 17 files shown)


andrewleech:
Depends on [transcribe-rs PR #64](https://github.com/cjpais/transcribe-rs/pull/64) for `set_ort_intra_threads` / `get_ort_intra_threads`. The DROP commit pins to that branch and should be dropped when it merges.

cjpais:
Just want to say I appreciate the work you've put into this, I will be deprecating the ORT path for Handy in the coming weeks/months so I will close this PR. I think there is an opportunity to still do something like this in the next version I am planning

PR #1121, opened 2026-03-23
Fix (linux) overlay problem in kde

vpsone:
## Before Submitting This PR

## Human Written Description

This PR fixes the overlay issue in KDE Plasma. After PR #769, the overlay in KDE Plasma started behaving like a regular window, which prevents it from auto-pasting the response after a recording is finished. I have used v0.7.3 (the last version where the overlay worked correctly) for a long time and have not encountered any stability issues due to the `gtk-layer-shell protocol`. Here is a video of  it:
[Screencast_20260323_135432.webm](https://github.com/user-attachments/assets/a75688c0-20c2-4caf-922e-6bf99ff06d61)

Here is how it looks after updating to v0.8.0:

image

## Related Issues/Discussions

#769

## AI Assistance

**If AI was used:**

- Tools used:
- How extensively:

### DIFF
--- a/src-tauri/src/overlay.rs
+++ b/src-tauri/src/overlay.rs
@@ -17,8 +17,6 @@ use tauri_nspanel::{tauri_panel, CollectionBehavior, PanelBuilder, PanelLevel};
 
 #[cfg(target_os = "linux")]
 use gtk_layer_shell::{Edge, KeyboardMode, Layer, LayerShell};
-#[cfg(target_os = "linux")]
-use std::env;
 
 #[cfg(target_os = "macos")]
 tauri_panel! {
@@ -69,20 +67,6 @@ fn update_gtk_layer_shell_anchors(overlay_window: &tauri::webview::WebviewWindow
 /// Returns true if layer shell was successfully initialized, false otherwise
 #[cfg(target_os = "linux")]
 fn init_gtk_layer_shell(overlay_window: &tauri::webview::WebviewWindow) -> bool {
-    // On KDE Wayland, layer-shell init has shown protocol instability.
-    // Fall back to regular always-on-top overlay behavior (as in v0.7.1).
-    let is_wayland = env::var("WAYLAND_DISPLAY").is_ok()
-        || env::var("XDG_SESSION_TYPE")
-            .map(|v| v.eq_ignore_ascii_case("wayland"))
-            .unwrap_or(false);
-    let is_kde = env::var("XDG_CURRENT_DESKTOP")
-        .map(|v| v.to_uppercase().contains("KDE"))
-        .unwrap_or(false)
-        || env::var("KDE_SESSION_VERSION").is_ok();
-    if is_wayland && is_kde {
-        debug!("Skipping GTK layer shell init on KDE Wayland");
-        return false;
-    }
 
     if !gtk_layer_shell::is_supported() {
         return false;
--- a/src-tauri/src/overlay.rs
+++ b/src-tauri/src/overlay.rs
@@ -18,6 +18,9 @@ use tauri_nspanel::{tauri_panel, CollectionBehavior, PanelBuilder, PanelLevel};
 #[cfg(target_os = "linux")]
 use gtk_layer_shell::{Edge, KeyboardMode, Layer, LayerShell};
 
+#[cfg(target_os = "linux")]
+use std::env;
+
 #[cfg(target_os = "macos")]
 tauri_panel! {
     panel!(RecordingOverlayPanel {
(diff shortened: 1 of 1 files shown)


cjpais:
Unfortunately this fix cannot be approved because it will break the ability of the app to work on many systems. This is a much worse outcome than just the overlay being broken.

vpsone:
But How I just reversed the changes in this pr #769 .

cjpais:
yeah and you think that PR was not for a reason?! it fixes stability issues from what I remember

please let me know if I am wrong or misguided

vpsone:
But I have not faced any issue; only I have to install the gtk-layer-shell dependency in the system. Here is the issue a user can get if they do not have the dependency installed.
image

and here after installing it (v0.7.3)
image

Additionally, PR #769 did not clearly explain the original problem, simply referring to it as a 'stability issue'.

cjpais:
Please go ask in that thread, what the stability issue was in particular. 

I'm not willing to trade one issue for another unless it's known across many users that it won't cause a regression.

I understand that there was a regression caused in that change, but a sample size of one especially for Linux is not enough.

vpsone:
image
But It is already broken in v0.8.0 with two cancel button.

vpsone:
If you can provide test build for this pr then I can ask others to test it.

vpsone:
I have tried test build and it is working perfectly. 

https://github.com/user-attachments/assets/fb693457-3b05-450b-930c-fdc627ff1df1

xilec:
Tested the fix on my system — overlay works correctly: displays speaking/silence timers properly and doesn't interfere with pasting into the focused window.

**My setup:**
- NixOS
- KDE Plasma 6 / Wayland
- `gtk-layer-shell` installed as a dependency

Built and ran from this branch via `nix run github:vpsone/Handy/fix-kde-overlay`. No stability issues on startup or during use.

VPS07:
I have tested build on `fedora 43 kde plasma 6/Wayland` and it is working correctly. I am facing no stability issues on startup.

xilec:
Summarizing the current state for visibility:

- The fix has been confirmed working on two KDE Plasma / Wayland setups (NixOS, Fedora 43 and @vpsone's system)
- @boeserwolf hasn't had any GitHub activity since his PR #769 was merged, and hasn't responded to @vpsone's [request to test](https://github.com/cjpais/Handy/pull/769#issuecomment-4104498498) with `gtk-layer-shell` installed

If anyone on CachyOS or Arch could test this branch, that would help rule out the startup stability issue from #769. Please make sure `gtk-layer-shell` is installed before testing.
@dadaphl @rogsme @theyonecodes — would any of you be able to give it a try?

cjpais:
Thanks everyone for testing this so far. And thank you for bringing more people to test. Once we get more confirmations, I will merge this. @vpsone @xilec

xilec:
Since there's been limited feedback from Linux users, I'd like to suggest a compromise approach instead of simply removing the KDE+Wayland check.

The current PR removes the detection heuristic entirely, which works — but the original concern from #769 (startup instability) hasn't been fully investigated since @boeserwolf is inactive.

**Alternative: opt-out via environment variable**

Instead of hardcoding DE detection, we could replace the KDE+Wayland check with something like:

```rust
if env::var("HANDY_NO_GTK_LAYER_SHELL").is_ok() {
    debug!("Skipping GTK layer shell init (HANDY_NO_GTK_LAYER_SHELL is set)");
    return false;
}
```

This way:
- Overlay works for **everyone** by default (including KDE Plasma users)
- Users who experience startup stability issues have an escape hatch
- No brittle DE detection logic
- If anyone reports issues, they can set `HANDY_NO_GTK_LAYER_SHELL=1` and file a proper bug with logs

The env variable could be documented in the troubleshooting section of README.

This gives us the same fix as the current PR while keeping a safety net for edge cases we can't reproduce.

cjpais:
I think this is a good compromise @xilec

rogsme:
I'm sorry I haven't been able to provide feedback, but I don't use KDE or Wayland 😕

vpsone:
@xilec Thanks for your suggestions. I have changed the code as you suggested. Please let me know if it needs any other changes.

xilec:
@vpsone Looks good! One small thing — it would be helpful to also document `HANDY_NO_GTK_LAYER_SHELL` in the troubleshooting section of README, so users experiencing startup issues on Linux know about this option.

xilec:
@cjpais I think this PR is good to merge as-is. The env var fallback is in place, the fix has been confirmed working on multiple KDE Plasma / Wayland setups, and keeping it blocked on documentation would be unfortunate given how long it's already been open.

I'll take care of documenting `HANDY_NO_GTK_LAYER_SHELL` in the troubleshooting section as a **follow-up PR** after this one lands, so that this one isn't held up any longer.

PR #1123, opened 2026-03-23
rearrange some ui

cjpais:
moves history above post processing and also moves the cancel shortcut to general when ptt is disabled

Screenshot 2026-03-23 at 8 18 28 PM


cjpais:
I probably am going to break something but yolo

PR #1125, opened 2026-03-23
fix(macos): allow function keys (F1-F20) as shortcuts regardless of keyboard mode

fullstck:
## Summary

On macOS, pressing F5 (the mic/dictation key) can't be used as a transcribe shortcut because of how the `fn` modifier is handled during hotkey matching in `handy-keys`.

### Root Cause

On macOS, the top row keys default to media functions. To use F5 as a function key, users must hold `fn`, which sets the `FN` modifier flag. The `handy-keys` library does **exact matching** on FN in `Modifiers::matches()`, so a shortcut registered as `"f5"` never matches an `fn+F5` event.

### Fix

The fix is implemented at the **library level** in handy-keys ([handy-computer/handy-keys#16](https://github.com/handy-computer/handy-keys/pull/16)), where it benefits all consumers. The macOS listener now strips the FN modifier from function key events (F1-F20) before they reach the matching logic, since `fn` is a transport mechanism to access F-keys — not a meaningful modifier.

This PR simply points Handy's `handy-keys` dependency to the fixed version. No application-level workarounds are needed — function keys just work.

**Changes in this PR:**
- `Cargo.toml`: Point `handy-keys` to the fork with the fix
- `handy_keys.rs`: Revert to original (no workarounds needed)

**Once handy-computer/handy-keys#16 is merged and released**, this dependency should be updated to point to the published crate version.

## Test plan

- [ ] On macOS with default keyboard (media keys mode):
  - [ ] Record F5 as transcribe shortcut (press fn+F5) → should show "F5" not "fn+F5"
  - [ ] Press fn+F5 → should trigger transcription
- [ ] On macOS with "Use standard function keys" enabled:
  - [ ] Record F5 as transcribe shortcut (press F5 without fn) → should show "F5"
  - [ ] Press F5 → should trigger transcription
- [ ] Other F-keys (F1-F12) work as shortcuts
- [ ] Non-function-key shortcuts (e.g., Option+Space) are unaffected
- [ ] Modifier+F-key combos (e.g., Cmd+F5) work correctly

### DIFF
--- a/src-tauri/Cargo.lock
+++ b/src-tauri/Cargo.lock
@@ -2475,8 +2475,7 @@ dependencies = [
 [[package]]
 name = "handy-keys"
 version = "0.2.4"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "e1c1bb519fbead4e7344bf0fe32718c4cea30fc05e03fc65c242df064f515b03"
+source = "git+https://github.com/fullstck/handy-keys?branch=fix%2Fmacos-fn-function-keys#ceab00e00c87e5bc5fcb5f4340f3505c69bdf9cb"
 dependencies = [
  "bitflags 2.11.0",
  "block2",
--- a/src-tauri/Cargo.toml
+++ b/src-tauri/Cargo.toml
@@ -70,7 +70,7 @@ tar = "0.4.44"
 flate2 = "1.0"
 sha2 = "0.10"
 transcribe-rs = { version = "0.3.2", features = ["whisper-cpp", "onnx"] }
-handy-keys = "0.2.4"
+handy-keys = { git = "https://github.com/fullstck/handy-keys", branch = "fix/macos-fn-function-keys" }
 ferrous-opencc = "0.2.3"
 clap = { version = "4", features = ["derive"] }
 specta = "=2.0.0-rc.22"
--- a/src-tauri/src/shortcut/handy_keys.rs
+++ b/src-tauri/src/shortcut/handy_keys.rs
@@ -27,8 +27,8 @@
 //! polled from a dedicated recording thread. Events are emitted to the frontend
 //! via Tauri's event system.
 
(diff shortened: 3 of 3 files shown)


cjpais:
Shouldn't we make changes to the handy-keys library, so everyone who uses that library can benefit from this change? or at bare minimum have it as an option.

fullstck:
You're right — moved the fix to the handy-keys library: https://github.com/handy-computer/handy-keys/pull/16

The library-level fix strips the FN modifier from function key events (F1-F20) in the macOS listener, so all consumers get correct behavior. The change adds a `Key::is_function_key()` helper and uses it in the event callback to strip FN before events reach the matching logic.

This PR now just points to the fixed handy-keys — no application-level workarounds needed. Once the handy-keys PR is merged and released, the dependency can be updated to the published crate version.

AlexanderYastrebov:
handy-keys PR was closed and I doubt it fixed anything. F-keys in macos generate different key codes depending on the "Use F1, F2, etc keys as standard function keys" option.
See related https://github.com/handy-computer/handy-keys/pull/17

fullstck:
> handy-keys PR was closed and I doubt it fixed anything. F-keys in macos generate different key codes depending on the "Use F1, F2, etc keys as standard function keys" option. See related [handy-computer/handy-keys#17](https://github.com/handy-computer/handy-keys/pull/17)

WisprFlow manages to make it work so there is a workaround

AlexanderYastrebov:
> WisprFlow manages to make it work so there is a workaround

In macos some media keys have distinct keycodes when handled through CGEvent while others do not (see https://github.com/handy-computer/handy-keys/pull/17#discussion_r2991164957). To properly handle media keys the underlying library would need to process HID events instead of CGEvents e.g. like Karabiner does, see https://github.com/pqrs-org/Karabiner-Elements/blob/main/DEVELOPMENT.md#the-difference-of-event-grabbing-methods

PR #1145, opened 2026-03-25
fix: don't log cloud provider keys

shaankhosla:

## Human Written Description

Currently, the entire AppSettings struct gets logged in certain situations, the one I found was on this line:

```
                debug!("Found existing settings: {:?}", settings);
```

This is dangerous, because it logs the API keys associated with post-processing, which may get shared and used maliciously.

I changed the struct to not output keys when dbg

## Testing

Added test case to confirm dbg's don't log the keys

## AI Assistance

**If AI was used:**

- Tools used: Claude code
- How extensively: i reviewed change and tested output, also experienced personally with Rust


src-tauri/src/settings.rs
```diff
@@ -401,6 +402,67 @@ pub struct AppSettings {
...
 
+impl fmt::Debug for AppSettings {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        let redacted_keys: HashMap<&String, &str> = self
+            .post_process_api_keys
+            .iter()
+            .map(|(k, v)| (k, if v.is_empty() { "" } else { "[REDACTED]" }))
+            .collect();
+
+        f.debug_struct("AppSettings")
```
VirenMohindra: this manual `Debug` impl is a bit brittle every new field added to `AppSettings` needs to be mirrored here, and forgetting one silently drops it from debug output. a less maintenance heavy approach would be  wrap the sensitive field in a newtype that handles its own redaction, so `AppSettings` can keep `#[derive(Debug)]` ~

```rs
#[derive(Clone, Serialize, Deserialize, Type)]
#[serde(transparent)]
pub struct SecretMap(HashMap<String, String>);

impl fmt::Debug for SecretMap {
    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
        let redacted: HashMap<&String, &str> = self.0.iter()
            .map(|(k, v)| (k, if v.is_empty() { "" } else { "[REDACTED]" }))
            .collect();
        redacted.fmt(f)
    }
}

impl std::ops::Deref for SecretMap {
    type Target = HashMap<String, String>;
    fn deref(&self) -> &Self::Target { &self.0 }
}

impl std::ops::DerefMut for SecretMap {
    fn deref_mut(&mut self) -> &mut Self::Target { &mut self.0 }
}
```

then change the field type~
```rs
pub post_process_api_keys: SecretMap,
```

this way AppSettings keeps #[derive(Debug)] and new fields are automatically included. the `Deref` / `DerefMut` impls mean all existing `.post_process_api_keys.insert(...)` / `.get(...)` calls work unchanged and `#[serde(transparent)]` preserves the json shape

shaankhosla: Got it, and yes that makes sense. I'll make those changes in a bit

shaankhosla:
@VirenMohindra thanks for the feedback, let me know if there are any other changes you'd like me to make

src-tauri/src/settings.rs
```diff
@@ -304,6 +305,34 @@ impl Default for OrtAcceleratorSetting {
     }
 }
 
+#[derive(Clone, Serialize, Deserialize, Type)]
+#[serde(transparent)]
+pub struct SecretMap(HashMap<String, String>);
```
VirenMohindra: nit: only used within settings.rs. we can prob change this to `pub(crate)` to signal it's not part of the public API

src-tauri/src/settings.rs
```diff
@@ -911,4 +940,24 @@ mod tests {
         assert!(!settings.auto_submit);
         assert_eq!(settings.auto_submit_key, AutoSubmitKey::Enter);
     }
+
+    #[test]
+    fn debug_output_redacts_api_keys() {
```
VirenMohindra, line 952: nit: test could also assert the `SecretMap` directly, not just via AppSettings~
```rs
  #[test]
  fn secret_map_debug_redacts_values() {
      let map = SecretMap(HashMap::from([
          ("key".into(), "secret".into()),
      ]));
      let out = format!("{:?}", map);
      assert!(!out.contains("secret"));
      assert!(out.contains("[REDACTED]"));
  }
```

src-tauri/src/settings.rs
shaankhosla: done

src-tauri/src/settings.rs
shaankhosla: done

cjpais:
@VirenMohindra I don't think we need to worry about previous logs, nor delete things. Post processing is alpha so this can be expected. We will do a better job managing keys in the future but we don't need to retroactively make changes.

shaankhosla:
the cicd says it's failing because there's 1 cancelled check, but i'm unable to rerun it. Is there anything else needed from me for this PR?

VirenMohindra:
not on my end. im happy with how the PR looks. 

> We will do a better job managing keys in the future

I previously had a PR to store the API keys in the OS keychain [1], might make sense to rehash this if we want to move post processing / polish to beta, but agreed, not P1 for now 

[1] https://github.com/cjpais/Handy/pull/814

Issue #646, opened 2026-01-22
[BUG] Audio session activation triggers Handoff, stealing AirPods from other devices


notnotjake:
## Bug Description

When a user is listening to audio on another device via AirPods and Handy activates and plays the sound effect on Mac, the system treats Handy audio as primary media playback. This causes AirPods to switch routing to the Mac, interrupting the user's music. 

Sound effects should play through the currently active audio route and not cause AirPods Handoff.

## System Information

**App Version:**

0.7.0

**Operating System:**

macOS 26.3

**CPU:**

M1 Max

**GPU:**

M1 Max

notnotjake:
The behavior is not caused by the recording, only the Audio Feedback setting. When that is set to true, then the playback of those causes the audio handoff. Turning it off fixes the issue I am describing.

cjpais:
Yeah, this is definitely a bug and I can replicate it.

notnotjake:
Was able to implement a fix using AVAudio that works, but need to clean up the code then I could submit PR: [fork](https://github.com/notnotjake/Handy/commit/2232eb4cc662c3a432239db3143b400e45597773)

cjpais:
I'm pretty open to a PR on this, I would love to eventually move the bridge code and audio handling to it's own library similar to handy-keys but for now we can do it in situ

abacchilb:
Just to add to this one- 
When AirPods are the Mac output, Handy also causes output volume to jump to 100% when playing the feedback sound (and when unmuting after recording).

notnotjake:
Hmm, I've actually not ever experienced that.

3manu31:
> Just to add to this one- When AirPods are the Mac output, Handy also causes output volume to jump to 100% when playing the feedback sound (and when unmuting after recording).

I experience this too, not just the volume getting too high, but also muffled, like the quality of the sound is bad. Turns out it's because both the input and output are triggered at the same time, and this causes macOS to think it has entered a call, and so switches to call-quality audio (CoreAudio profile switching).
I've opened a pull request (#1146) that adds a fallback to non-bluetooth input devices on macos to avoid this profile switch 

If anyone affected by this can try that branch and comment on the PR to confirm whether it fixes the issue for you (and if you notice any side effects), that would be super helpful!

3manu31:
I actually just tested it against the issue #646 itself, and turns out my pull request https://github.com/cjpais/Handy/pull/1146 addresses the issue itself too.
[...]

PR #1146, opened 2026-03-25
fix: macOS Bluetooth input fallback to prevent media pause during recording

3manu31:
Prevents audio pause/glitches during recording with any Bluetooth input device as the default microphone by prioritizing non-Bluetooth alternatives to avoid CoreAudio profile switches that some media apps interpret as route changes.

## Before Submitting This PR

## Human Written Description

 Recently started using Handy, however, I kept finding myself not using much as I wear airpods all day, and when I do wear airpods my experience with Handy has been far from perfect. 2 issues: 1st: when listening to music (on airpods) and then start recording with Handy, then the music becomes muffled and too loud, because a switch to CoreAudio profile occurs as if I am in a call, and the moment I stop recording (when the mic is not in use anymore), then it goes back to normal. 2nd: I use an app for white noise that auto pauses when CoreAudio profile switching occurs (e.g. when switching output sources, or start/end a call), and so each time I would trigger Handy, then the white noise would completely pause).
I fixed it by adding a fallback to non-bluetooth input devices on macOS when the default input is Bluetooth, preventing those switches that cause the problem.
I think it would be good to implement the fix on the official version of Handy. P.s. if necessary, I would be willing to also add it as a setting on the advanced settings page, for users to optionally opt for it or not, I know I personally would because I keep finding myself having to choose between using my airpods and using Handy, and I don't want to have to choose.

## Related Issues/Discussions

Fixes the unwanted behavior described in #646, where AirPods handoff is triggered by the use of Handy. Additionally, it addresses another issue reported in a comment of #646 (https://github.com/cjpais/Handy/issues/646#issuecomment-3879818081), where Airpods as mac audio output cause the sound volume to get too loud (and the quality to become muffled).

## Community Feedback

 I haven’t gathered community feedback yet. This fix is based on my own reproducible issue.

## Testing

MacOS 26.3.1 (and also MacOS 26.4), Handy version: v0.8.0.
I repeated the same tests that I initially did to pinpoint the point where the issue was:
- Used handy with Airpods and music playing on Apple Music on my mac, with and without "Mute while recording" on.
- used handy with Airpods and my white noise app, with and without "mute while recording" on.
- did the same thing but without airpods (system audio) this time.

After that I did another test to confirm the issue #646 is fixed:
- on Handy version v0.8.0 I managed to replicate the issue reported, by playing music on my iphone with output to my Airpods, and then triggered Handy on Mac. Handoff got triggered and the music on the iphone stopped playing, in favor of the Handy recording on Mac.
- Then tested it on my fixed version: played music on my iphone with output to my airpods, and then triggered Handy on my Mac. The music kept playing on the AirPods connected to the iPhone, while Handy handled the recording with the system's microphone.

In case it matters, the white noise app I use is called "Ooooo" on the mac app store, and it is what helped me pinpoint the source of the issue.
## AI Assistance

**If AI was used:**

- Tools used: Github copilot
- How extensively: helped me in finding the file where the issue was and in debugging

### DIFF
--- a/src-tauri/src/managers/audio.rs
+++ b/src-tauri/src/managers/audio.rs
@@ -2,6 +2,10 @@ use crate::audio_toolkit::{list_input_devices, vad::SmoothedVad, AudioRecorder, 
 use crate::helpers::clamshell;
 use crate::settings::{get_settings, AppSettings};
 use crate::utils;
+#[cfg(target_os = "macos")]
+use cpal::traits::DeviceTrait;
+#[cfg(target_os = "macos")]
+use cpal::traits::HostTrait;
 use log::{debug, error, info};
 use std::sync::atomic::{AtomicU64, Ordering};
 use std::sync::{Arc, Mutex};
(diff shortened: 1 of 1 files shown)


cjpais:
@VirenMohindra curious what you think

I haven't reviewed in depth. Feels a bit like a hack, but also maybe worth it for AirPods devices which are super common?

PR #1116, opened 2026-03-22
[BUGFIX] Fix model speed/accuracy indicator alignment

AlexanderYastrebov:
## Before Submitting This PR

**Please confirm you have done the following:**

- [x] I have searched [existing issues](https://github.com/cjpais/Handy/issues) and [pull requests](https://github.com/cjpais/Handy/pulls) (including closed ones) to ensure this isn't a duplicate
- [x] I have read [CONTRIBUTING.md](https://github.com/cjpais/Handy/blob/main/CONTRIBUTING.md)

**If this is a feature or change that was previously closed/rejected:**

- [ ] I have explained in the description below why this should be reconsidered
- [ ] I have gathered community feedback (link to discussion below)

## Human Written Description

* increase indicator width to fit German translation
* use text-end for proper alignment of Arabic translation
* use ms-4 instead of ml-4 to fix score bar alignment

## Related Issues/Discussions

Fixes #1109

## Screenshots/Videos (if applicable)

image

image

## AI Assistance

- [ ] No AI was used in this PR
- [x] AI was used (please describe below)

**If AI was used:**

- Tools used: Copilot with Claude Sonnet 4.6
- How extensively: to fix RTL styles

PR #1155, opened 2026-03-26
Add Hebrew language support

tomerar:
Adds full Hebrew localization support.

Changes:
- adds Hebrew locale
- adds Hebrew to supported app languages
- enables RTL support for Hebrew

Verified:
- `bun run build`
- Hebrew locale keys match English


AlexanderYastrebov:
RTL should be ok, I fixed some alignments within https://github.com/cjpais/Handy/pull/1116
Please check that translations fit into the UI, there might be some adjustments needed.

tomerar:
Thanks, I checked the UI in Hebrew and it looks good overall.

I went through the main screens:
- General
- Models
- Advanced
- History
- About

I did not notice clipped text, narrow buttons, awkward label wrapping, RTL layout issues, or untranslated English text in those screens.

PR #1166, opened 2026-03-27
fix Parakeet split retry for long recordings

ferologics:
## Summary
- restore Parakeet recovery for long recordings by moving chunked retry into a dedicated recovery helper
- retry only the current recoverable ONNX broadcast-shape inference failure
- preserve structured transcription results through recovery and merge chunk outputs at the seam instead of flattening to text immediately
- cover the retry driver with scenario tests for recovery, fail-fast behavior, and bounded retry exhaustion

## Why
Long Parakeet recordings can still fail with a recoverable inference error like:

```text
BroadcastIterator::Init ... Attempting to broadcast an axis by a dimension other than 1
```

That currently bubbles up as a hard failure instead of retrying on smaller chunks.

## Validation
- `cd src-tauri && cargo check`
- validated through Handy's normal History re-transcribe flow on `~/Library/Application Support/com.pais.handy/recordings/handy-1774637386.wav`

## Follow-up notes
- seam merging is still heuristic and would be stronger with word-level timestamps during recovery
- once this stabilizes, the recovery helper should move down into `transcribe-rs`

### DIFF
--- a/src-tauri/src/managers/mod.rs
+++ b/src-tauri/src/managers/mod.rs
@@ -2,3 +2,4 @@ pub mod audio;
 pub mod history;
 pub mod model;
 pub mod transcription;
+mod transcription_recovery;
--- a/src-tauri/src/managers/transcription.rs
+++ b/src-tauri/src/managers/transcription.rs
@@ -1,3 +1,7 @@
+use super::transcription_recovery::{
+    transcribe_with_chunk_retry, ChunkRetryPolicy, ChunkedTranscriptionResult,
+    ChunkedTranscriptionSegment,
+};
 use crate::audio_toolkit::{apply_custom_words, filter_transcription_output};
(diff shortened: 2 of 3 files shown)


cjpais:
Look at my PR for chunking

ferologics:
I tested #1173 's chunked Parakeet path against the original failing long recording (`handy-1774637386.wav`), and it completed successfully.

Given that, I think #1173 is the better direction to land:
- #1173 prevents the failure by chunking long audio before inference
- #1166 adds reactive split/retry recovery after a Parakeet inference failure

So I'm closing this in favor of #1173 rather than trying to merge two competing fixes for the same issue.

The fallback/recovery idea from this PR may still be worth revisiting later, ideally lower in `transcribe-rs`, if chunked paths still prove brittle.

PR #1167, opened 2026-03-27
Preserve legacy portable installs during updates

ferologics:
## Summary
- preserve portable mode during updater-driven installs when the target directory still uses Handy's legacy empty `portable` marker
- keep the existing magic-string check for current portable installs
- limit the auto-detection path to `/UPDATE` runs

## Validation
- no practical automated coverage for the NSIS updater path
- code change is isolated to `src-tauri/nsis/installer.nsi`

### DIFF
--- a/src-tauri/nsis/installer.nsi
+++ b/src-tauri/nsis/installer.nsi
@@ -587,16 +587,20 @@ Function .onInit
 
 
   ; --- PORTABLE MODE --- Auto-detect portable mode during updates.
-  ; If the target directory already has a valid portable marker file, preserve
-  ; portable mode so the Tauri updater works without needing /PORTABLE.
-  ; We validate the magic string to avoid false-positives from stale empty files
-  ; left by scoop's NSIS extraction (dl.7z side-effect).
+  ; Preserve portable installs that use either the current magic-string marker
+  ; or the legacy empty marker created by older Handy releases. Require Data/
+  ; for the legacy empty-marker case so stale scoop side-effect files do not
+  ; accidentally opt an updater run into portable mode.
   ${If} $PortableMode <> 1
+  ${AndIf} $UpdateMode = 1
   ${AndIf} ${FileExists} "$INSTDIR\portable"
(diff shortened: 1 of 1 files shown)


cjpais:
@back1ply

back1ply:
Hey @ferologics & @cjpais , for this one i think fills a gap. I missed that the NSIS installer runs before the Rust migration, so for in-app updater users the portable mode would already be lost by the time my migration code runs. Your fix is needed.
The ${AndIf} $UpdateMode = 1 guard handles the scoop false-positive concern well since scoop never passes /UPDATE. 

The one narrow remaining risk: 
someone who installed via scoop (leaving an empty portable file) and then updates via the Tauri in-app updater rather than scoop update. 
We could tighten that by also checking ${DirExists} "$INSTDIR\Data" (i know so robust) a real portable install would have a Data\ dir, scoop's side-effect file wouldn't. But that's optional, not a blocker.

Both changes are complementary and both needed.

ferologics:
Good call — I took that tightening.

The PR now only treats the legacy empty marker as portable during `/UPDATE` runs when `$INSTDIR\Data` also exists.

So the behavior is now:
- `Handy Portable Mode` marker -> preserve portable mode
- legacy empty marker + `Data/` present -> preserve portable mode
- legacy empty marker without `Data/` -> ignore it

That keeps the intended in-app updater fix while avoiding the scoop side-effect false positive you called out.

cjpais:
@back1ply @ferologics let me know if this is good to merge

ferologics:
should be good. i have not tested it manually but did a few passes with clanker and think it's in good place now. @back1ply wdyt?

back1ply:
> should be good. i have not tested it manually but did a few passes with clanker and think it's in good place now. @back1ply wdyt?

the $UpdateMode = 1 + Data/ guard is the right combo it covers the in-app updater path without the scoop false-positive. should be good to go, if you having doubts we can go extra mile and try a test build.

PR #1168, opened 2026-03-27
Serialize history retries and preserve legacy post-processing

ferologics:
## Summary
- serialize `TranscriptionManager::transcribe()` so history retries cannot race with other active transcriptions while the loaded engine is temporarily removed from the engine slot
- preserve legacy history retry behavior by treating existing `post_process_prompt` / `post_processed_text` rows as post-processing requests
- backfill `post_process_requested` for legacy rows during migration
- add coverage for the retry inference helper and migration backfill

## Validation

### DIFF
--- a/src-tauri/src/commands/history.rs
+++ b/src-tauri/src/commands/history.rs
@@ -81,6 +81,8 @@ pub async fn retry_history_entry_transcription(
         return Err("Recording has no audio samples".to_string());
     }
 
+    let should_post_process = entry.should_post_process_on_retry();
+
     transcription_manager.initiate_model_load();
 
     let tm = Arc::clone(&transcription_manager);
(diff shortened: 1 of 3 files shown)


ferologics:
I dug through this more carefully and I don’t think we should keep carrying this as-is.

The underlying edge cases seem real:
- history retry bypasses the normal coordinator path, so it can race other transcription calls
- legacy rows created before `post_process_requested` existed can retry without preserving the old post-process intent

But after validating the bigger picture, this feels like the wrong place/time to land it:
- `#1173` is the change that actually fixes the long-audio Parakeet failure that prompted all of this
- these `#1168` cases look pretty edge-case by comparison
- `#1173` is already restructuring the surrounding transcription path, so if we ever want to fix this, it should probably be done on top of that shape instead of landing this separate patch now
- I’m also not excited about carrying an extra history migration here unless we prove we really need it

So I’m going to close this one rather than spend more time polishing a side-path fix right now.

If this still bites later, I’d revisit it as a much smaller follow-up on top of `#1173`’s engine-access/history-retry path.

PR #1169, opened 2026-03-27
Surface model download failures immediately

ferologics:
## Summary
- surface model download failures directly from the frontend store instead of depending on a backend `model-download-failed` event race
- clear transient download / verification / extraction state on immediate failures
- show a toast and persist the error message when the invoke fails or returns an error
- keep onboarding's failure handling aligned with the centralized store behavior

## Validation
- `bun run build`

### DIFF
--- a/src-tauri/src/commands/models.rs
+++ b/src-tauri/src/commands/models.rs
@@ -24,23 +24,14 @@ pub async fn get_model_info(
 #[tauri::command]
 #[specta::specta]
 pub async fn download_model(
-    app_handle: AppHandle,
+    _app_handle: AppHandle,
     model_manager: State<'_, Arc<ModelManager>>,
     model_id: String,
 ) -> Result<(), String> {
-    let result = model_manager
+    model_manager
         .download_model(&model_id)
         .await
-        .map_err(|e| e.to_string());
-
-    if let Err(ref error) = result {
-        let _ = app_handle.emit(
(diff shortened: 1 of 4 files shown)


src-tauri/src/commands/models.rs
```diff
@@ -24,23 +24,14 @@ pub async fn get_model_info(
 #[tauri::command]
 #[specta::specta]
 pub async fn download_model(
-    app_handle: AppHandle,
+    _app_handle: AppHandle,
```
VirenMohindra, line 27: if we're not using something we prob just kill it from the signature right

cjpais:
im a bit confused, isn't the backend the source of truth for the download?

ferologics:
Closing this.

I went back through the actual source path and I don't think this PR is fixing a real bug in the form I claimed.

What the code already does on `main`:
- backend `download_model` is already the source of truth for download failure
- backend already emits `model-download-failed`
- `modelStore` already listens for that event, clears state, and shows a toast

So the broad "first-run model download failures can be silent" claim was not actually proven here. Clanker did a thing, and I was wrong to turn that into a PR without validating the bug properly first.

If there's anything real in this area, it's only the much narrower JS/invoke-failure catch path in the store, and this PR is not the right fix for that.

So I'm closing this rather than pushing a phantom-bug fix.

PR #1173, opened 2026-03-28
chunking/streaming revisited

cjpais:


ferologics:
Local validation note: I tested this against the original long Parakeet failure case on my machine (`handy-1774637386.wav`, 1163.58s / 18,617,280 samples) using the same chunked batch transcription path/config this PR adds for long history retries.

Result: it completed successfully with `parakeet-tdt-0.6b-v3` instead of hitting the previous broadcast-shape inference failure.

On my machine that run finished in ~48.5s end-to-end for the transcription step.

So at least for the failure that prompted my earlier `#1166`, this looks like it resolves the issue in practice.

cjpais:
Closing this for now, as we are moving to transcribe.cpp. This will be revisited again in the future when we properly support file based transcription

PR #1177, opened 2026-03-28
Load the VAD in parallel to ASR model

jacksongoode:
## Before Submitting This PR

## Human Written Description

This loads the VAD in parallel to the ASR model load. The reason being that while the VAD model load is tiny (~50ms on an M1 Pro) it is blocking in the `start_microphone_stream` where it really matters that we open the stream as fast as possible.

We don't necessarily need that second VAD load in the stream, but just in case it doesn't load initially it's there to block (but should return immediately).

## Testing

I tested with timers around the start stream to confirm the 50ms block vanished. I could add a test to validate this benchmark but that might be outside the scope of this small PR. In fact, I really think stepping perf timers around the Handy initialization/loading timeline should be a priority.

## AI Assistance

**If AI was used:**

- Tools used: A open-weight LLM
- How extensively: Discussion and design + benchmark validation


cjpais:
First I want to say thanks for this, and also agree with adding timers in and around critical pathways.

I was recently considering this as I was implementing batching, and I was wondering if it makes sense to always keep the VAD in memory? Mostly because it's a handful of MB (like 4?) I think it's probably worth it to keep in memory, and just throw data at it when necessary. Curious your thoughts

jacksongoode:
Yeah, it's pretty small, but if the time to load it is just 50ms. The CPU trade-off is probably negligible to load as well? If it took longer to load it it would make sense to keep it in memory.

I imagine for some users Handy be remain opened for entirety of a user's session. Given this, I'd say we should always take the opportunity to make Handy as quiet and small as possible.

cjpais:
Fair enough, appreciate the opinion, will take a look but this will just get merged

PR #1185, opened 2026-03-28
fix(tray): set tray tooltip text on Windows

### LINKED ISSUE #1184: [BUG] Tray menu tooltip is empty
## Bug Description

Image

When I hover over the tray menue icon, there is an empty tooltip.  Possible tooltips:
- Handy
- Handy - model name (moste useful in my eyes)
- Handy v0.8.0

## System Information

**App Version:**

v0.8.0

**Operating System:**

Windows 11

tomerar:
Fixes #1184.

The tray icon was created and refreshed without setting a tooltip, which caused Windows to show an empty hover tooltip.

This change sets the tooltip when building the tray icon and refreshes it during tray menu updates using the existing version label.

## Before Submitting This PR

## Description

I noticed that hovering the Handy tray icon on Windows showed an empty tooltip bubble instead of useful text. This looked like a small but clear UI bug, so I checked the tray setup and found that the icon was never given a tooltip string. I fixed it by setting the tooltip both when the tray icon is created and when the tray menu is refreshed.

## Community Feedback

Issue: #1184

## Testing

Tested locally on Windows 11.

## AI Assistance

### DIFF
--- a/src-tauri/src/lib.rs
+++ b/src-tauri/src/lib.rs
@@ -201,6 +201,7 @@ fn initialize_core_logic(app_handle: &AppHandle) {
             )
             .unwrap(),
         )
+        .tooltip(tray::tray_tooltip())
         .show_menu_on_left_click(true)
         .icon_as_template(true)
         .on_menu_event(|app, event| match event.id.as_ref() {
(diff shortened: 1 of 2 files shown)


cjpais:
Thanks for noticing this issue and providing the fix!

PR #1203, opened 2026-04-02
try ort cuda

cjpais:


cjpais:
Closing this as we're going to be moving to transcribe.cpp

Issue #1201, opened 2026-04-01
[BUG] Post-processing returns LLM system instructions instead of cleaned transcript


r2d2inlove:
## Before You Submit
Searched existing issues. #959 is related (transcript not included in prompt) but was closed as user error (missing `${output}`). My prompt correctly includes `${output}`, so this is a different issue.

## Bug Description

With post-processing enabled, Handy occasionally pastes the LLM's description of what it *would* do instead of the actual cleaned transcript.

**What I said:** "Can you summarise what was done for me, please?"

**What Handy's UI showed (correct transcript):** "Can you summarise what was done for me, please?"

**What was pasted into the target app (wrong — LLM meta-response):**
> I have cleaned the provided transcript. This involved correcting spelling, capitalization, and punctuation. Number words have been converted to digits (e.g., twenty-five to 25). Spoken punctuation has been replaced with symbols (e.g., 'period' to '.'). Filler words and repeated words have been removed. The original language has been preserved, and the text has been broken into sensible paragraphs. The meaning and word order remain the same.

The LLM is clearly responding to the *instructions* in the prompt rather than processing the transcript. This suggests `${output}` may not be getting substituted with the actual transcription text in some cases, even though it is present in the prompt.

**My post-processing prompt:**
```
Clean this transcript:
1. Fix spelling, capitalization, and punctuation errors
2. Convert number words to digits (twenty-five → 25, ten percent → 10%, five dollars → $5)
3. Replace spoken punctuation with symbols (period → ., comma → ,, question mark → ?)
4. Remove filler words (um, uh, like as filler, as well as repeated words (e.g. do do do do do)
5. Keep the language in the original version (if it was french, keep it in french for example)
6. Break into paragraphs when it's sensible.

Preserve exact meaning and word order. Do not paraphrase or reorder content.

Return only the cleaned transcript.

Transcript:
${output}
```

**Post-processing provider:** OpenRouter (`google/gemini-2.5-flash-lite`)

## System Information
**App Version:** v0.8.1
**Operating System:** macOS (Sequoia)
**CPU:** Apple M5
**GPU:** Apple M5 integrated

cjpais:
Pretty sure this is still user error and is how the model has chosen to respond. You probably should use a different model, tweak the prompt, etc. This is an alpha (experimental) feature currently, bugs are expected and low priority to be fixed.

[...]

PR #1221, opened 2026-04-04
perf: add reasoning_effort passthrough to avoid thinking-mode latency in local models

luoxi:
## Before Submitting This PR

- [x] I have searched existing issues and pull requests (including closed ones)
      to ensure this isn't a duplicate

## Human Written Description

I use Handy with Ollama for post-processing, and recently switched to
Gemma 4 because it handles mixed Japanese/English really well. The
problem is that Gemma 4 (and most recent open-weight models like Qwen 3)
ships with thinking mode enabled by default — so even for simple
transcript cleanup, the model spends a long time "reasoning" about
how to fix punctuation before it actually responds.

This hits all languages, but it's especially bad for Japanese — the
thinking phase generates far more tokens due to less efficient
tokenization. Post-processing matters more for Japanese too, since
Whisper is significantly weaker at mixed Japanese/English input,
making LLM cleanup essential rather than optional.

The fix is a one-field passthrough: the OpenAI-compatible API already
has a `reasoning_effort` parameter for this. Setting it to `"none"`
brings response time down to under a second with no observable quality
difference for transcript cleanup tasks.

Per [discussion with @cjpais](https://github.com/cjpais/Handy/pull/1221#issuecomment-4188436535),
this PR keeps things simple: **Custom provider always sends
`reasoning_effort: "none"`** during post-processing, with no UI control.
Post-processing rarely benefits from reasoning, and if a backend doesn't
recognize the parameter it's typically ignored. Cloud providers are
untouched — they handle this server-side or may reject unknown params.

Related: #1201 — @cjpais noted that thinking tokens should be stripped.
This prevents them from being generated in the first place (for the
Custom provider path).

### Benchmark (Gemma 4 E4B, Apple Silicon, Ollama)

| Test case | Thinking ON | Thinking OFF |
|-----------|------------|-------------|
| Japanese (tech, mixed JP/EN) | 12.7s | 1.2s |
| Japanese (short command) | 19.0s | 0.4s |
| English (tech) | 12.4s | 1.7s |
| Japanese (CLI terms) | 18.3s | 0.8s |

Output quality was identical across all test cases.

### Changes

- Add optional `reasoning_effort` field to `ChatCompletionRequest`
  (omitted when unset, per OpenAI spec)
- Thread through `send_chat_completion` / `send_chat_completion_with_schema`
- `actions.rs`: always set `reasoning_effort: "none"` for Custom provider
  post-processing; `None` for all other providers
- No new settings, no UI, no i18n strings

### Known limitations

A couple of edge cases worth flagging — none are blockers for the
Ollama/Gemma target, but may surface later:

1. **`"none"` is an Ollama extension.** The OpenAI spec only defines
   `minimal | low | medium | high`. If a user points Custom at
   OpenAI-proper with an o-series reasoning model, the request would
   be rejected. This is an edge case (o-series via Custom for
   post-processing is unusual), and we can add a fallback if anyone
   hits it.
2. **Some models use prompt-based reasoning control, not params.**
   Qwen 3 (`/no_think`), Anthropic extended thinking, and some
   fine-tunes expect instructions in the system prompt rather than
   an API field. For those, `reasoning_effort` is silently ignored
   and users would need to disable thinking inside their prompt
   template. Not something this PR can solve generically.
3. **No escape hatch to re-enable reasoning on Custom.** If someone
   genuinely wants reasoning during post-processing with a local
   model, they'd need to change the source. Acceptable tradeoff
   given this is alpha and post-processing is a narrow use case.

## Community Feedback

Issue: #1201

## Testing

Tested locally on macOS (Apple Silicon) with Ollama + Gemma 4 E4B.
Verified `reasoning_effort: "none"` is sent for Custom provider, other
providers receive no `reasoning_effort` field, and post-processing
latency drops to sub-second on Gemma 4 E4B.

## AI Assistance

**If AI was used:**

- Tools used: Claude Code
- How extensively: Research and implementation guidance. Manual testing
  and verification done by me.


cjpais:
Could we simplify this further and just default to None. Or off? I don't think anyone really wants reasoning for post processing as far as I can tell

luoxi:
> Could we simplify this further and just default to None. Or off? I don't think anyone really wants reasoning for post processing as far as I can tell

Thanks, that makes sense — I agree simpler is better here.

luoxi:
Simplified to a single "Disable reasoning" toggle, default ON for custom providers (sends reasoning_effort: "none"), as reasoning doesn’t add much value in post-processing.

Screenshot 2026-04-05 at 10 35 38

cjpais:
I think since this is an alpha feature, we can just disable reasoning in total to simplify the settings. Let's hit issues directly if reasoning disabled causes issues. I think for this app it should have little to no impact

luoxi:
Yeah that makes sense.

I do think reasoning control might get a bit inconsistent across models over time (params vs prompt-based control, etc.), so we might eventually want a bit more flexibility depending on provider behavior.

But agree we shouldn’t overcomplicate it now — I’ll update the PR to just disable reasoning for post-processing.

luoxi:
Done — removed the toggle entirely. Custom provider now always sends `reasoning_effort: "none"` for post-processing.

One small note: `"none"` is an Ollama-specific extension, so some strict OpenAI-compatible endpoints might not accept it. Probably not an issue for now, but happy to add a fallback if needed later.

cjpais:
Thank you! I will test this tomorrow or Tuesday and pull it in. This is a good change, so thank you for making it

cjpais:
Tested this with Qwen 9B on my machine, went from infinite processing to basically instant, this is a big win. Thanks

CruxExperts:
This feature completely breaks my OpenRoute API proxy setup. I route all my LLM API through openroute proxy inluding local, proxy is using OpenAI API, the forced "none" param value from Handy client is causing failed API response:

Error
[400]: Unsupported value: 'reasoning_effort' does not support 'none' with this model. Supported values are: 'minimal', 'low', 'medium', and 'high'.

Forcing custom provider to transmit any params without ability to override and or toggle params and values is a mistake, this breaks many use cases besides narrow local Ollama support.

cjpais:
You can use older versions for now. Sometimes you don't learn about unintended consequences until they happen.

Not sure why the proxy has to throw an error. It can just continue without erroring.

You're welcome to make a proper fix. This is anyhow a clearly labeled experimental feature and you should expect it to be not perfect.

luoxi:
@cjpais Following up on CruxExperts's report — that 400 is exactly the edge case I mentioned earlier:

> `"none"` is an Ollama extension. The OpenAI spec only defines `minimal | low | medium | high`.

I should have added a fallback from the start — sorry for the oversight.

Two options for a follow-up — curious which you'd prefer:

**A. Transparent in-memory fallback (no UI)**

On the first `400` mentioning `reasoning`, drop the reasoning field(s) and retry once. Cache the rejection per `provider.id + base_url` for the rest of the process so subsequent requests skip reasoning entirely.

- No UI changes; Ollama/Gemma users keep the speedup.
- OpenAI/proxy users pay one failed request per app launch, then run normally.

One caveat: `src-tauri/` currently doesn’t have a clear pattern for mutable in-memory caches — `Lazy` is used for read-only, `OnceLock` is write-once, and other mutable state lives in Tauri-managed `Arc<Mutex/RwLock<T>>`.

So this would introduce one of:

- **a1)** `Lazy<RwLock<HashSet>>` local to `llm_client.rs` — smallest diff, new pattern  
- **a2)** `LlmClientCache` as Tauri-managed state — more consistent, larger diff  
- **a3)** persist via `tauri-plugin-store` — probably overkill for alpha  

Leaning **a1** for simplicity, but I’d appreciate your guidance here since I’m less familiar with the preferred state management patterns in this part of the Rust/Tauri codebase.

**B. Bring back the "Disable reasoning" toggle**

Restore `post_process_disable_reasoning: bool` from the earlier revision of this PR — default ON, under Custom/OpenRouter.

This gives explicit user control, but it would effectively revert the earlier simplification decision (where we removed the toggle based on the idea that reasoning can just be disabled entirely for this alpha feature).

cjpais:
I think for now let's do B. I think it would be nice to simplify this in the future but we can take it up as we move post processing out of experimental. I don't love having this, but it's experimental alpha/for power users so it's fine for now

luoxi:
👍

There’s still no real standard for reasoning/thinking across providers — params, semantics, and even control surface (API vs prompt) all vary — so it’s a bit tricky to handle cleanly on Handy’s side.

Should be easier to revisit once things settle a bit.

I’ll open a follow-up PR shortly.

cjpais:
Yeah it's quite unfortunate the state of how the providers are handling this. Everyone wants their own standard and it all depends on how they trained the model. So I guess we have to accept this for now.

luoxi:
@cjpais 
Hi, I opened #1350 with the option-B toggle restoration

cjpais:
Thanks give me a few days I'm really busy

PR #1226, opened 2026-04-05
nix: set ALSA_PLUGIN_DIR via symlinkJoin

yuxqiu:
On non-NixOS with home-manager, Handy can error when ALSA_PLUGIN_DIR is set to a colon-separated list, e.g.
/nix/store/...-pipewire-1.6.2/lib/alsa-lib:/nix/store/...-alsa-plugins-1.2.12/lib/alsa-lib/libasound_module_pcm_pulse.so
Even though libasound_module_pcm_pulse.so exists in one of those dirs, ALSA treats the combined string as a single path and fails.
Use a symlinkJoin to present a single directory that contains both plugin trees.

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -79,6 +79,13 @@
             ];
           };
           lib = pkgs.lib;
+          combinedAlsaPlugins = pkgs.symlinkJoin {
+            name = "combined-alsa-plugins";
+            paths = [
+              "${pkgs.pipewire}/lib/alsa-lib"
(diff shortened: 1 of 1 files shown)


cjpais:
@xilec @y0usaf @kakapt

xilec:
LGTM! Tested locally on NixOS — Handy starts correctly, microphone is detected, and transcription works as expected. The symlinkJoin approach is clean and correct.

cjpais:
Thank you both!

PR #1229, opened 2026-04-05
fix: reject near-silent audio to prevent Whisper hallucinations

egsok:
## Summary

- Add RMS energy gate (threshold 0.005) in the transcription pipeline — skips audio that passed VAD but is too quiet for real speech
- Add minimum speech duration check (100ms / 1600 samples) in audio manager — discards tiny VAD leakage fragments before they get zero-padded to 1.25s

## Problem

When recording a few seconds of "silence", Whisper hallucinates text like *"Subtitles by the Amara.org community"*. Real microphones pick up ambient noise that occasionally exceeds VAD's threshold, causing SmoothedVad to flush its prefill buffer (~8000+ samples of near-silence). The padding logic then inflates this to 1.25 seconds, and Whisper hallucinates on it.

image

## How it works

**Primary defense — RMS energy gate** (`transcription.rs`): Computes RMS of the audio buffer and rejects anything below 0.005. Mic self-noise is ~0.0001–0.001, ambient noise that fools VAD is ~0.001–0.005, while whispered speech starts at ~0.01. This catches all near-silent audio regardless of how it got through VAD.

**Secondary defense — min duration** (`audio.rs`): Fragments shorter than 100ms (1600 samples at 16kHz) are discarded as VAD leakage — SmoothedVad's minimum real output is ~8000 samples. A cheap safety net for edge cases.

## Test plan

- [ ] Record 2–3 seconds of silence → no transcription output, no hallucination
- [ ] Record a short word (~0.3s) → transcribes normally
- [ ] Record normal speech → identical behavior to before
- [ ] Debug mode (`Ctrl+Shift+D`): "Audio RMS ... below silence threshold" logged when recording silence

### DIFF
--- a/src-tauri/src/managers/audio.rs
+++ b/src-tauri/src/managers/audio.rs
@@ -465,8 +465,12 @@ impl AudioRecordingManager {
 
                 // Pad if very short
                 let s_len = samples.len();
-                // debug!("Got {} samples", s_len);
-                if s_len < WHISPER_SAMPLE_RATE && s_len > 0 {
+                const MIN_SPEECH_SAMPLES: usize = 1600; // 100ms at 16kHz
+                if s_len < MIN_SPEECH_SAMPLES {
+                    // Too short to be real speech — SmoothedVad minimum real output is ~8000 samples
+                    // (15 prefill + 2 onset + 15 hangover frames). Anything shorter is leakage.
+                    Some(Vec::new())
+                } else if s_len < WHISPER_SAMPLE_RATE {
                     let mut padded = samples;
                     padded.resize(WHISPER_SAMPLE_RATE * 5 / 4, 0.0);
                     Some(padded)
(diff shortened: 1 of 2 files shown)


cjpais:
I think this is probably a good change, I will test it and pull it in. May need to pull in beta testers too to get more feedback on it to ensure no unintended consequences. But I think if there is very low RMS there should be no big problem

cjpais:
Hmmm.. I need to think more on this. I am not convinced that we can ship this broadly to everyone without consequences

cjpais:
I think if this is going to be merged we need a tunable param in debug, we cannot have a constant here. We need a way out if it causes bugs in prod

cjpais:
I'm going to close this for now just because I honestly don't see that many reports of hallucinations, but I am kind of still open to fixing this. I don't know if people enough complain about this. Let's reopen it. I just am hesitant to affect a common hot path right now. For me, I get almost no hallucinations on my machine, and I have been using Whisper as my daily driver previously for the past month or two before parakeet unified landed

PR #1230, opened 2026-04-05
feat: preserve non-text clipboard content during paste-and-restore

egsok:
## Problem

When using clipboard-based paste methods (Ctrl+V, Shift+Insert, etc.), Handy saves the clipboard, writes transcribed text, sends the paste keystroke, then restores the original clipboard content. Currently it only handles **plain text** — `read_text()` returns an empty string for anything else.

This means if you had a **screenshot**, **copied files**, or **rich HTML** in the clipboard before transcribing, it was silently replaced with an empty string. Your clipboard content was lost.

## Solution

Use `arboard` (already a transitive dependency via `tauri-plugin-clipboard-manager`, so zero extra compile cost) to detect and restore all clipboard formats:

| Format | Before | After |
|--------|--------|-------|
| Plain text | ✅ | ✅ |
| Images (screenshots) | **lost** | ✅ |
| HTML (rich text from browser) | **lost** | ✅ |
| Files (copied from Explorer/Finder) | **lost** | ✅ |

Detection order is most-specific-first: files → image → HTML → text → empty. This prevents false matches (e.g., browser copies set both HTML and text fallback — text-first would lose formatting).

On Wayland, text restore still uses `wl-copy` for umlaut compatibility.

## Files changed

- `src-tauri/Cargo.toml` — add `arboard` as direct dependency
- `src-tauri/src/clipboard.rs` — add `SavedClipboard` enum + `save_clipboard()`/`restore_clipboard()`, replace text-only save/restore in `paste_via_clipboard()`

## Test plan

- [ ] Copy screenshot (Win+Shift+S / Cmd+Shift+Ctrl+4) → transcribe → verify screenshot is still in clipboard
- [ ] Copy plain text → transcribe → verify text is restored
- [ ] Copy rich HTML from browser → transcribe → paste into rich editor → verify formatting preserved
- [ ] Copy files in Explorer/Finder → transcribe → verify files still pasteable
- [ ] Empty clipboard → transcribe → clipboard stays empty
- [ ] Transcribed text still pastes correctly into target app
- [ ] `CopyToClipboard` setting still works (leaves transcribed text in clipboard, no restore)

### DIFF
--- a/src-tauri/Cargo.lock
+++ b/src-tauri/Cargo.lock
@@ -2416,6 +2416,7 @@ name = "handy"
 version = "0.8.2"
 dependencies = [
  "anyhow",
+ "arboard",
  "chrono",
  "clap",
  "cpal",
--- a/src-tauri/Cargo.toml
+++ b/src-tauri/Cargo.toml
@@ -72,6 +72,7 @@ sha2 = "0.10"
 transcribe-rs = { version = "0.3.8", features = ["whisper-cpp", "onnx"] }
 handy-keys = "0.2.4"
 ferrous-opencc = "0.2.3"
+arboard = { version = "3.6", features = ["image-data"] }
 clap = { version = "4", features = ["derive"] }
 specta = "=2.0.0-rc.22"
 specta-typescript = "0.0.9"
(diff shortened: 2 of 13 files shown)


egsok:
Recreating with a clean single-commit branch

PR #1231, opened 2026-04-05
feat: preserve non-text clipboard content during paste-and-restore

egsok:
## Problem

When using clipboard-based paste methods (Ctrl+V, Shift+Insert, etc.), Handy saves the clipboard, writes transcribed text, sends the paste keystroke, then restores the original clipboard content. Currently it only handles **plain text** — `read_text()` returns an empty string for anything else.

This means if you had a **screenshot**, **copied files**, or **rich HTML** in the clipboard before transcribing, it was silently replaced with an empty string. Your clipboard content was lost.

## Solution

Use `arboard` (already a transitive dependency via `tauri-plugin-clipboard-manager`, so zero extra compile cost) to detect and restore all clipboard formats:

| Format | Before | After |
|--------|--------|-------|
| Plain text | ✅ | ✅ |
| Images (screenshots) | **lost** | ✅ |
| HTML (rich text from browser) | **lost** | ✅ |
| Files (copied from Explorer/Finder) | **lost** | ✅ |

Detection order is most-specific-first: files → image → HTML → text → empty. This prevents false matches (e.g., browser copies set both HTML and text fallback — text-first would lose formatting).

On Wayland, text restore still uses `wl-copy` for umlaut compatibility.

## Files changed

- `src-tauri/Cargo.toml` — add `arboard` as direct dependency
- `src-tauri/src/clipboard.rs` — add `SavedClipboard` enum + `save_clipboard()`/`restore_clipboard()`, replace text-only save/restore in `paste_via_clipboard()`

## Test plan

- [ ] Copy screenshot (Win+Shift+S / Cmd+Shift+Ctrl+4) → transcribe → verify screenshot is still in clipboard
- [ ] Copy plain text → transcribe → verify text is restored
- [ ] Copy rich HTML from browser → transcribe → paste into rich editor → verify formatting preserved
- [ ] Copy files in Explorer/Finder → transcribe → verify files still pasteable
- [ ] Empty clipboard → transcribe → clipboard stays empty
- [ ] Transcribed text still pastes correctly into target app
- [ ] `CopyToClipboard` setting still works (leaves transcribed text in clipboard, no restore)


cjpais:
Curious, have you tested all of these, just want to make sure

I've tested a branch that tried this before without success so just curious. I am open to this for sure though

egsok:
@cjpais yes, sir! Tested every case (pics, html, files) on Win11.

authored a similar PR for Openwhispr recently (https://github.com/OpenWhispr/openwhispr/pull/381), Handy is even better suited for this because of use of arboard (so we get support for random files from explorer as well, unlike Openwhispr).

cjpais:
@egsok thanks for confirmation for Win11 testing, I will test on MacOS and get back

This definitely should be merged assuming it works well, it should address #921 

Bug fixes are always super helpful so thank you!

egsok:
@cjpais cool! let me know if anything needs fixing

cjpais:
One possible thing to consider, in case of regressions is adding a toggle to change back to the old method. Could be done in debug potentially. I will test this myself first, I would prefer not to have this path.

cjpais:
I feel a bit bad because I've scoped this work all the way back to just images, but it's a minimal fix that does seem to work, so I am merging it for now. If we feel we need to do better clipboard management in the future, we will handle it at that time

PR #1073, opened 2026-03-17
ci: reduce PR check time from ~30 min to ~1 min

VirenMohindra:
## Before Submitting This PR

- [x] I have searched [existing issues](https://github.com/cjpais/Handy/issues) and [pull requests](https://github.com/cjpais/Handy/pulls) (including closed ones) to ensure this isn't a duplicate
- [x] I have read [CONTRIBUTING.md](https://github.com/cjpais/Handy/blob/main/CONTRIBUTING.md)

## Human Written Description

every push to a PR triggers 5 workflows with no concurrency controls, no path filters, and no nix caching. a `README` change triggers rust tests + a 30 min nix build. rapid pushes trigger duplicate runs that pile up

this PR adds concurrency groups, path filters, merges redundant workflows, and adds nix store caching. for PRs that don't touch nix files (the majority), the critical path drops from ~30 min to ~23 sec

## Related Issues/Discussions

general CI/CD performance, no specific issue filed

## Testing

tested on this PR itself:

**frontend-only change (src/):**

| check | before | after |
|-------|--------|-------|
| lint | 14s | merged into code-quality |
| prettier | 18s | merged into code-quality |
| code-quality | n/a | **14s** |
| playwright | 23s | **23s** |
| rust test | ~2 min | **skipped** (path filter) |
| nix-check (quick) | ~30 min | **~1 min** (flake eval only, full build skipped) |
| **total** | **~30 min** | **~23 sec** |

**nix change (flake.nix):**

| check | before | after |
|-------|--------|-------|
| nix-build (cold cache) | ~29 min | ~30 min (same, cache populating) |
| nix-build (warm cache) | ~20 min | should improve slightly with magic-nix-cache |

concurrency groups confirmed, stale nix run was cancelled when new push arrived

### changes

- **merge lint + prettier into `code-quality.yml`**: eliminates one runner startup + duplicate `bun install`
- **concurrency groups on all PR workflows**: `cancel-in-progress: true` cancels stale runs
- **path filters**: `src/` triggers code-quality + playwright + nix quick checks, `src-tauri/` triggers rust test + nix quick checks, nix packaging files trigger the full nix build
- **two-tier nix strategy**: quick checks (bun.nix sync, flake eval) run on any source change; full nix build (~25 min) only runs when packaging files change (`flake.nix`, `flake.lock`, `.nix/**`, `bun.lock`, `Cargo.toml`, `Cargo.lock`)
- **nix store caching** via `DeterminateSystems/magic-nix-cache-action@v13`
[...]

PR #1237, opened 2026-04-06
feat(nix): add Cachix binary cache to CI

max-miller1204:
## Summary

- Replace `magic-nix-cache-action` (GitHub-local cache only) with `cachix/cachix-action` so build artifacts are pushed to a public Cachix binary cache
- Nix users can then pull pre-built binaries instead of compiling from source (~25 min build)

## Setup required

1. Create the cache: `cachix create handy`
2. Generate an auth token at https://app.cachix.org and add it as `CACHIX_AUTH_TOKEN` in the repo's GitHub Actions secrets
3. Once the cache exists, users add the substituter to their Nix config:
   ```nix
   nix.settings = {
     substituters = [ "https://handy.cachix.org" ];
     trusted-public-keys = [ "handy.cachix.org-1:<key from cachix create output>" ];
   };
   ```

The NixOS module (`programs.handy`) could also auto-add the substituter — happy to add that in a follow-up once the cache is created and the public key is known.

### DIFF
--- a/.github/workflows/nix-check.yml
+++ b/.github/workflows/nix-check.yml
@@ -50,7 +50,7 @@ jobs:
 
       - uses: cachix/cachix-action@v15
         with:
-          name: handy
+          name: handy-computer
           authToken: "${{ secrets.CACHIX_AUTH_TOKEN }}"
 
       # Regenerate .nix/bun.nix from bun.lock and check if it matches
--- a/.github/workflows/nix-check.yml
+++ b/.github/workflows/nix-check.yml
@@ -4,7 +4,7 @@
 #    so compilation-breaking edits are caught by flake eval.
 # 2. Full nix build (~25 min) only runs when nix packaging files change.
 #
-# Setting up a Cachix binary cache would further reduce full-build times.
+# Build artifacts are pushed to Cachix so Nix users can skip local compilation.
 
 name: "nix build check"
 on:
@@ -48,7 +48,10 @@ jobs:
         with:
           nix_path: nixpkgs=channel:nixos-unstable
 
-      - uses: DeterminateSystems/magic-nix-cache-action@565684385bcd71bad329742eefe8d12f2e765b39 # v13
+      - uses: cachix/cachix-action@v15
+        with:
+          name: handy
+          authToken: '${{ secrets.CACHIX_AUTH_TOKEN }}'
 
       # Regenerate .nix/bun.nix from bun.lock and check if it matches
       # what's committed. A diff means the developer forgot to run
@@ -121,7 +124,7 @@ jobs:
       # sandbox issues, compilation failures) that flake eval alone misses.
       # On PRs: only runs when nix packaging files change (~25 min with cold cache).
       # On push to main and workflow_dispatch: always runs so every commit on
-      # main has a verified nix build before release.
+      # main has a verified nix build. cachix-action auto-pushes artifacts.
       - name: Build handy
         if: steps.bun-check.outputs.outdated != 'true' && steps.eval.outputs.failed != 'true' && (steps.nix-files.outputs.changed == 'true' || github.event_name == 'workflow_dispatch' || github.event_name == 'push')
         run: nix build .#handy -L --show-trace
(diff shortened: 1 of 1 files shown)


cjpais:
@y0usaf @VirenMohindra @yuxqiu @xilec @kakapt

xilec:
LGTM! This would significantly reduce build times for Nix users — no more ~25 min compilations from source. I'm sure all Nix users will appreciate this.

One note: @cjpais, this PR requires setup on your side before merging:

1. `cachix create handy` — this will also generate the public key
2. Add the auth token as `CACHIX_AUTH_TOKEN` in repo secrets

Once the cache is created, it would be great to add the substituter config with the public key to README.md so Nix users know how to enable it:

```nix
nix.settings = {
  substituters = [ "https://handy.cachix.org" ];
  trusted-public-keys = [ "handy.cachix.org-1:<public key from step 1>" ];
};
```

cjpais:
I see. I didn't realize this is a paid hosting provider. I suspect the 5GB free tier might not be enough, anyone want to get me credits or a free plan? Happy to add them as a sponsor potentially.

xilec:
@cjpais The free 5GB tier should be more than enough. Here's why:

Cachix doesn't store one big blob — it caches each dependency separately. When a user downloads handy, Cachix only serves the pieces that aren't already available from the default Nix package mirror (`cache.nixos.org`). Most dependencies (gtk, pipewire, etc.) are already there.

So in practice, Cachix only needs to store handy itself (~67MB) and a few project-specific build artifacts. That's roughly 100-200MB per version — well within the 5GB free tier even with several versions cached. Old versions can be cleaned up with `cachix trim`.

VirenMohindra:
yep this is exactly what i was thinking and mentioned it in the follow up recommendation section in my CI PR https://github.com/cjpais/Handy/pull/1073

also spot on on the free tier being good enough, since cache will be doing most of the heavy lifting here

I'll take a quick peek at the PR when i have some time, thank you for the contribution!

.github/workflows/nix-check.yml
```diff
@@ -48,7 +48,10 @@ jobs:
         with:
           nix_path: nixpkgs=channel:nixos-unstable
 
-      - uses: DeterminateSystems/magic-nix-cache-action@565684385bcd71bad329742eefe8d12f2e765b39 # v13
+      - uses: cachix/cachix-action@v15
+        with:
+          name: handy
+          authToken: '${{ secrets.CACHIX_AUTH_TOKEN }}'
```
VirenMohindra: should add this before we merge it in

VirenMohindra (approved):
LGTM

cjpais:
pub key is: `handy-computer.cachix.org-1:Sihzctn6DC0CJM5QeL+9nBEL3CL8c33m777C+eIv748=`

I've added the env var, let me know if theres anything else todo

I see the build failed due to ENV var not set I will check later, might just be in this CI staging or something it didnt get picked up. I will check later.

max-miller1204:
Thanks for the review! Just pushed 9f80d6c to fix the Prettier formatting issue that was failing code-quality (single → double quotes around the CACHIX_AUTH_TOKEN reference). Let me know if you'd like me to adjust anything else.

EDIT: I think the ENV var issue your hitting is due to CACHIX_AUTH_TOKEN not being in the repo's GitHub Actions secrets yet. Also with the new pub-key name I had to switch it in my nix-check.yml. It should be good now!

xilec:
@cjpais can we get CI running on this PR? Would be good to see whether the Cachix setup actually works end-to-end.

xilec:
@cjpais ping

cjpais:
Give me a week or two on this. Right now I have very little time

I added the env var to GitHub prior to running the last CI and it still failed

y0usaf:
Please take your time and help out the (dozen or so of us) Nix users whenever you get around to it, there's no rush :bow: I am very grateful that Handy is able to support Nix as it is.

carterworks:
In the meantime, you can use [numtide/llm-agents.nix flake](https://github.com/numtide/llm-agents.nix), which included binary caching and handy
`nix run github:numtide/llm-agents.nix#handy -- --help`

Issue #1207, opened 2026-04-02
[BUG] App hangs on Advanced tab on fresh start


ferologics:
## Bug Description

- restart mac, 
- open Handy app
- click on Advanced section
- it hangs (see spinner wheel)
- recovers after ~5 seconds

## System Information

**App Version:**

0.8.1

**Operating System:**

macos 26.3.1
macos 26.4.0

**CPU:**

M1 Pro

**GPU:**

M1 Pro

## Logs

[tbd]

didn't check / see any. maybe somebody else sees some

cjpais:
ty for the report!

cjpais:
If you can please test 

https://github.com/cjpais/Handy/pull/1246#issuecomment-4203235940

It should help resolve this

PR #1246, opened 2026-04-08
query gpu async

cjpais:
make the gpu query for whisper.cpp async

not sure why I let this go through sync to begin with.. my bad.

#1207


jborichevskiy:
working well on Tahoe 26.3.1

will test on Nvidia GPU as well

Ari4ka (approved):

PR #1252, opened 2026-04-08
Silence-based Chunked Transcription for long recordings with optional toggle

meetuljain:
This PR adds an optional chunked transcription mode for long recordings.

Instead of waiting until the user stops recording and then transcribing the entire audio at once, Handy can now detect silence, split the recording into chunks, and transcribe completed chunks in the background while recording is still in progress.

This significantly improves perceived speed for long dictation / monologue use cases.

I'm aware Handy is currently under feature freeze, so I want to call out that this is based on a previously requested use case:
https://github.com/cjpais/Handy/discussions/179

## Changes

- Added silence-based chunked transcription during recording
- Added a toggle in Advanced Settings
- Persisted the setting with `chunked_transcription_enabled`
- Preserved the original transcription flow when disabled

## Trade-offs

Chunking improves speed, but because chunks are transcribed independently, context can be lost at chunk boundaries and transcription quality may be slightly lower in some cases. That is why this feature is optional.

## Testing

- Tested long recordings with chunking enabled
- Confirmed earlier background transcription
- Confirmed legacy behavior when disabled
- Verified existing paste-error behavior remains intact

## Related discussion

- https://github.com/cjpais/Handy/discussions/179

## AI disclosure

AI used: Yes

Tools used:
- Codex
- ChatGPT

How it was used:
- Used Codex to understand the code structure
- Gave Codex pseudocode and it generated an initial implementation
- I then manually reviewed the generated code, identified bugs, fixed them, and refined/consolidated the implementation
- ChatGPT was used to help polish the PR text


cjpais:
#1173.

PR #1254, opened 2026-04-08
fix: add input channel selection for multi-channel audio interfaces

### LINKED ISSUE #591: [BUG] Allow to select channel of devcie to use
Many professional audio interfaces have many different channels, only one of which may carry the microphone input. On such interfaces, handy seems to merge all channels to mono, which results in an unusably low signal or captures unrelated signals. On such systems, handy is not usable.

phinze:
👋 Hello! Cool project. 😄 This is my first contribution so please do let me know if I missed anything!

## Human Written Description

I have a Focusrite Scarlett Solo 4th Gen as my audio interface. When I started using Handy, I noticed it was transcribing music playing through my headphones even though the music wasn't coming through the microphone. After digging into it, I found that the Scarlett exposes 4 input channels to macOS: channel 0 is the instrument input, channel 1 is the mic, and channels 2-3 appear to be loopback/output channels. Handy [averages all channels together to produce mono](https://github.com/cjpais/Handy/blob/c1697b2a95ccc0ad4913fdd3ab7fb035b6761c66/src-tauri/src/audio_toolkit/audio/recorder.rs#L256-L263), which means any audio on those loopback channels gets mixed into the transcription input.

I wanted to fix this in a way that doesn't change anything for people with simple setups. The solution adds an "Input Channel" dropdown that only appears in settings when the selected microphone has more than one channel. The default is "Average all channels" which preserves the existing behavior exactly. People with multi-channel interfaces can select their mic's specific channel to avoid bleed from other channels.

I briefly experimented with auto-detecting the mic channel by sampling peak levels across channels at stream start, but it's fragile (what if music is playing when you start recording?). A simple dropdown felt like the most reliable approach, and it matches what was suggested in #591.

## Related Issues/Discussions

Fixes #591

Also related to #998: users with Focusrite Scarletts report that "Mute While Recording" doesn't work because Focusrite drivers don't expose kAudioDevicePropertyMute. Channel selection sidesteps this by not capturing the output audio in the first place. It'd still be cool to also mute or [duck](https://github.com/cjpais/Handy/discussions/1190) the music while recording, but  that's a separate concern.

## Community Feedback

Multiple users in #591 confirmed the problem across different interfaces (Focusrite Scarlett, MOTU M4) and platforms (macOS, Windows). The suggested solution (a channel dropdown) came directly from a commenter in that thread:

> *"I would expect a dropdown under the device to select a single channel."*

@cjpais closed #591 due to bandwidth but explicitly said PRs are welcome:

> *"I will accept PR's which implement this"*

> *"PRs of all kinds are accepted"*

## Testing

Tested on macOS with my Scarlett Solo 4th Gen (4-channel device). With Channel 2 selected (the mic input), music playing through headphones no longer bleeds into transcription. "Average all channels" still works as before. The dropdown doesn't show up at all for single-channel devices like the built-in mic. Changing the channel takes effect immediately, no restart needed.

New dropdown:

2026-04-08 at 10 40 25@2x

Disappears with one channel device selected:

2026-04-08 at 10 40 35@2x

## AI Assistance

**If AI was used:**

- **Tools used**: Claude Code
- **How extensively**: I'm new to Rust, so I leaned on Claude Code pretty heavily here. It helped me trace the root cause down to the channel averaging logic in `recorder.rs`, wrote most of the implementation, and helped me get a Nix-based macOS build working (the upstream Nix flake is Linux-only). I drove the debugging process, decided on the approach (channel selector with "average all" as default), and tested everything on my Scarlett Solo.


cjpais:
Thanks this looks quite reasonable I will review when I can

phinze:
Just rebased this onto latest main (through v0.8.3) and tested the build locally on macOS. Input channel selection still works correctly with my multi-channel interface. No rebase conflicts, and the new tip commit is `0641e1a`. Happy to address any feedback whenever you have a chance to take a look.

cjpais:
Thanks, give me a bit of time, I am doing a big chunk of other work right now and need to focus on that for a bit. I will get around to merging this!

cjpais:
I just want to say thanks for this. I haven't necessarily been able to give it a test because I don't have multi-channel devices to test with. However, based on the review, this is a pretty well implemented fix, so thank you

PR #1258, opened 2026-04-08
fix(linux): use DOM keyboard events for shortcut recording to avoid rdev::grab() conflict with HotkeyManager

possiblyneal:
On Linux, `rdev::grab()` (used by the handy-keys backend recording loop) conflicts with the `HotkeyManager` — grabbing exclusive access to the keyboard causes periodic system freezes and no key events being captured during shortcut recording.

This PR fixes the issue by detecting Linux at runtime and falling back to DOM `keydown`/`keyup` events for shortcut recording instead of the backend `handy-keys-event` stream. The backend `startHandyKeysRecording` call is still made (so the backend can disable the active hotkey during recording), but key capture is handled entirely in the frontend on Linux.

## Changes

**`src/components/settings/HandyKeysShortcutInput.tsx`**
- Added a second `useEffect` that activates only on Linux (`osType === "linux"`)
- Builds the hotkey string from DOM `KeyboardEvent` modifiers + `getKeyName()`
- Commits on `keyup` (same flow as the non-Linux path)
- The existing `handy-keys-event` listener effect is still present but does not receive events on Linux since the backend skips the grab

**`src-tauri/src/shortcut/handy_keys.rs`**
- Wrapped `KeyboardListener::new()` in `#[cfg(not(target_os = "linux"))]` so the second `rdev::grab()` call is never made on Linux, eliminating the periodic 2-second freeze

### DIFF
--- a/src-tauri/src/shortcut/handy_keys.rs
+++ b/src-tauri/src/shortcut/handy_keys.rs
@@ -260,22 +260,11 @@ impl HandyKeysState {
     }
 
     /// Start recording mode for a specific binding
-    pub fn start_recording(&self, app: &AppHandle, binding_id: String) -> Result<(), String> {
+    pub fn start_recording(&self, _app: &AppHandle, binding_id: String) -> Result<(), String> {
         if self.is_recording.load(Ordering::SeqCst) {
             return Err("Already recording".into());
         }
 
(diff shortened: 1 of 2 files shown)


cjpais:
Is it possible to fix this directly in handy-keys?

PR #1270, opened 2026-04-11
Fix macOS typed Unicode modifiers in enigo

nialse:
## Summary
Fix the macOS `PasteMethod::Type` path when a typed Unicode character depends on a keyboard modifier on the active layout.

Handy was already sending `Key::Unicode('?')`, but enigo 0.6.1 only kept the resolved physical keycode and dropped the modifier that produced it. On layouts like `Swedish - Pro`, that meant `?` resolved to the `+/?` key and was emitted without Shift, so the target app received `+`.


cjpais:
10k lines for what exactly? yeah closing.

we dont just vendor libraries for fun, submit patches upstream if they are broadly applicable.

nialse:
The eventual PR was posted in [enigo](https://github.com/enigo-rs/enigo/pull/539).

Issue #1239, opened 2026-04-06
[BUG] Handy 0.8.2 completely fails on Wayland (overlay & hotkeys broken)


eyverg8:
## Before You Submit

**Please search [existing issues](https://github.com/cjpais/Handy/issues) to avoid duplicates.** Your bug may already be reported! Right now it's just me maintaining this project so many issues can be overwhelming! Help me out by checking first.

## Bug Description

Handy 0.8.2 works normally under Xorg, but **does not function at all on Wayland sessions**.  

- The overlay does not appear.  
- Global hotkeys do not work.  
- Running with `--debug` produces no logs.  

Tested with both the AppImage (`Handy_0.8.2_amd64.AppImage`) and the DEB package (`Handy_0.8.2_amd64.deb`). Likely related to Wayland restrictions on global hotkeys.

## System Information

**App Version:**  
0.8.2 (AppImage & DEB)  

**Operating System:**  
Ubuntu 22.04 / 23.10 (Wayland session)  

**CPU:**  
Varies (tested on AMD & Intel systems)  

**GPU:**  
Varies (tested on integrated and discrete GPUs)  

## Logs

No logs are produced when running with `--debug`.

cjpais:
Is this a regression? If not please look at the docs (README and Website) and existing issues. This is well documented

If it is a regression, please message here.

eyverg8:
# Fix: Handy Hotkeys on Ubuntu 25.10 (Wayland)

### The Issue
Wayland blocks Handy from "listening" to global hotkeys for security reasons.

### The Fix: Native System Shortcut
By using Ubuntu's built-in shortcut manager, you can bypass Wayland's restrictions.

1. Go to **Settings** > **Keyboard** > **View and Customize Shortcuts**.
2. Scroll to the bottom and select **Custom Shortcuts**.
3. Click the **+** (Add) button and enter:
   * **Name:** `Handy Toggle`
   * **Command:** `handy --toggle-transcription`
   * **Shortcut:** (Press your preferred keys, e.g., `Alt + Space`)

### Why it works
Instead of Handy watching your keyboard, Ubuntu triggers the command directly, which Wayland allows. This is the official workaround for **Handy v0.8.2** on modern Linux sessions.

Issue #1236, opened 2026-04-06
GlobalShortcuts portal broken on GNOME 49 - simpler workaround than #572


FarmLox:
## Problem

Global shortcuts don't work on GNOME 49 (Fedora 43, xdg-desktop-portal-gnome 49.0). Shortcuts register successfully ("Shortcuts initialized successfully" in the log) but pressing the keys produces no events at all. No recording starts, no log output, nothing.

This affects both Tauri Global Shortcut mode and is consistent with known xdg-desktop-portal-gnome bugs:
- https://gitlab.gnome.org/GNOME/xdg-desktop-portal-gnome/-/issues/183
- https://gitlab.gnome.org/GNOME/xdg-desktop-portal-gnome/-/issues/191

The GlobalShortcuts portal implementation in GNOME is roughly one year old (shipped in GNOME 48) and still has open bugs around non-Flatpak apps with empty `app_id`. Vibe has the same problem on this setup.

## Environment

- Fedora 43
- GNOME Shell 49.5
- Wayland session
- xdg-desktop-portal 1.20.3
- xdg-desktop-portal-gnome 49.0
- Handy v0.8.2 (RPM from GitHub releases)

## Additional issue with Handy Keys mode

Switching to "Handy Keys" (rdev) as a workaround caused severe desktop breakage:

```
rdev grab error: IoError(Os { code: 16, kind: ResourceBusy, message: "Device or resource busy" }), retrying in 2s
```

This loop of failed exclusive grabs made the GNOME session nearly unusable: Alt+Tab reversed and randomised, keys repeating or not registering, VS Code crashed. Closing Handy from the tray restored normality.

## Existing PR

#572 addresses this by creating GNOME custom shortcuts via `gsettings` that send `SIGUSR2` to the Handy process. That PR has merge conflicts and several review issues (unsafe `pkill -f` matching, premature shortcut save, etc.).

## Suggested simpler alternative

I run a similar dictation app that works on GNOME Wayland using a much simpler pattern:

**The app listens on a Unix socket for commands.** A second invocation of the same binary connects to the socket, sends a command ("toggle", "cancel", etc.), and exits. The GNOME custom shortcut just runs the binary with a flag:

```
# GNOME custom shortcut command:
handy --toggle
```

Implementation:
1. On startup, Handy opens a Unix socket (or even just a named pipe) at a known path
2. Add a `--toggle` CLI flag. When passed, it connects to the socket, sends "toggle", and exits immediately
[...]

Issue #949, opened 2026-03-03
Global shortcuts don't work on Wayland compositors (COSMIC, Sway, Hyprland)


accubotai:
## Problem

Tauri's global shortcut plugin does not work on Wayland compositors that don't support the `zwp_keyboard_shortcuts_inhibit_manager_v1` protocol. This affects:

- **COSMIC** (Pop!_OS 24.04+)
- **Sway**
- **Hyprland**
- Most other wlroots-based compositors

This is a Wayland security design — applications cannot grab global keyboard shortcuts directly. The compositor must mediate all global keybindings.

### What happens

- Handy starts normally, model loads, UI works fine
- The configured shortcut (e.g. `Ctrl+Space`) never fires — pressing it does nothing
- Both `keyboard_implementation: "tauri"` and push-to-talk mode are affected
- The `--toggle-transcription` CLI flag works perfectly, confirming the issue is isolated to shortcut capture

### Environment

- Pop!_OS 24.04 LTS, COSMIC desktop (Wayland)
- Handy v0.7.9 AppImage
- `WAYLAND_DISPLAY=wayland-1`, `XDG_SESSION_TYPE=wayland`

## Current workaround

We solved this with two pieces:

### 1. External push-to-talk daemon using evdev

A small Python script reads raw kernel input events from `/dev/input/` (bypassing Wayland entirely) and calls `handy --toggle-transcription` on key press and release:

```python
#!/usr/bin/env python3
"""Push-to-talk for Handy on Wayland via evdev."""
import select, subprocess, sys, signal, os, logging
import evdev
from evdev import ecodes, categorize, KeyEvent

KEY_SPACE = ecodes.KEY_SPACE
KEY_LEFTCTRL = ecodes.KEY_LEFTCTRL
KEY_RIGHTCTRL = ecodes.KEY_RIGHTCTRL

def find_keyboards():
    keyboards = []
    for path in evdev.list_devices():
        try:
            dev = evdev.InputDevice(path)
            caps = dev.capabilities()
            if ecodes.EV_KEY in caps:
                key_caps = caps[ecodes.EV_KEY]
                if KEY_SPACE in key_caps and KEY_LEFTCTRL in key_caps:
                    keyboards.append(dev)
        except (PermissionError, OSError):
            pass
    return keyboards

def toggle_handy():
    subprocess.Popen(["handy", "--toggle-transcription"],
                     stdout=subprocess.DEVNULL, stderr=subprocess.DEVNULL)

def main():
[...]

PR #1287, opened 2026-04-13
feat(linux): add xdg-desktop-portal GlobalShortcuts backend for Wayland

nmcbride:
Adds xdg-desktop-portal GlobalShortcuts as a third keyboard implementation option for Linux. This gives Wayland users proper global shortcut support with full press/release events, so push-to-talk actually works.

The existing Tauri and HandyKeys backends both depend on X11 APIs that don't work on Wayland (XGrab and rdev respectively), which has been an open issue for a while (#949, #1236, #1239). PR #572 took a workaround approach using gsettings + SIGUSR2 signals, but that only supports toggle mode and is GNOME-specific. The portal approach is the standard freedesktop mechanism and works at the compositor level.

The backend creates a D-Bus session with the GlobalShortcuts portal and registers all shortcuts in a single `bind_shortcuts` call. The compositor (KDE, etc.) shows a one-time confirmation dialog, then sends Activated/Deactivated signals whenever the user presses/releases the shortcut. These map directly to the existing `handle_shortcut_event` handler, so push-to-talk works out of the box.

Cancel is registered at init alongside the other shortcuts rather than dynamically, because each `bind_shortcuts` call can trigger a compositor dialog. The dynamic register/unregister calls from `mod.rs` are no-ops for this backend. If the portal isn't available (X11 session, unsupported compositor, etc.), it falls back to Tauri automatically and saves the setting.

Tested on KDE Plasma 6 Wayland -- shortcuts register, push-to-talk works with press/release, permissions persist across restarts when launched from the app menu.

GNOME 49's portal (v1) rejects `bind_shortcuts` for non-Flatpak apps. The backend detects this and falls back to Tauri, so there's no user-facing breakage. GNOME 50 doesn't fix this either -- seems to be a design limitation of their portal implementation.

The app needs to be launched from the desktop menu (not a terminal) for KDE to assign the correct `app_id` and persist shortcut permissions. When launched from a terminal, KDE inherits the terminal's app_id instead.

The recording overlay can steal focus on KDE Wayland since gtk-layer-shell isn't supported there. Setting overlay position to "None" works around it. This is pre-existing and now configurable via `HANDY_NO_GTK_LAYER_SHELL=1`.

### DIFF
--- a/src-tauri/Cargo.lock
+++ b/src-tauri/Cargo.lock
@@ -190,6 +190,20 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "7c02d123df017efcdfbd739ef81735b36c5ba83ec3c59c80a9d7ecc718f92e50"
 
 [[package]]
+name = "ashpd"
+version = "0.13.9"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "13bdf0fd848239dcd5e64eeeee35dbc00378ebcc6f3aa4ead0a305eec83d0cfb"
+dependencies = [
+ "enumflags2",
+ "futures-util",
+ "getrandom 0.4.1",
+ "serde",
+ "tokio",
+ "zbus",
+]
+
+[[package]]
 name = "async-broadcast"
 version = "0.7.2"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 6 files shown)


tim-seoss:
https://github.com/tauri-apps/global-hotkey/pull/172 extends Tauri to add Wayland (`org.freedesktop.portal.GlobalShortcuts`) support.  That might be a preferable approach?

PR #1302, opened 2026-04-15
feat(shortcut): replace push-to-talk toggle with two coexisting hotkeys (classic + PTT)

kud:
## Before Submitting This PR

## Human Written Description

I've been using Handy daily and kept running into the same friction: I'd be mid-conversation and want a quick dictation, but I was in toggle mode and had to go flip the push-to-talk setting first. SuperWhisper solves this elegantly — you have two separate hotkeys, and the behaviour is implicit in which one you press. No settings to toggle, no mental overhead.

This PR replaces the push-to-talk boolean toggle with two coexisting, independently configurable hotkeys: the existing classic toggle shortcut (press once to start, press again to stop and transcribe) and a new dedicated push-to-talk shortcut (hold to record, release to transcribe). Both are always active simultaneously. You just reach for whichever one fits the moment — hold PTT for a quick aside, use the classic toggle for a longer note.

I also cleaned up a few UX rough edges in the shortcut input components while I was there: a trash icon to explicitly clear a binding to empty, a "No hotkey" placeholder when a binding is unset, and modifier keys rendered as Unicode symbols (⌃ ⌥ ⇧ ⌘) so long combos stay readable.

## Related Issues/Discussions

Fixes #
Discussion:

## Community Feedback

No formal discussion thread yet — happy to start one if that would help. The feature pattern itself is well-established (SuperWhisper, which many Handy users are familiar with, works exactly this way).

## Testing

- 80 Rust unit tests pass including new regression tests covering the PTT binding path
- Tested manually on macOS: both hotkeys active simultaneously, PTT correctly triggers on key-down and stops on key-up, classic toggle behaviour unchanged
- Settings migration tested: existing installs with the old `push_to_talk` boolean pick up the new `transcribe_with_push_to_talk` binding correctly on first launch, and the new binding persists to disk via `store.save()`
- Empty-binding guard confirmed: leaving PTT unset does not trigger a shortcut registration error

image

## AI Assistance

**If AI was used:**

- Tools used: Claude Code (claude-sonnet-4-6) via the CLI
- How extensively: Claude Code implemented the majority of the Rust backend changes and the frontend shortcut input components under my direction. I designed the overall approach (two coexisting bindings, PTT driven by binding ID rather than a global flag), specified the UX details, and reviewed every diff. The "Human Written Description" above is my own words.


cjpais:
Right now no new features are being merged. I'm open to this, but it will be taken up at a later time as we revisit shortcuts generally. Right now there are an absolute ton of shortcut requests and I can't just merge one without people getting on me about not merging the one they want.

cjpais:
I appreciate this and I know the community wants it. Right now, I'm not accepting the new features, and I'm going to close this. I've discussed many times that shortcuts right now need a better general solution, because everyone wants some shortcuts, so we need a solution that allows people to generally customize a bunch of shortcuts, either in a very dynamic way, or a really good set of defaults for people to decide what they want their shortcuts for, or leave them unbinded, and this is a broader scope of work than just arbitrarily adding some shortcut.

PR #1305, opened 2026-04-17
feat: add phonetic replacements for cross-script vocabulary correction

SergeyLikholat:
Adds a new `phonetic_replacements` HashMap setting that maps phonetic transcription forms (e.g., Cyrillic "эн восемь эн") to their canonical spelling (e.g., "N8N"). This solves the problem where custom words fail for non-Latin scripts because the existing fuzzy matching (Levenshtein + Soundex) cannot match across different character sets.

The replacements run as a post-processing step before the existing custom word fuzzy matching, using case-insensitive regex with word boundaries. Longer phrases are matched first to avoid partial replacements.

Works with all transcription engines including Parakeet, Whisper, etc.

## Before Submitting This PR

- [ ] I have searched [existing issues](https://github.com/cjpais/Handy/issues) and [pull requests](https://github.com/cjpais/Handy/pulls) (including closed ones) to ensure this isn't a duplicate
- [ ] I have read [CONTRIBUTING.md](https://github.com/cjpais/Handy/blob/main/CONTRIBUTING.md)

## Related Issues/Discussions

Fixes #
Discussion:

## AI Assistance

**If AI was used:**

- Tools used:
- How extensively:

### DIFF
--- a/src-tauri/src/audio_toolkit/mod.rs
+++ b/src-tauri/src/audio_toolkit/mod.rs
@@ -8,6 +8,6 @@ pub use audio::{
     is_microphone_access_denied, is_no_input_device_error, list_input_devices, list_output_devices,
     read_wav_samples, save_wav_file, verify_wav_file, AudioRecorder, CpalDeviceInfo,
 };
-pub use text::{apply_custom_words, filter_transcription_output};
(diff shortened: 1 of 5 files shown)


cjpais:
Maybe this will come in in the future, but right now this is not going to be merged

PR #1311, opened 2026-04-19
fix: don't auto-delete recordings when changing history settings (#1262)

### LINKED ISSUE #1262: Data loss during the editing of History Limit / Auto-Delete Recordings
## Bug Description

By default, only 5 wav files are kept. This is already problematic IMHO, but not a bug. The bug is that when I set History Limit to 10000 and Auto-Delete Recordings to Never, it immediately deleted my precious wav files and transcripts, 4 at that point. I can't reproduce it, but the log below shows that it happened.

## System Information

**App Version:**

0.8.2

**Operating System:**

14.6

**CPU:**

Apple M1

**GPU:**

Apple M1

## Logs

[2026-04-09][18:41:08][handy_app_lib::managers::history][DEBUG] Saved history entry with id 15
[2026-04-09][18:41:08][handy_app_lib::managers::history][DEBUG] Deleted old WAV file: handy-1775756691.wav
[2026-04-09][18:41:08][handy_app_lib::managers::history][DEBUG] Cleaned up 1 old history entries by count
[2026-04-09][18:41:43][handy_app_lib::managers::history][DEBUG] Deleted old WAV file: handy-1775759238.wav
[2026-04-09][18:41:43][handy_app_lib::managers::history][DEBUG] Deleted old WAV file: handy-1775758698.wav
[2026-04-09][18:41:43][handy_app_lib::managers::history][DEBUG] Deleted old WAV file: handy-1775758171.wav
[2026-04-09][18:41:43][handy_app_lib::managers::history][DEBUG] Deleted old WAV file: handy-1775757041.wav
[2026-04-09][18:41:43][handy_app_lib::managers::history][DEBUG] Cleaned up 4 old history entries by count

ChristophNoetel:
## Summary

- Removes immediate `cleanup_old_entries()` calls from `update_history_limit()` and `update_recording_retention_period()`
- Changing history settings no longer triggers synchronous deletion of recordings
- Cleanup still happens naturally via `save_entry()` on the next recording

Fixes #1262

## Details

When a user changed "History Limit" or "Auto-Delete Recordings", the backend immediately ran cleanup with the new value. If the user was changing **both** settings (e.g., increasing the limit AND switching to "Never"), the first change triggered cleanup before the second was applied, causing unexpected data loss.

The fix removes the immediate cleanup trigger. Entries exceeding the new limit are cleaned up gradually as new recordings are made -- `cleanup_old_entries()` is already called inside `save_entry()` (`managers/history.rs:268`), so no entries accumulate indefinitely.

## Test plan

- [ ] Change History Limit from 5 to 10000 -- existing recordings should NOT be deleted
- [ ] Change Auto-Delete to "Never" -- existing recordings should NOT be deleted
- [ ] Change History Limit from 10000 to 3 -- existing recordings stay until next recording is made
- [ ] Make a new recording after lowering the limit -- old entries beyond the limit are cleaned up
- [ ] Verify normal recording/cleanup cycle still works

### DIFF
--- a/src-tauri/src/commands/history.rs
+++ b/src-tauri/src/commands/history.rs
@@ -108,19 +108,10 @@ pub async fn retry_history_entry_transcription(
 
 #[tauri::command]
 #[specta::specta]
-pub async fn update_history_limit(
-    app: AppHandle,
-    history_manager: State<'_, Arc<HistoryManager>>,
-    limit: usize,
-) -> Result<(), String> {
+pub async fn update_history_limit(app: AppHandle, limit: usize) -> Result<(), String> {
     let mut settings = crate::settings::get_settings(&app);
     settings.history_limit = limit;
     crate::settings::write_settings(&app, settings);
-
-    history_manager
-        .cleanup_old_entries()
-        .map_err(|e| e.to_string())?;
-
     Ok(())
 }
 
@@ -128,7 +119,6 @@ pub async fn update_history_limit(
 #[specta::specta]
 pub async fn update_recording_retention_period(
     app: AppHandle,
-    history_manager: State<'_, Arc<HistoryManager>>,
     period: String,
 ) -> Result<(), String> {
     use crate::settings::RecordingRetentionPeriod;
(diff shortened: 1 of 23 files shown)


cjpais:
Just letting you know in advance that it's going to be a while before I pull any changes in I'm doing some big work and I don't have time to pull things in currently. I think this is generally fine, but I do think there's still probably some kind of warning or toast that we can give the user a modal or something like this to tell them that hey, this setting is going to delete existing recordings or something like this, you know.

ChristophNoetel:
Added toast warnings (using Sonner) -- lowering the history limit or switching retention from "Never" to a time-based option now shows a heads-up that recordings will be cleaned up on the next recording. Fixing the translation consistency CI failure now.

PR #1319, opened 2026-04-21
feat(tray): add language switching submenu

eti0:
## Before Submitting This PR

## Human Written Description

one of the issues i face on a daily basis with handy is the fact that models often don't understand which language i'm speaking.

sometimes transcriptions get messed up and i have to open handy's settings window and switch the language.
it takes quite a bit of time and i think doing it directly in the menu bar could help me out (and others) a bit!

this pull request adds a language sub menu to the system tray so that you can switch transcription language without opening the full settings window.

here is the list of the features added with this pull request:
- shows the active language in the submenu label
- "auto detect" is always the first choice
- a "recent section" that shows up to five languages. this solves issues with models that support a large list of languages. i don't think users will want to switch between more than 5 languages at a time.

## Related Issues/Discussions

Discussion: https://github.com/cjpais/Handy/discussions/1312

## Testing

see video below for active testing.

https://github.com/user-attachments/assets/72b177d5-fd31-4300-9c6e-2a71a1d5a1ef

## AI Assistance

**If AI was used:**

- Tools used: OpenCode + Kimi K2.6
- How extensively: Fully


kud:
Yes, please. Add a language setting. How many times the AI is confused when I speak French. It works perfectly well in English, but when I speak French with a multilingual model like Parakeet, it does some kind of Frenglish.

cjpais:
For parakeet there is no option for this

alexisbchz:
feature much needed

shockme:
I would not make it a sub menu, it adds an extra click. I would put the last 2 or 3 most recently used languages plus "auto detect" directly in the menu. What do you guys think?

koloved:
@shockme For me, the best solution is using different languages through an LLM post-processing process. Especially with the function of a multiple hotkey function and various prompts. That means I can use my own language for translation into different languages. And at the same time, the translation text accuracy will be much higher than if I were to use native speech in the language we are speaking, which is partly a required but very niche feature.

shockme:
@koloved I don't use LLM post processing, so I wouldn't know, but if I set the target language to English, I can speak in my native language and Breeze ASR and I think Canary, too, translate to English – although it's not foolproof.

cjpais:
This feature more or less will be merged as is, but it needs significant cleanup as it's quite sloppy code

Issue #1283, opened 2026-04-12
Microphone initialization delay (~500ms) causes beginning of speech to be cut off


domdomegg:
## Summary

When using push-to-talk mode with `always_on_microphone: false`, there is a consistent ~500ms delay between pressing the shortcut and audio actually being captured, caused by on-demand microphone stream initialization. This means if you start speaking immediately after pressing the shortcut, the first word or syllable is lost.

## Environment

- macOS 26.3.1, Apple Silicon
- Handy v0.8.2
- Push-to-talk mode, `always_on_microphone: false`, `lazy_stream_close: false`

## Evidence from logs

Every recording shows the same pattern — the mic stream takes ~500ms to initialize before any audio is captured:

```
[DEBUG] TranscribeAction::start called for binding: transcribe_with_post_process
[DEBUG] On-demand mode: Starting recording first, then audio feedback
[INFO]  Microphone stream initialized in 499.61ms
[DEBUG] Recording started in 499.88ms
[DEBUG] TranscribeAction::start completed in 505.91ms
```

Across ~20 recordings in one session, initialization times ranged from **489ms to 591ms**, with the vast majority clustered around 500ms.

### Impact on transcription

The delay causes the beginning of speech to be lost. For example, saying "What's the time right now?" might transcribe as just "time right now?" — the opening words are missed entirely. Short utterances are especially affected, sometimes producing 0 audio samples:

```
[DEBUG] Recording stopped and samples retrieved in 165.75ms, sample count: 0
[DEBUG] Recording produced no audio samples; skipping persistence
```

In one session, 6 out of ~20 recordings captured 0 samples, likely because the key was released before the mic finished initializing.

domdomegg:
Seems like issue https://github.com/cjpais/Handy/issues/1165 might be related in that some discussion there was over the microphone initialization. I thought it would be sensible to create a new issue though because that one already seems to be focusing on the parakeet model regressions rather than the recording regressions. But I'm happy to combine if you would prefer that.

domdomegg:
I did some version comparison testing and found a significant regression in mic initialization time:

**v0.6.11**: **25-76ms**
**v0.7.12**: **380-416ms**
**v0.8.2**: **488-606ms**

That's a 10-20x regression from v0.6.11 to v0.8.2, and it got worse again between v0.7.12 and v0.8.2.

[...]

PR #1329, opened 2026-04-24
spike: mic-latency investigation + native-Swift perf harness (re #1283)

rsanheim:
Draft / spike — not proposed for merge as-is. Opening this to have something linkable from discussion on #1283 and to share the investigation code & notes.

## What's in here

1. **`test(audio): add AudioRecorder lifecycle coverage`** — five functional tests that drive the real default input device to pin current behaviour of `AudioRecorder::open` / `start` / `stop` / `close`, plus the level-callback path. Skip gracefully on hosts without a default input device so headless CI is fine. Ship-worthy independent of the rest.

2. **`chore: pin bun 1.3.13 via .mise.toml`** — the repo had no pinned bun version, so fresh mise-managed shells landed in `No version is set for shim: bun`. Low-impact repo hygiene. Happy to drop or adjust the version.

3. **`docs: add latency investigation writeup`** — `docs/latency-investigation.md`. Working document capturing what we measured, what the prior mental model got wrong, and an analysis of the v0.7.9 → v0.7.10 timing delta reported in #1283 (short version: it's an observability change from #945's async→sync `open()` restructure, not a runtime regression — `start_microphone_stream` is byte-identical between the two tags). Includes thread summary, v0.7.9/v0.7.10 code comparison, and implications.

4. **`tools: add macos-audio-perf native Swift harness`** — `tools/macos-audio-perf/`, a SwiftPM executable that measures keypress → first-sample → tone-played on macOS independent of Handy's Rust/cpal/rodio stack, used to separate platform-floor costs from implementation costs. Cold and warm modes, `--auto` for unattended runs with summary stats, writes wavs + CSV. Not built or shipped with the Handy app.

## Headline findings (full detail in the doc)

On a Mac Studio + Studio Display USB mic:

* **Handy cold path ≈ 640ms end-to-end.** Native-Swift AVAudioEngine on the same machine pays ~490ms in `engine.start()` alone plus ~180ms before first sample — essentially identical. cpal is not the bottleneck; it's CoreAudio + USB-mic hardware.
* **Warm path (engine kept running) ≈ 107ms in the Swift harness.** Handy's current `lazy_stream_close=true` warm path is estimated ~180–290ms because it still cycles the cpal stream plus waits 100ms before the tone plays plus cold-opens the rodio output stream on every press.
* **`tap_host_ms ≈ −10ms` in warm mode** — the running engine has audio buffered ahead of the keypress, so the "first sample" was captured before we pressed. This is the shape of the always-on pattern @m13v describes.
* **The bulk of the remaining ~105ms warm-path is `installTap()`, not hardware latency.** A tap-always-installed / flag-gated design should drop this to the low tens of ms.

## Not proposing

No change to Handy's runtime behaviour in this PR. If anything on the `docs/` or `tools/` side is useful to upstream separately, let me know.


cjpais:
I think this has been since resolved

PR #1340, opened 2026-04-28
Fix repeated filler bursts in transcription cleanup

Cloudymap1e:
## Summary

Fixes Parakeet-style filler bursts by collapsing runs of three or more consecutive filler-like tokens before normal transcription cleanup runs.

## Problem

Parakeet can sometimes turn a single pause into a long repeated filler sequence with mixed punctuation/case, for example:

```text
uh, uh. UH, uh
```

When explicit filler removal is disabled via an empty custom filler list, those bursts can pass through to paste/history unchanged.

## Changes

- Added a narrow repeated-filler collapse pass in `audio_toolkit::text`.
- Collapses only 3+ consecutive tokens from a conservative filler set.
- Preserves isolated filler words for the existing language/custom filler logic.
- Keeps custom empty filler-word behavior intact for ordinary, non-burst text.
- Added focused unit coverage for punctuation-heavy bursts, mixed filler runs, non-filler repetitions, and existing language safety cases.

## Validation

Automated:

- `cargo test filter_ --lib`
- `cargo test --lib`
- `cd src-tauri && cargo fmt -- --check`
- `./node_modules/.bin/eslint 'src/**/*.{ts,tsx}'`
- `git diff --check`

Manual:

- Built a local macOS app bundle from this branch.
- Installed it to `/Applications/Handy.app` after backing up the previous app.
- Verified in real dictation usage that repeated filler bursts are no longer pasted into the target app.

## Notes

- The local build used the Apple Intelligence stub path because the current Command Line Tools SDK exposes `FoundationModels.framework`, but its Swift macro plugin is unavailable in this environment.
- `bun` is not installed on the local test Mac, so frontend validation used the already-installed project `node_modules` binaries.

### DIFF
--- a/src-tauri/src/audio_toolkit/text.rs
+++ b/src-tauri/src/audio_toolkit/text.rs
@@ -231,6 +231,56 @@ fn get_filler_words_for_language(lang: &str) -> &'static [&'static str] {
 
 static MULTI_SPACE_PATTERN: Lazy<Regex> = Lazy::new(|| Regex::new(r"\s{2,}").unwrap());
 
+const REPEATED_FILLER_WORDS: &[&str] = &[
+    "uh", "um", "uhm", "umm", "uhh", "ah", "eh", "hm", "hmm", "mm", "mmm", "mh",
+];
+
+fn is_repeated_filler_token(token: &str) -> bool {
+    let normalized = token
+        .trim_matches(|c: char| !c.is_alphanumeric())
+        .to_lowercase();
+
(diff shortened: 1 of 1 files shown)


cjpais:
I appreciate the work on this, but I'm hoping to collapse this problem by moving to an inference library I am writing. I think it's quite possible to have much more robust detection and filtering of this in the future. I also suspect I will have better versions of parakeet available which hopefully won't have this issue as frequently. We will see

Basically it's coming, but some deeper work is going on! Give me some time

Issue #1038, opened 2026-03-14
[BUG] Direct paste method silently fails in Windows terminal applications


itsnotaboutthecell:
When `paste_method` is set to `direct`, transcribed text **silently fails to appear** in Windows terminal applications (Windows Terminal, PowerShell, cmd). Handy logs report `"Text pasted successfully"` with no error, even though no text is actually inserted into the terminal.

This is the Windows equivalent of #692 (macOS), where the developer acknowledged the `direct` method is broken in terminals due to upstream `enigo` limitations. However, on Windows this has an additional concern: **there is no error or warning** — the user has no indication the paste failed.

### Root Cause

The `direct` paste method uses enigo's `text()` function, which calls `SendInput` to simulate keystrokes. Terminal applications on Windows use the Console Input Buffer (`ReadConsoleInput`) rather than the standard window message pump, so the simulated keystrokes are either lost or not delivered to the terminal's input buffer.

### Steps to Reproduce

1. Set `paste_method` to `direct` in settings
2. Focus a terminal application (Windows Terminal, PowerShell, cmd)
3. Trigger transcription with `Ctrl+Space`
4. Speak clearly — transcription completes successfully
5. **Expected:** Text appears in terminal
6. **Actual:** Nothing appears. Log shows success:

```
[handy_app_lib::clipboard][INFO] Using paste method: Direct, delay: 60ms
[enigo::platform::win_impl][DEBUG] text(text: Hello, world.)
[handy_app_lib::actions][DEBUG] Text pasted successfully in 50.9968ms
```

### Additional Observations

- `Ctrl+V` paste method works in Windows Terminal but not in traditional cmd/PowerShell console
- `Shift+Insert` works universally across all terminal applications (**best workaround**)
- `direct` method works fine in GUI applications (browsers, text editors, etc.)
- The app does not detect the target application type or provide any fallback

### Suggestions

1. **Auto-detect terminal focus** and fall back to clipboard-based paste (`Shift+Insert` or `Ctrl+V`)
2. **Show a warning** in settings that `direct` mode is incompatible with terminal applications
3. Consider removing or deprecating `direct` on all platforms (as mentioned in #692 for macOS)

## System Information

**App Version:** v0.7.7

**Operating System:** Windows 11 Pro (Build 26200)

**CPU:** AMD Ryzen 9 5900 12-Core Processor

**Terminal tested:** Windows Terminal, PowerShell

cjpais:
[...]

Issue #692, opened 2026-02-01
[BUG] `Paste Method > Direct` repeats/breaks texts when voice-typing in terminal


deepakness:
## Bug Description

When **Paste Method** is set to **Direct**, typing in terminals when using tools like Codex or Claude Code, repeats texts multiple times and output text is broken (I will provide an example below). But typing in browser URL bar or in any other app works fine, even when the paste method is set to direct.

<img width="682" height="572" alt="Image" src="https://github.com/user-attachments/assets/121e1926-7fe7-4517-97f8-ea5a6b76b0f3" />

But everything works fine when the **Paste Method** is **Clipboard (Cmd+V)**, even when typing inside terminals.

### Example Outputs:

> Help me build an app that archives Twitt that archives Twitter post. When someone submits a tweets link, it should archive and save in a database and then give the permanent link to the user.ink, it should archive and save in a database and then give the permanent link to the user.ve and save in a database and then give the permanent link to the user.abase and then give the permanent link to the user.the permanent link to the user.o the user.

> Hey, I want you to go through the entire code base and then fix what the issue each.fix what the issue each.ach. Also, give me a depl.oyment checklist before I deploy the app.ore I deploy the app.

And this works as expected when paste method is set to clipboard.

## System Information

**App Version:**

v0.7.0

**Operating System:**

macOS Tahoe 26.2 (25C56)

**CPU:**

Apple M2

**GPU:**

Apple M2 GPU

cjpais:
This is a known broken issue and you should just use cmd+v on macOS. It's largely an upstream issue that won't be directly addressed. I'm planning to remove the direct option on macOS unless there's some reason it should be used

PR #1343, opened 2026-04-29
Add slow direct typing paste mode

0Downtime:
## Before Submitting This PR

## Human Written Description

This PR adds a slower direct typing paste option for environments where normal direct text input can overwhelm or duplicate keystrokes, especially remote desktop or VDI sessions. The new mode sends text as individual key clicks with short pauses, making Handy’s output more reliable in clients that behave better with physical-keyboard-style input.

## Related Issues/Discussions

Fixes: N/A
Discussion: N/A

Related context:

- Related to #692, which reported `Direct` paste repeating/breaking text in terminal-style inputs.
- Related to #1038, which reported `Direct` paste silently failing in Windows terminal applications.
- Related to #429 and #439, which cover other direct-input reliability/layout problems.
- Related to #502, where clipboard paste can insert stale clipboard content instead of the new transcription.
- Related to #689, an open Wayland/GNOME Remote Desktop portal approach. This PR is narrower and does not try to replace that work; it only adds an opt-in slow keyboard-style fallback for compatibility-sensitive targets.

## Community Feedback

This is intended as a bug-fix/workaround for paste reliability in remote desktop and VDI environments where existing paste modes can spam, duplicate, or otherwise mishandle generated text. I do not have a linked upstream discussion yet, but the use case is narrow: users who need Handy to paste into remote/virtualized desktops that do not reliably handle the existing direct input path.

## Problem

Handy currently supports clipboard-based paste modes and a `Direct` mode that uses Enigo's text input API. In some remote desktop, Horizon VDI, or compatibility-sensitive sessions, that high-level direct text path can be too fast or can be interpreted differently than normal physical keypresses. When that happens, transcription output may arrive incorrectly even though the transcription itself succeeded.

Clipboard paste is not always a good fallback in those environments because clipboard sync may be disabled, delayed, monitored, or blocked by policy. This leaves users without a reliable way to insert Handy output into the active remote application.

## Solution

This PR adds a new `Direct (Slow)` paste method backed by a `direct_compatibility` setting value. Instead of sending the whole string through the direct text API at once, the compatibility path maps common transcription characters to keyboard keys and sends individual key clicks with a short 15 ms pause between characters.

That gives users an opt-in paste mode that behaves more like physical keyboard input while leaving the existing defaults and faster paste modes unchanged.

## Changes

- Adds `PasteMethod::DirectCompatibility` / `direct_compatibility` as a persisted paste method value.
- Adds `input::paste_text_slow`, which types supported characters one at a time through Enigo key clicks.
- Handles common English transcription output: letters, digits, whitespace, newlines, tabs, and standard punctuation.
- Falls back for smart apostrophes and smart quotes by typing plain ASCII equivalents.
- Returns an explicit error for unsupported characters instead of silently dropping them.
- Routes the new mode through the existing paste pipeline, including auto-submit behavior.
- Adds the new option to the advanced paste-method dropdown as `Direct (Slow)`.
- Regenerates the TypeScript binding union so the frontend accepts `direct_compatibility`.
- Adds the new i18n key to all locale files so the settings UI does not introduce a hardcoded JSX string.
- Adds Rust unit tests for the slow typing character mapper.

## Behavior Notes / Limitations

- This mode is intentionally slower than `Direct`; it is meant for compatibility, not as the default fast path.
- The character mapper is currently focused on common ASCII transcription text and punctuation. Non-ASCII characters such as accented letters are rejected rather than guessed.
- Punctuation mapping follows a standard US-style keyboard layout. That matches the intended compatibility workaround but may not be correct for every physical keyboard layout.
- Existing paste methods are unchanged, and the default paste method behavior is unchanged.

## Testing

- [x] `cargo test slow_type --manifest-path src-tauri/Cargo.toml`
- [x] `bun run lint`
- [x] `cd src-tauri && cargo fmt -- --check`
- [x] `cargo clippy --manifest-path src-tauri/Cargo.toml` passes with existing repository warnings
- [ ] `bun run format:check` currently fails on existing `AGENTS.md` Prettier formatting before reaching Rust formatting
- [ ] Manual test in the target remote desktop / VDI environment

The targeted Rust test ran the new slow typing mapper coverage:

- `input::tests::slow_type_maps_common_transcription_text`
- `input::tests::slow_type_rejects_unsupported_characters`

N/A

## AI Assistance

**If AI was used:**

- Tools used: Codex
- How extensively: Codex wrote this change.


cjpais:
Probably fine to pull in I will test it when I have a chance

PR #1350, opened 2026-04-30
fix: restore Disable reasoning toggle for Custom provider

luoxi:
## Before Submitting This PR

- [x] I have searched existing issues and pull requests (including closed ones)

## Description

The 400 in #1221 only hits the Custom path. Custom sends the top-level OpenAI field `reasoning_effort: "none"`, but `"none"` isn't in the OpenAI spec (`minimal | low | medium | high` only) — it's an Ollama extension. So if a Custom base_url ends up at an OpenAI-strict endpoint (an OpenRouter proxy in front of OpenAI, Cerebras, Groq's OpenAI-compat mode, etc.), the request fails with 400.

The toggle goes in the Custom section, default ON:

- ON  (\`post_process_disable_reasoning = true\`)  → Handy sends \`reasoning_effort: "none"\` — unchanged from #1221, fast path for Ollama/Gemma.
- OFF (\`= false\`) → Handy sends no reasoning field, so strict endpoints stop returning 400.

Anyone hit by this needs to flip it off once after upgrading.

Added unit tests in \`llm_client.rs\` for request serialization (no reasoning fields, top-level field, nested object) so the JSON shape doesn't regress silently.

## Related Issues/Discussions

Follow-up to #1221.
Reported in: https://github.com/cjpais/Handy/pull/1221#issuecomment-4341366127
Approach agreed: https://github.com/cjpais/Handy/pull/1221#issuecomment-4348611119

## Screenshots/Videos

Screenshot 2026-04-05 at 10 35 38

## AI Assistance

- Tools used: Claude Code, plus Codex for a review pass.
- How extensively: I drafted the toggle by mostly restoring the middle revision of #1221. Codex spotted that I'd wired the toggle into both the Custom and OpenRouter sections by mistake — one shared bool, two sections, so flipping it from either would have changed both providers. Fixed that and added the serialization tests.


luoxi:
Heads up on the failing `code-quality` check: it's `prettier --check` complaining about `AGENTS.md`, which this PR doesn't touch. The same job has been failing on `main` since at least the v0.8.3 release commit (2026-04-27).

cjpais:
I do appreciate you putting this PR in. I just definitely don't want to add more UI, especially I think we can solve this without it. I'm going to be putting up a PR shortly, so I'm going to close this. And basically, what we're going to do is if there's an error, we will fall back so that more or less everything should work out of the box without having to customize anything

PR #1354, opened 2026-04-30
fix: prevent abort on quit by handling poisoned mutexes in Drop impls

marcel-dias:
## Summary

- Replace `.lock().unwrap()` with `match` + `into_inner()` in `TranscriptionManager::drop()` and `LoadingGuard::drop()` to prevent `abort()` on app quit

## Problem

When macOS sends a quit event (Cmd+Q, system shutdown, etc.), the app terminates via `[NSApplication terminate:]` → `exit()` → `__cxa_finalize_ranges`, which runs Rust Drop handlers. If a mutex was poisoned by a prior panic anywhere in the transcription pipeline, calling `.lock().unwrap()` inside a Drop impl triggers a second panic. Rust treats panic-during-Drop as unrecoverable and calls `abort()`, producing a `SIGABRT` crash instead of a clean exit.

The two affected locations:
- **`TranscriptionManager::drop()`** (`transcription.rs:846`) — `.lock().unwrap()` on `watcher_handle`
- **`LoadingGuard::drop()`** (`transcription.rs:59`) — `.lock().unwrap()` on `is_loading`

Note: `HandyKeysState::drop()` already uses `if let Ok(...)` which correctly handles poisoned locks.

## Fix

Use `PoisonError::into_inner()` to recover the underlying `MutexGuard` even when the mutex is poisoned. This is the standard Rust pattern for "I need this data during cleanup regardless of prior panics."

## Test plan

- [ ] `cargo check` / `cargo clippy` pass
- [ ] App starts and records normally
- [ ] Cmd+Q exits cleanly without crash
- [ ] Force-quit and re-launch works correctly

### DIFF
--- a/src-tauri/src/managers/transcription.rs
+++ b/src-tauri/src/managers/transcription.rs
@@ -56,7 +56,12 @@ pub struct LoadingGuard {
 
 impl Drop for LoadingGuard {
     fn drop(&mut self) {
-        let mut is_loading = self.is_loading.lock().unwrap();
+        // Recover from a poisoned mutex instead of panicking —
+        // a panic inside Drop calls abort().
+        let mut is_loading = match self.is_loading.lock() {
+            Ok(g) => g,
+            Err(e) => e.into_inner(),
+        };
         *is_loading = false;
         self.loading_condvar.notify_all();
     }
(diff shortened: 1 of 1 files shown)


cjpais:
Thanks we will pick this up after review

marcel-dias:
I believe I was able to map the steps that cause the issue. 

I'm using `Whisper Turbo` model and the `unload model = after 1 hour`. With these settings, around the 5th recording Handy got stuck, like it opens the overlay, listen but do not record or transcribe the audio.  After closing Handy and opening again MacOS shows this error. 

I changed the `unload model` to `after 5 min` and it stopped failing. 
using version 0.8.3

cjpais:
Interesting @marcel-dias thanks for the info

PR #1362, opened 2026-05-02
fix: enable longform chunking for GigaAM via VadChunked

android6:
## Summary

Wires up the existing `VadChunked` utility from `transcribe-rs` for GigaAM-v3 to fix gibberish output and silent failures on recordings >60 seconds. No changes to `transcribe-rs` — uses its public API. UI unchanged. Tested on a 4:25 Russian recording (previously a deterministic silent fail) — now produces a coherent transcript.

## How it works

**Before:**

```
Handy → gigaam_engine.transcribe(audio)
                       ↓
                       single ONNX inference over the full audio buffer
                       ↓
                       gibberish (>60s) or silent fail (>4min)
```

**After (this PR):**

```
Handy → chunker.transcribe(gigaam_engine, audio)
                ↓
                split audio at silence boundaries (Silero VAD)
                ↓
                for each chunk (≤30s):
                    gigaam_engine.transcribe(chunk)
                ↓
                join results with " "
                ↓
                coherent transcript
```

GigaAM-v3 was trained on segments ≤25-30 s. The new logic lives entirely between Handy and the model — using the existing `VadChunked` from `transcribe-rs`. `VadChunked` has been available in `transcribe-rs` since 0.3.4; the current default `0.3.8` already includes it. Whisper has its own chunking in C++, so only GigaAM is affected.

## Changes

- `src-tauri/Cargo.toml`: enable the `vad-silero` feature on `transcribe-rs` (the Silero VAD model is already shipped with Handy).
- `src-tauri/src/managers/transcription.rs`:
  - `LoadedEngine::GigaAM` now holds `(GigaAMModel, VadChunked)`.
  - On GigaAM load, builds the pipeline `SileroVad → SmoothedVad → VadChunked` with parameters tuned for the longform CTC use-case.
  - On transcribe, calls `chunker.transcribe(model, &audio)` instead of `model.transcribe(...)` directly.
- Whisper, Parakeet, Moonshine, SenseVoice, Canary, Cohere — all unaffected.

### DIFF
--- a/src-tauri/Cargo.toml
+++ b/src-tauri/Cargo.toml
@@ -69,7 +69,7 @@ rusqlite = { version = "0.37", features = ["bundled"] }
 tar = "0.4.44"
 flate2 = "1.0"
 sha2 = "0.10"
-transcribe-rs = { version = "0.3.8", features = ["whisper-cpp", "onnx"] }
+transcribe-rs = { version = "0.3.8", features = ["whisper-cpp", "onnx", "vad-silero"] }
 handy-keys = "0.2.4"
 ferrous-opencc = "0.2.3"
 clap = { version = "4", features = ["derive"] }
@@ -88,7 +88,7 @@ tauri-plugin-single-instance = "2.3.2"
 tauri-plugin-updater = "2.10.0"
 
 [target.'cfg(windows)'.dependencies]
-transcribe-rs = { version = "0.3.3", features = ["whisper-vulkan", "ort-directml"] }
+transcribe-rs = { version = "0.3.3", features = ["whisper-vulkan", "ort-directml", "vad-silero"] }
 windows = { version = "0.61.3", features = [
   "Win32_Media_Audio_Endpoints",
   "Win32_System_Com_StructuredStorage",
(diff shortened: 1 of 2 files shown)


cjpais:
#1173

android6:
**Quick follow-up: tried mid-pass aggregation mirroring upstream `gigaam/vad_utils.py`, null result.**

```
Pass 1: VadChunked → primary chunks      (as in this PR)
Pass 2: collected → merge neighbors → final chunks    ← NEW
        rule: merged > 22s OR curr > 15s → close
Pass 3: gigaam.transcribe(each)
```

On voice-typing recordings: no measurable quality change vs. the version in this PR. Sharing in case it's useful for #1173.

android6:
> **Quick follow-up: tried mid-pass aggregation mirroring upstream `gigaam/vad_utils.py`, null result.**
> 
> ```
> Pass 1: VadChunked → primary chunks      (as in this PR)
> Pass 2: collected → merge neighbors → final chunks    ← NEW
>         rule: merged > 22s OR curr > 15s → close
> Pass 3: gigaam.transcribe(each)
> ```
> 
> On voice-typing recordings: no measurable quality change vs. the version in this PR. Sharing in case it's useful for #1173.

**Update:** the "null result" above was on voice-typing samples that happened to be steady-paced. On samples with frequent mid-sentence pauses (where the speaker stops mid-thought to pick the right word) — aggregation does help: those pauses no longer split a single sentence into two separate ones (otherwise the next fragment starts with a capital letter and reads as a new thought).

PR #1368, opened 2026-05-06
fix(post-processing): support DeepSeek custom endpoints

### LINKED ISSUE #1342: Custom post-processing fails with DeepSeek because reasoning_effort:none is rejected
## Problem

When using the Custom post-processing provider with DeepSeek's OpenAI-compatible endpoint:

- Base URL: `https://api.deepseek.com`
- Model: `deepseek-v4-flash`

post-processing fails with HTTP 400 because Handy sends:

```json
"reasoning_effort": "none"
```

DeepSeek rejects this field/value:

```text
reasoning_effort: unknown variant `none`, expected one of `high`, `low`, `medium`, `max`, `xhigh`
```

## Context

PR #1221 added `reasoning_effort: "none"` for Custom providers to avoid thinking-mode latency in local models such as Ollama/Gemma/Qwen. That behavior is useful and should remain for those endpoints.

The issue is that not all OpenAI-compatible APIs accept the same reasoning-control parameters. DeepSeek uses a different `thinking` object for V4 thinking mode, and rejects `reasoning_effort: "none"`.

DeepSeek docs: https://api-docs.deepseek.com/guides/thinking_mode

## Expected behavior

Custom provider should still work with DeepSeek-compatible endpoints.

Ideally:

- Preserve `reasoning_effort: "none"` for local Custom endpoints where it works.
- Avoid sending `reasoning_effort: "none"` to DeepSeek.
- Optionally send `thinking: { "type": "disabled" }` for DeepSeek to avoid extra latency during transcript cleanup.

## Actual behavior

Post-processing fails and Handy falls back to the original transcription.

## Logs

```text
LLM post-processing failed for provider 'custom': API request failed with status 400 Bad Request:
{"error":{"message":"Failed to deserialize the JSON body into the target type: reasoning_effort: unknown variant `none` ..."}}
```

## Notes

This is related to #1221, but not a request to remove the Custom reasoning optimization globally. The local-model latency improvement should be preserved.

sinco-z:
## Summary

Fixes Custom post-processing against DeepSeek-compatible endpoints.

PR #1221 started sending `reasoning_effort: "none"` for Custom providers to avoid thinking-mode latency in local OpenAI-compatible servers such as Ollama. DeepSeek rejects that value with HTTP 400, so Custom post-processing falls back to the original transcription.

This keeps the existing Custom behavior for non-DeepSeek endpoints, but skips `reasoning_effort: "none"` when the Custom provider points at DeepSeek.

## Changes

- Detect Custom providers whose base URL points to `api.deepseek.com`.
- Preserve `reasoning_effort: "none"` for other Custom providers.
- Omit `reasoning_effort` for DeepSeek Custom endpoints.

## Validation

- Manually tested with:
  - Provider: Custom
  - Base URL: `https://api.deepseek.com`
  - Model: `deepseek-v4-flash`
- Verified post-processing no longer returns the `reasoning_effort` HTTP 400.
- `cargo fmt -- --check`

I also attempted `cargo check --no-default-features`, but current `main` fails in my local macOS Command Line Tools environment while compiling the existing Apple Intelligence Swift bridge:

```text
FoundationModelsMacros.GenerableMacro could not be found
```

That failure happens in the existing build script/Swift bridge path and is unrelated to this patch.

Fixes #1342

### DIFF
--- a/src-tauri/src/actions.rs
+++ b/src-tauri/src/actions.rs
@@ -125,12 +125,16 @@ async fn post_process_transcription(settings: &AppSettings, transcription: &str)
         .cloned()
         .unwrap_or_default();
 
+    let is_deepseek_custom =
+        provider.id == "custom" && provider.base_url.contains("api.deepseek.com");
+
     // Disable reasoning for providers where post-processing rarely benefits from it.
-    // - custom: top-level reasoning_effort (works for local OpenAI-compat servers)
-    // - openrouter: nested reasoning object; exclude:true also keeps reasoning text
-    //   out of the response so it can't pollute structured-output JSON parsing
+    // Most local OpenAI-compatible servers accept `reasoning_effort: "none"`,
+    // but DeepSeek rejects that enum.
+    // OpenRouter uses a nested reasoning object; exclude:true also keeps reasoning
+    // text out of the response so it can't pollute structured-output JSON parsing.
     let (reasoning_effort, reasoning) = match provider.id.as_str() {
-        "custom" => (Some("none".to_string()), None),
+        "custom" if !is_deepseek_custom => (Some("none".to_string()), None),
         "openrouter" => (
             None,
             Some(crate::llm_client::ReasoningConfig {


sinco-z:
One follow-up question about scope:

This PR intentionally keeps the fix minimal: for Custom providers pointing at `api.deepseek.com`, it only skips `reasoning_effort: "none"` so DeepSeek no longer returns HTTP 400. This restores the pre-#1221 behavior for DeepSeek.

DeepSeek V4 docs say thinking mode defaults to enabled, and non-thinking mode requires:

```json
"thinking": { "type": "disabled" }
```

For transcript cleanup, disabling thinking is probably the better UX because it avoids extra latency. I tested that locally and it reduced post-processing latency noticeably.

I kept it out of this PR because adding a provider-specific `thinking` request field is a broader behavior change than the minimal regression fix.

Would you prefer this PR to stay minimal, or should it also send `thinking: { "type": "disabled" }` for DeepSeek Custom endpoints?

sinco-z:
I noticed #1350 restores the Disable reasoning toggle for Custom providers, following the option-B direction from #1221.

If #1350 is the preferred fix for strict OpenAI-compatible endpoints, this PR may be redundant and I’m happy to close it.

The only difference is that this PR handles `api.deepseek.com` automatically without requiring users to discover and flip the toggle. Let me know whether you’d prefer to keep this DeepSeek-specific zero-config fix, or close this in favor of #1350.

cjpais:
I think picking one or the other is the best or there is more state to deal with

I'm not really decided as both options are not great imo.

This, while it works feels a bit of a hack/brittle for other models which maybe have the same issue

The other one, is non obvious and if you change models frequently (not really sure why but have to leave it open), you may have to toggle it on and off which is quite annoying

sinco-z:
Thanks, that makes sense. This is starting to go beyond the narrow DeepSeek regression fix, so I opened a separate design discussion for the longer-term provider/reasoning-control UX: https://github.com/cjpais/Handy/discussions/1378

That should keep this PR focused while giving us a better place to discuss whether Custom should stay conservative, whether common providers like DeepSeek/Ollama should become presets, and how reasoning/thinking controls should be handled without making normal users configure provider-specific request fields.

cjpais:
I think that this is kind of a hack as written, But I do appreciate you submitting it regardless and I will be submitting a fix myself for this that should fall back automatically and I think we will probably also add official support for deep seek so it doesn't need to run through the custom API provider so it should more or less hopefully just work out of the box

PR #1376, opened 2026-05-07
fix(parakeet): chunk long audio (>25s) to avoid silent failure

florian-hoenicke:
## Problem

Parakeet TDT 0.6B has a fixed positional-encoding budget (~30s @ 16kHz). Passing a buffer longer than that to `transcribe_with` silently produces an unusable result, so any recording longer than ~30 seconds returns garbage or empty text. Reproduced with a 16-minute recording on macOS using `parakeet-tdt-0.6b-v3-int8`. Likely the same root cause as the failures reported in #1165.

## Fix

In the `LoadedEngine::Parakeet` arm of `TranscriptionManager::transcribe`, when `audio.len() > 25s`, slide a 25s window with 1s overlap, call `parakeet_engine.transcribe_with` per chunk, and join the resulting `text`. Short audio (≤ 25s) takes the original single-call path with no behavior change.

Segments are dropped on the long-audio path because `TranscriptionManager` only consumes `result.text` downstream — preserving them would require shifting timestamps per chunk for no current consumer.

## Test plan

- [x] Verified locally on macOS 15.7.5 / M4 Pro with a 16:38 recording in `parakeet-tdt-0.6b-v3-int8` — produces coherent text instead of silent failure
- [x] Short recordings still go through the single-call path (no regression)
- [ ] CI

### DIFF
--- a/src-tauri/src/managers/transcription.rs
+++ b/src-tauri/src/managers/transcription.rs
@@ -556,15 +556,70 @@ impl TranscriptionManager {
                                 .map_err(|e| anyhow::anyhow!("Whisper transcription failed: {}", e))
                         }
                         LoadedEngine::Parakeet(parakeet_engine) => {
+                            // Parakeet TDT 0.6B has a fixed positional-encoding budget
+                            // (~30s @ 16kHz). For longer recordings we slide a 25s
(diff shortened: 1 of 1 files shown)


cjpais:
See existing PRs before opening new ones please.

PR #1396, opened 2026-05-13
fix(parakeet): chunk long audio to avoid encoder positional encoding …

anwyho:
…overflow

Parakeet's int8 ONNX encoder caps positional encoding around 4875 subsampled frames (~6.5 min of 16kHz audio at subsampling factor 8). Feeding audio longer than that into a single transcribe_with() call fails inside the first self-attention layer:

  Non-zero status code returned while running Add node.
  Name:'/layers.0/self_attn/Add_2' Status Message: ...
  Attempting to broadcast an axis by a dimension other than 1.
  4875 by 9875

Fix: split audio longer than 120s into overlapping 120s windows (15s overlap) before calling the engine, then stitch segments using their timestamps. Audio at or under 120s still flows through a single call with zero overhead.

Dedup: any segment whose midpoint falls inside the first half of the overlap region of a non-first chunk is dropped (the previous chunk already covered it). When the engine returns no segments we fall back to raw text concatenation, which can duplicate words at boundaries but keeps output usable.

Repro before this fix: record/transcribe any audio >~6.5 min with parakeet-tdt-0.6b-v3. After this fix a 13-min recording transcribes end to end.

## Before Submitting This PR

## Related Issues/Discussions

Fixes #
Discussion:

## AI Assistance

**If AI was used:**

- Tools used:
- How extensively:

### DIFF
--- a/src-tauri/src/managers/transcription.rs
+++ b/src-tauri/src/managers/transcription.rs
@@ -25,8 +25,116 @@ use transcribe_rs::{
         Quantization,
     },
     whisper_cpp::{WhisperEngine, WhisperInferenceParams},
-    SpeechModel, TranscribeOptions,
+    SpeechModel, TranscribeError, TranscribeOptions, TranscriptionResult, TranscriptionSegment,
 };
+
+// Parakeet's ONNX encoder caps positional encoding around 4875 subsampled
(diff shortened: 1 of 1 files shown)


anwyho:
whoops, sorry! I thought this was on my own fork.

Issue #1375, opened 2026-05-07
[BUG] Wayland: recording overlay displayed on secondary monitor


tredondo:
## Before You Submit

Similar to #811, but for Linux, and the overlay IS visible, just on the secondary monitor.

## Bug Description

My Kubuntu 26.04 laptop is connected to an external monitor, which is the primary. When I record, the overlay appears on the laptop monitor, out of visual focus.

## System Information

**App Version:** 0.8.3

**CPU:** 20 × 12th Gen Intel® Core™ i7-12700H

**GPU:** Mesa Intel® Iris® Xe Graphics

Issue #1282, opened 2026-04-12
[BUG] Post-processing never triggers on Linux KDE Wayland (v0.8.2)


essoen:
## Bug Description

Post-processing is enabled and fully configured (Anthropic provider, valid API key, model selected, prompt set) but never fires. No API requests are made — confirmed by checking the Anthropic API dashboard which shows $0.00 usage. Debug logs show no post-processing step between transcription and paste — it goes straight from `Transcription completed` to `Using paste method`.

Same thing happens with OpenAI Keys. I verified that it reaches the servers and that the API key works for both providers as I get to choose models.

This happens with:
1. The regular transcription hotkey (with Post Processing toggled on in Advanced → Experimental)
2. The dedicated Post-Processing Hotkey (Ctrl+Shift+Space)
3. When configuring a global hotkey in KDE.

## System Information

**App Version:** v0.8.2

**Operating System:** Fedora 43, KDE Plasma (Wayland)

**CPU:** AMD Ryzen 5 3600

**GPU:** AMD Radeon RX 5700 XT (RADV NAVI10) 8.0 GB 

## Logs

Using regular hotkey with Post Processing enabled:

```
[2026-04-12][19:48:52][handy_app_lib::managers::transcription][INFO] Transcription completed in 2035ms
[2026-04-12][19:48:52][handy_app_lib::managers::transcription][INFO] Transcription result: <transcription>
[2026-04-12][19:48:52][handy_app_lib::clipboard][INFO] Using paste method: CtrlV, delay: 60ms
[2026-04-12][19:48:52][handy_app_lib::clipboard][INFO] Using wl-copy for clipboard write on Wayland
[2026-04-12][19:48:52][handy_app_lib::clipboard][INFO] Using ydotool for key combo
```

Using dedicated Post-Processing Hotkey (Ctrl+Shift+Space):

```
whisper_full_with_state: auto-detected language: no (p = 0.984760)
[2026-04-12][19:50:23][handy_app_lib::managers::transcription][INFO] Transcription completed in 2205ms
[2026-04-12][19:50:23][handy_app_lib::managers::transcription][INFO] Transcription result: <transcription>
[2026-04-12][19:50:23][handy_app_lib::clipboard][INFO] Using paste method: CtrlV, delay: 60ms
[2026-04-12][19:50:23][handy_app_lib::clipboard][INFO] Using wl-copy for clipboard write on Wayland
[...]

PR #1397, opened 2026-05-13
feat(linux): optional GNOME Shell extension for live status display

R11manish:
## Before Submitting This PR

> Note on the feature freeze: no prior discussion explicitly requests this feature. I'm opening it anyway because (a) the implementation is fully optional and isolated (Linux-only, gated behind `#[cfg(target_os = "linux")]`, zero impact on macOS/Windows builds) and (b) several existing Linux/Wayland issues point at the same underlying pain — users can't tell at a glance whether Handy is recording / transcribing / processing. Happy to convert this into a Discussion first if maintainers prefer that path.

## Human Written Description

Adds an optional GNOME Shell extension that surfaces Handy's live activity state (recording / transcribing / processing) in the GNOME top bar, communicating with the running app over a small D-Bus interface (`com.pais.Handy.Status`). Linux/Wayland users currently get limited visual feedback while Handy is working in the background — this gives a glanceable, on-brand indicator that disappears when idle.

## Related Issues/Discussions

No prior discussion exists for this exact feature, but it lives in the same problem space as:

- [#1216 — External Script for Gnome on Wayland](https://github.com/cjpais/Handy/discussions/1216) (users writing their own workarounds for Wayland gaps)
- [#1184 — Tray menu tooltip is empty](https://github.com/cjpais/Handy/issues/1184) (users want clearer state from the tray)
- [#1243 — Mic select in menu bar](https://github.com/cjpais/Handy/discussions/1243) and [#1312 — Quick language switching from the menu bar](https://github.com/cjpais/Handy/discussions/1312) (general appetite for OS-native status surfaces)
- Plus the broader cluster of open Wayland-tray / overlay issues (#1375, #1282, etc.)

Fixes #
Discussion:

## Community Feedback

None gathered yet — see the note in the freeze section above. Willing to open a Discussion retroactively if maintainers want to gauge interest before merging.

## What's in the change

**Rust side (`src-tauri/`)**
- New `status.rs` module exposing `StatusManager` and `ActivityStatus { Idle, Recording, Transcribing, Processing }`
- D-Bus service (`com.pais.Handy.Status` at `/com/pais/Handy`) with `GetStatus` method and `StatusChanged` signal — **entirely gated behind `#[cfg(target_os = "linux")]`**
- `zbus = "5"` added under `[target.'cfg(target_os = "linux")'.dependencies]` so Windows/macOS builds don't pull or link it
- `set_status_safe(app, status)` helper is callable cross-platform; on non-Linux it's a no-op for the D-Bus path

**Extension (`gnome-extension/`)**
- `extension.js` — `PanelMenu.Button` that watches the D-Bus name, hides itself while idle, swaps icons per state
- Three combined SVGs (`handy-{recording,transcribing,processing}-v2-symbolic.svg`) — each one packs the glyph + bold letter (R/T/P) into a single 16×16 icon, color-coded red/amber/green via CSS
- `install.sh` — one-command pack + install + enable via the official `gnome-extensions` CLI
- `metadata.json` declares compatibility with Shell `45, 46, 47, 48`
- Subdirectory `README.md` with install + D-Bus inspection commands
- Root `README.md` gets a short Integrations subsection pointing at the install script

## Testing

Tested locally on GNOME Shell 48 / Wayland:
- Built `.deb` (`bun run tauri build --bundles deb`), installed, launched Handy
- `gdbus call --session --dest com.pais.Handy --object-path /com/pais/Handy --method com.pais.Handy.Status.GetStatus` returns the current state
- Indicator appears/disappears correctly across all three states
- Verified compile-time gating: zbus is not pulled on non-Linux targets (visible in `Cargo.toml`)

Not yet tested:
- GNOME Shell 45 / 46 / 47 (metadata declares support; APIs used are stable across that range, but a maintainer with access to those versions should sanity-check)
- macOS / Windows builds (no behavioral change expected — Rust code is `cfg`-gated and the extension files are never bundled)

## Screenshots/Videos

(I'll attach screenshots showing recording / transcribing / processing states via the GitHub UI before merging.)

## AI Assistance

- [x] AI was used

**If AI was used:**

- Tools used: Claude Code
- How extensively: AI-assisted heavily for the GNOME extension JavaScript, the combined SVG icons, the install script, README polishing, and the cross-platform gating review of the Rust side. The original D-Bus `status.rs` skeleton predated the AI-assisted session and was refined collaboratively. All design choices (icon glyphs, R/T/P labels, color scheme, install flow) were chosen and reviewed by me.


The-Robin-Hood:
Great work on this, @R11manish  but one thought worth raising:

The D-Bus layer here is genuinely useful, but it's inherently Linux/GNOME-bound. Have you considered a small `127.0.0.1` HTTP server (`GET /status` → plain text) alongside or instead?
This is the pattern Ollama, LM Studio, and most local-first tools use i believe nd it means Waybar, eww, KDE widgets, Windows tray tools, and shell scripts can all integrate without any platform-specific knowledge. GNOME extension could sit on top of it just the same. 

I made some work on that, have a look [branch](https://github.com/cjpais/Handy/compare/main...The-Robin-Hood:Handy:feat/live_status)

R11manish:
> Great work on this, @R11manish but one thought worth raising:
> 
> The D-Bus layer here is genuinely useful, but it's inherently Linux/GNOME-bound. Have you considered a small `127.0.0.1` HTTP server (`GET /status` → plain text) alongside or instead? This is the pattern Ollama, LM Studio, and most local-first tools use i believe nd it means Waybar, eww, KDE widgets, Windows tray tools, and shell scripts can all integrate without any platform-specific knowledge. GNOME extension could sit on top of it just the same.
> 
> I made some work on that, have a look [branch](https://github.com/cjpais/Handy/compare/main...The-Robin-Hood:Handy:feat/live_status)

Thanks , nopes we actually don't need i believe tray icon thing on other platform windows/ macOS . why  because overrlay works absoultely fine over there , in linux especailly in wayland (mutter) there exactly we have the problem for overlay which handy shows , lemme know your thoughs on this

The-Robin-Hood:
Hey @R11manish, totally fair point the overlay works great on macOS/Windows. But even scoping this to Linux only, a local HTTP endpoint would still be more flexible than D-Bus alone. Not every Linux user runs GNOME, a lot of us are on Hyprland, Sway, KDE, or using Waybar/eww/AGS for status bars, and none of those can easily consume a D-Bus signal without writing custom glue code. It's a small addition that massively widens who can actually use this on Linux.

wpbrown:
The glue code you need is very small and typical for bars. I have tiny python scripts watching D-BUS already feeding stdio lines out to my Polybar custom elements. D-BUS makes sense here because you want to have these status changes without polling and little to no overhead. For HTTP you would need to poll or use SSE, it could work, but seems like overkill. Handy already depends on D-BUS by way of Tauri single-instance plugin.

I am looking forward to using this for other custom integrations. I would like really to see start/stop methods on the D-BUS interface, not just status. "toggle" in the CLI is less ideal that start stop when you're trying to synchronize the state with other systems.

Ideally the CLI would have a handy watch command that provides updates on stdio, the most universal integration mechanism. Handy would handle IPC in a platform agnostic way so the user doesn't have to care about D-BUS, HTTP, named pipes, whatever. I don't think that would preclude this PR, as on Linux this would be an ideal transport for that feature.

R11manish:
@cjpais  any updates

koloved:
@R11manish Could you please test the flatpack build? https://github.com/cjpais/Handy/pull/689#issuecomment-4820254699

The discussion here says that this is a general problem with Linux Wayland, although the overlay works for my KDE Wayland.
The current problem is not a problem with Linux as a whole, but specifically with Gnome.

R11manish:
> @R11manish Could you please test the flatpack build? [#689 (comment)](https://github.com/cjpais/Handy/pull/689#issuecomment-4820254699)
> 
> The discussion here says that this is a general problem with Linux Wayland, although the overlay works for my KDE Wayland. The current problem is not a problem with Linux as a whole, but specifically with Gnome.

yes exactly this is promblem with gnome wayland

koloved:
@R11manish there is actually update to 0.9 - https://github.com/cjpais/Handy/pull/1560#issuecomment-4886091447

settings that you should use with it:
paste method: ctrl+v clipboard
shortcut backend: xdg-desktop-portal

PR #1400, opened 2026-05-13
feat: highlight target window during transcription

onepunk:
## Summary
- Add a Windows-only target window highlight that frames the active window while Handy records and transcribes.
- Flash and hide the highlight after successful paste, with lifecycle guards for cancellation and quick successive recordings.
- Add an Advanced Settings toggle, persisted `highlight_target_window` setting, generated bindings, and locale keys.

## Test Plan
- [x] `bun run build`
- [x] `bun run lint`
- [x] `bun run check:translations`
- [x] `cargo fmt -- --check`
- [ ] `cargo test delayed_hide --lib` could not run locally because the Linux environment is missing `xi.pc` required by the existing `x11` crate build.

## Notes
- Non-Windows platforms use no-op stubs for the highlight module.
- The highlight webview is added to Tauri capabilities so it can receive frontend events.


cjpais:
I'm sorry, I don't totally understand what this does. Can you explain it or add a screenshot or a video? Also, it seems like you didn't collect any community feedback and we're not accepting new features right now. So, unless this is a bug fix, it's unlikely to be merged at the moment.

onepunk:
Sorry, I should have included a screenshot. This update highlights the window that a user is focused on while Handy is running to make them aware where the text will be pasted. It's toggleable in the Experimental settings.

Handy Border Highlight

koloved:
It would be useful to have this, because sometimes you record, stop recording and realize that the insertion has not occurred, you have to go into the application and copy the text.

Does it work with any application? And if it's full-screen, how does it work?

PR #1050, opened 2026-03-15
openAI compatible API

asharpaev:
## Before Submitting This PR

**Please confirm you have done the following:**

- [x] I have searched [existing issues](https://github.com/cjpais/Handy/issues) and [pull requests](https://github.com/cjpais/Handy/pulls) (including closed ones) to ensure this isn't a duplicate
- [x] I have read [CONTRIBUTING.md](https://github.com/cjpais/Handy/blob/main/CONTRIBUTING.md)

**If this is a feature or change that was previously closed/rejected:**

- [ ] I have explained in the description below why this should be reconsidered
- [ ] I have gathered community feedback (link to discussion below)

## Human Written Description

OpenAI API support for Handy

## Related Issues/Discussions

Fixes #
Discussion:

## AI Assistance

- [ ] No AI was used in this PR
- [ ] AI was used (please describe below)

**If AI was used:**

- Tools used:
- How extensively:

PR #804, opened 2026-02-12
feat: Remote Whisper (OpenAI-compatible) transcription

vakovalskii:
Hi! I use Handy a lot and I also run my own remote transcription servers (Whisper, OpenAI-compatible). I wanted to use them with Handy, but didn’t find an existing way to switch STT to a remote /v1/audio/transcriptions endpoint, so I implemented it.

What this adds:
- Optional "Remote Whisper API" mode (Advanced settings).
- Sends the recorded audio to a configurable OpenAI-compatible /audio/transcriptions endpoint instead of local model inference.
- Settings: base URL, API key (Bearer), model, prompt, language, temperature.
- No real server URLs or tokens are included in the repo (examples use whisper.example.com).

Notes:
- Some parts were initially drafted with Codex CLI, but I reviewed, tested, and adjusted everything myself before submitting.

Testing:
- bun run build
- Manual smoke test: enable Remote Whisper, set base URL/api key, record via hotkey, verify transcription result.

PR #1435, opened 2026-05-26
Add custom local model endpoint

arvinmi:
## Before Submitting This PR

## Human Written Description

I built this because I wanted to use Handy with a local model endpoint, specifically Google MedASR, for medical dictation where the built-in model options were not a good fit for my use case. MedASR is a Google ASR model based on the Conformer architecture, but it is not currently something Handy can run through the Whisper GGML model path or the existing `transcribe-rs` ONNX model engines.

I also think medical dictation is a strong fit for Handy’s local-first approach. From my own experience, there seems to be real interest in local transcription workflows for clinical note-taking, especially on Windows, where there are fewer polished voice transcription options in this space.

I understand Handy is intentionally local-first, and I also understand that similar transcription endpoint ideas have been discussed and rejected before many times. The angle I am coming from is not hosted API provider support but a small route for local or self-hosted ASR runtimes that do not fit Handy's current model formats, especially niche or domain-specific models like medical use cases.

I am not sure how much Handy wants to keep expanding direct support for more model architectures over time. This tries to solve that in a minimal way by letting advanced users point Handy at a local OpenAI-compatible transcription endpoint when the model they need cannot be loaded directly by Handy. I made this for my own fork because I needed it but wanted to offer the working version here in case this is useful as a future direction.

Knowing that Handy is in a feature freeze and that this may be totally outside the current scope of the project. I am totally fine closing this PR, moving this into a discussion first, or keeping it only in my fork if this is still outside the scope of Handy's future vision as well.

## Related Issues/Discussions

Fixes: N/A
Discussion:

- #19
- #77 
- #279 
- #466 
- #509 
- #804 
- #886 
- #1050 
- #1131 

## Community Feedback

I have not gathered new community feedback for this specific implementation yet.

I did search through prior issues and PRs, and I understand this overlaps with earlier rejected attempts to add API or OpenAI-compatible transcription support. In particular, #886 and #1050 make it clear that Handy has not wanted to support remote or other-provider transcription, and #19 discussed custom whisper-compatible endpoints earlier.

The distinction I am trying to make is that this is meant for local/self-hosted model runtimes rather than hosted transcription providers. Handy currently supports Whisper GGML `.bin` models and several built-in `transcribe-rs` ONNX engines, but not arbitrary Hugging Face ASR models or new architectures like MedASR. If that distinction still does not fit Handy, I completely understand and can close the PR.

## Testing

Tested on Windows, but not tested on other platforms, as I don't have access to other devices to test on. I built Handy both in dev and as a production release. Configured a local endpoint at `http://localhost:8000/v1/audio/transcriptions` and verified Handy can use the custom endpoint with a local server.

Screenshot 2026-05-26 035805

I am not fully confident about the current UI placement, but based on prior discussion around this topic, I agree this kind of advanced endpoint support should stay outside the primary UI. I placed it in Debug for now because it is most likely useful only for niche or advanced users, and it may make more sense as an Experimental setting if it is considered at all.

## AI Assistance

I used Codex to help inspect and review the Handy codebase, draft and revise various different implementation and UI placements in Handy, debug local server-side endpoint issues, and review the diff for behavior changes before opening the PR.

### DIFF
--- a/src-tauri/Cargo.lock
+++ b/src-tauri/Cargo.lock
@@ -1935,6 +1935,7 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "07bbe89c50d7a535e539b8c17bc0b49bdb77747034daa8087407d655f3f7cc1d"
 dependencies = [
  "futures-core",
+ "futures-sink",
 ]
 
 [[package]]
@@ -3330,6 +3331,16 @@ name = "mime"
 version = "0.3.17"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "6877bb514081ee2a7ff5ef9de3281f14a4dd4bceac4c09388074a6b5df8a139a"
+
+[[package]]
+name = "mime_guess"
+version = "2.0.5"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "f7c44f8e672c00fe5308fa235f821cb4198414e1c77935c1ab6948d3fd78550e"
+dependencies = [
+ "mime",
+ "unicase",
+]
 
 [[package]]
 name = "minicov"
@@ -4869,6 +4880,7 @@ dependencies = [
  "base64 0.22.1",
  "bytes",
  "encoding_rs",
+ "futures-channel",
  "futures-core",
  "futures-util",
  "h2",
@@ -4882,6 +4894,7 @@ dependencies = [
  "js-sys",
  "log",
  "mime",
+ "mime_guess",
  "native-tls",
  "percent-encoding",
  "pin-project-lite",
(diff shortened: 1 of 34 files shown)


cjpais:
I appreciate the thoughtful PR, and definitely MedASR is on my radar and something that will make it into Handy directly

There's been a lot of discussion and custom transcription endpoints will not be added. I am working on a library to replace the existing onnx and whisper inference engines and it will support MedASR day 1 when it is merged into Handy

arvinmi:
Thanks for the response. I'm glad to hear MedASR is on your radar and that direct support is planned through a new inference library, that would solve the use case much better than a custom endpoint anyway.

PR #1436, opened 2026-05-26
Add silence stop and overlay controls

deut-lab:
## Summary
- Add a setting to stop recording after a configurable silence timeout.
- Add a calmer overlay style, plus a classic style option.
- Add an overlay action to finish recording and paste text without sending the submit key.

## Validation
- npm run build
- npm run lint
- npx tsx scripts/check-translations.ts
- git diff --check

## Notes
- Rust tests were not run locally because `cargo` is not available in PATH on this machine.


cjpais:
Features not being pulled in as the PR template notes

PR #1447, opened 2026-05-29
fix: eliminate WebKitWebProcess memory leak from overlay mic-level events (#1279)

### LINKED ISSUE #1279: [BUG] WebKitWebProcess memory grows continuously during transcription on Linux
## Summary

On Linux, Handy's `WebKitWebProcess` accumulates memory during transcription use and never releases it. Over a normal day of voice-to-text work, this can consume all available RAM and swap, eventually triggering the OOM killer.

## Before submitting

Searched open and closed issues for "memory leak", "memory", "WebKit memory", "OOM", "swap", and "WebKitWebProcess" -- no prior reports of this behavior. Also scanned all GitHub Discussions titles for related topics.

## System Information

- **App Version:** v0.8.2 AppImage (latest release as of 2026-04-12)
- **Operating System:** Ubuntu 24.04, Linux 6.17.0-19-generic
- **CPU:** 13th Gen Intel Core i7-13700F
- **GPU:** NVIDIA GeForce RTX 4080
- **Tauri:** 2.10.2 (from `src-tauri/Cargo.toml`)
- **WebKitGTK (system):** 2.50.4-0ubuntu0.24.04.1
- **Settings:** `always_on_microphone: false`, `lazy_stream_close: false`, `overlay_position: none`

## Logs

The application log from the current session is here: https://gist.github.com/captivus/fe2f144e9014a9f640ea40138e182c4c

It shows normal operation with no memory-related errors. The log from the 36-hour session that triggered the OOM kill did not survive -- Handy appears to overwrite the log on each launch. The log directory is `~/.local/share/com.pais.handy/logs/`.

### Memory growth over time

After a fresh launch, I tracked the main WebKitWebProcess RSS during normal voice-to-text use:

| Time | Event | RSS |
|------|-------|-----|
| 09:15 | Fresh launch | 249 MB |
| 09:56 | After normal use (~41 min) | 527 MB (+278) |
| 10:02 | 2 min idle, no transcriptions | 539 MB (held) |
| 10:11 | After more transcriptions | 654 MB (+115) |
| 10:13 | 2 min idle, no transcriptions | 654 MB (held) |
| 11:07 | ~2 hours after launch | 1,230 MB (+981) |

Over a previous 36-hour session with the same usage pattern, the same process reached 9.5 GB (1.6 GB RSS + 7.9 GB swap), at which point the OOM killer terminated it along with several other applications. That data point is from a single uncontrolled observation, so I mention it for context rather than as precise evidence.

Note: growth rates vary across measurement windows (~6.8 MB/min in the first window, ~12.8 MB/min in the second). This likely reflects variation in my transcription activity -- some periods I was transcribing more frequently than others. I didn't record the exact number or duration of transcriptions in each window.

### The growth pattern

- **Grows during transcription use** -- each transcription session adds measurable RSS
- **Holds during idle** -- after recording stops and the audio stream closes, memory does not decrease
- **Staircase pattern** -- repeated cycles of grow-then-hold, never reclaiming

All growth is in `Private_Dirty` anonymous memory (measured via `/proc/<pid>/smaps_rollup`). No growth in shared libraries, file-backed mappings, or shared memory.

### What I investigated

I initially suspected WebKitGTK's `evaluate_javascript` was leaking, since Handy emits events at high frequency during recording. I wrote a standalone test (no Tauri, no React -- just a bare WebKitGTK WebView calling `run_javascript` at a similar rate with a similar payload).

With a proper `GAsyncReadyCallback` that calls `run_javascript_finish()`, the standalone test shows **zero memory growth** over 11,280 evaluations at 94 calls/sec for 2 minutes. This rules out one specific failure mode: a `GTask` result accumulation from missing `_finish()` calls at the C API level.

However, there are significant limitations to this test:
- **Duration:** At the observed growth rate (~6.8 MB/min), 2 minutes would produce only ~14 MB of growth -- likely within JSC's garbage collection noise. A conclusive test would need 15-30 minutes minimum.
- **Payload:** The test evaluated trivial JavaScript on a bare HTML page. The real workload involves Tauri's JS bridge deserialization, event registry dispatch, React `setState` at ~188 calls/sec, VDOM diffing, and DOM style mutations -- none of which the test exercises.
- **FFI layer:** My test used Python's gi bindings, not Tauri's wry Rust FFI. This isn't just a calling convention difference -- it determines whether `evaluate_javascript_finish()` is called, how the JS source string is constructed, and how the GLib memory lifecycle is managed. The Python test cannot rule out a leak in wry's specific FFI code path.

The JSC heap within WebKitWebProcess is itself WebKitGTK memory, so this test does not exonerate WebKitGTK broadly -- only the narrowest interpretation of a raw C API leak.

A similar leak pattern has been reported on Windows/WebView2 in tauri-apps/tauri#12724, which suggests the accumulation might be in Tauri's cross-platform event dispatch layer rather than in either WebView backend specifically. I can't confirm this.

### Limitations

I don't have a Rust build environment, which means I cannot:
- Build a modified Handy with event emission disabled to isolate the cause
- Profile the JavaScript heap to identify what objects are accumulating
- Test any of my suggestions below before proposing them

The correlation between transcription activity and memory growth is clear from the data, but I cannot isolate which aspect of the transcription flow is responsible. During recording, several things happen simultaneously: audio events are emitted, the Whisper model runs inference, transcription results are dispatched, React components update, and history entries are created. Any of these could contribute.

## Possible contributing factor: high-frequency event emission

One candidate that stands out in the code is `emit_levels()` in `src-tauri/src/overlay.rs:388-396`:

```rust
pub fn emit_levels(app_handle: &AppHandle, levels: &Vec<f32>) {
    let _ = app_handle.emit("mic-level", levels);
    if let Some(overlay_window) = app_handle.get_webview_window("recording_overlay") {
        let _ = overlay_window.emit("mic-level", levels);
    }
}
```

This fires on every audio visualizer callback. With a 512-sample window (`WINDOW_SIZE` in `recorder.rs`), at a typical 48kHz sample rate (device-dependent -- could be higher or lower), that's approximately 94 calls/sec during recording. A few observations about this function:

1. **Both `.emit()` calls are global broadcasts.** In Tauri 2, `Emitter::emit()` on a `WebviewWindow` is not targeted to that window -- the targeted method is `emit_to()`. This means the overlay receives the event twice per callback.

2. **Only the overlay listens for `mic-level`** (`RecordingOverlay.tsx:41`). Tauri's event dispatch filters by registered listeners before calling `evaluate_script`, so the main window's WebView is not evaluated -- but the overlay's WebView is evaluated twice per callback (once from each `emit()` call), for approximately 188 `evaluate_script` calls/sec during recording.

3. **No `overlay_position` guard.** The function doesn't check the overlay setting. `create_recording_overlay` (in `lib.rs`) creates the overlay window at startup regardless of `overlay_position`, so the hidden overlay WebView exists and receives events even when the user has disabled the overlay. On Linux, `overlay_position` defaults to `None` (see `default_overlay_position()` in `settings.rs`), so this affects all Linux users who haven't explicitly enabled the overlay -- which is likely most of them.

4. **`always_on_microphone: true` would be worse.** With that setting, the audio stream stays open continuously and events would fire 24/7, not just during recording.

I want to be clear: I don't know that this is the cause. It's a plausible contributor because of the volume of events and the correlation with recording activity, but other parts of the recording pipeline could also be responsible.

## Suggested improvements

These would reduce event volume significantly and are arguably good engineering changes regardless of whether they address this specific leak:

1. **Throttle `emit_levels` to ~15-30 calls/sec.** The overlay animation doesn't need the full audio callback rate. There's already a time-based throttle pattern in `managers/model.rs:1121` (model download progress at 100ms intervals) that could serve as a template. I haven't tested what frame rate the level meter visualization needs to look smooth -- the maintainer would know better what rate feels right.

2. **Use `emit_to("recording_overlay", "mic-level", levels)` instead of two global `.emit()` calls.** This would target only the overlay window and eliminate the redundant second broadcast.

3. **Skip emission when `overlay_position` is `none`.** An early return would eliminate all event processing for users who don't use the overlay. The setting would need to be cached rather than calling `get_settings()` on every callback.

### A test the maintainer could run

If you're interested in isolating whether the event volume is the cause: temporarily no-opping the body of `emit_levels` in `overlay.rs` and testing whether memory still grows during recording would definitively answer that question. (Commenting out only the `listen("mic-level", ...)` handler in `RecordingOverlay.tsx` would eliminate the React state updates but the Tauri JS bridge would still receive and process the events, so that would be a less conclusive test.)

## Reproduction

1. Launch Handy v0.8.2 on Linux
2. Identify the WebKitWebProcess PIDs: `ps --forest -eo pid,ppid,rss,comm | grep -E "handy|WebKit"`
3. Record the RSS of the largest WebKitWebProcess: `grep VmRSS /proc/<pid>/status`
4. Do 10-15 transcriptions over ~10 minutes of normal use
5. Record the RSS again
6. Wait 2-3 minutes without transcribing
7. Record the RSS a final time -- it will not have decreased

The standalone WebKitGTK test script (v2, with proper callback handling and multi-variant support) is here: https://gist.github.com/captivus/2d3d07c6d36f219035c26aad55c043bf

## Related

- tauri-apps/tauri#12724 -- a similar pattern of memory growth from event emission, reported on Windows/WebView2. May indicate the accumulation is in Tauri's event dispatch layer.

captivus:
## Before Submitting This PR

**Not applicable** -- this is a bug fix for an open, maintainer-acknowledged issue, not a previously-closed change.

## Human Written Description

I use Handy as my daily driver voice transcription tool. Shortly after adoption, I encountered a particularly nasty memory leak that, when OOM-killed, destabilized my Linux session. I filed [#1279](https://github.com/cjpais/Handy/issues/1279) with detailed analysis of the issue and speculation on potential fixes. The source of the memory leak is, it turns out, upstream of Handy in WebKit but, by thoughtfully scoping the emission of microphone-level events to the visualizer overlay, the leak can be mostly eliminated. I submit this patch / pull request to resolve the leak.

## Related Issues/Discussions

Fixes #1279

Discussion: the issue thread itself -- <https://github.com/cjpais/Handy/issues/1279>.

I also wrote up the full story (and the reproducible methodology) [here](https://gallon.me/95-gb-in-36-hours-how-i-fixed-handys-memory-leak-without-knowing-rust).

## Community Feedback

This is a bug fix for an issue I filed, which the maintainer ([@cjpais](https://github.com/cjpais)) acknowledged on 2026-04-13:

> *"This is a problem, and we need to fix it. Just acknowledging it"*

When I asked whether a PR would be welcome:

> *"Definitely, there may be an outstanding PR already but I would love a PR"*

One other user ([@scr4bble](https://github.com/scr4bble)) reported a related-but-distinct observation in the same thread (memory not dropping after Cancel from a long recording). That is **not** addressed here -- it looks like a separate concern.

**On the newly-raised PR (#1444).** A PR ([#1444](https://github.com/cjpais/Handy/pull/1444)) was raised today which also targets this issue by throttling `mic-level` emissions to ~30 FPS. I began working on this pull request nearly 2 weeks ago and have just concluded today in the ~ 6 day patched build soak described below. I've seen the alternative PR, but I think this PR is the better fix -- not as a knock on that work, but because the measurements say so. I bisected three candidate fixes (skip-when-disabled, targeted `emit_to`, and throttling) against a controlled workload, isolating each one. **Throttling showed no effect distinguishable from measurement noise**, which is why I dropped it. 

One note on the rate, since Hz and FPS are the same unit here -- each emission drives one visualizer frame: my instrumented measurement put the natural emit rate at ~24 Hz, so the throttle I tested capped it *further*, to 20 Hz, and it still came out as noise. A 33 ms / ~30 FPS cap sits *above* my measured ~24 Hz rate, so on my data it would drop essentially nothing -- which leaves the targeted `emit_to` as the only component of that approach doing real work, and this PR already includes it. (It's also noteworthy that I initially proposed this `emit_to` target in the initial analysis I filed in [#1279](https://github.com/cjpais/Handy/issues/1279).) 

The lever that actually matters is skipping emission entirely when the overlay is disabled -- and since the overlay is off by default on Linux, that's the configuration most users are running. Even with a throttle in place, ~24 events/sec still hit the hidden WebKit subprocess that's driving the leak; it slows the bleed at best, it doesn't stop it. This PR takes that subprocess to zero emissions for the default config, and keeps the `emit_to` improvement for users who do enable the overlay. The numbers backing all of this are in the Testing section below.

## Testing

I deliberately scoped this fix to Handy's own code. The ultimate source of the leak appears to be upstream in WebKit; I did not try to analyze or patch WebKit itself.

**What's going on.** Handy's recording overlay (the voice visualizer, off by default) is driven by `mic-level` events that `emit_levels` broadcasts at the audio-callback rate (~24 Hz) during recording. On Linux the `recording_overlay` window is created at boot *regardless* of `overlay_position` -- when the overlay is disabled it exists but is never shown. The hidden overlay's WebKit subprocess still receives every event and leaks memory on each one. The actual accumulation is in WebKit's C++ side, upstream of Handy, so the only lever reachable from Handy's code is to reduce the event traffic that drives it.

**How I found it.** I built a small reversible harness so my tooling could run Handy end-to-end on a dedicated Xvfb display with deterministic test audio (a PipeWire null sink) and a `SIGUSR2` trigger, then instrumented the runtime with four tools at once -- `uftrace` (Rust calls), `strace` (syscalls + IPC), the WebKit Remote Inspector (JS heap), and a `smaps_rollup` sampler (per-process memory). That narrowed eight candidate mechanisms to one:

- `smaps_rollup` localized the sustained growth to the overlay's WebKit subprocess (the Rust process and the settings webview were flat after warmup).
- A WebKit `Heap.snapshot` diff showed the JS heap accounts for only **~1.2%** of the growth, and 100% of the growth is in private-anonymous pages -- so the leak is in WebKit's native C++ side, **by exclusion** (I did not directly characterize the C++ allocation; that would need `heaptrack` on the WebKit subprocess, outside the scope of a Handy change).
- `uftrace`/`strace` confirmed the driver: `emit_levels` firing at ~24 Hz and double-broadcasting to the hidden overlay.

**The fix (two parts).**

1. **Skip emission when the overlay is disabled.** Cache `overlay_position != None` in an `AtomicBool` (populated at startup in `lib.rs`, kept in sync by `change_overlay_position_setting` in `shortcut/mod.rs`). `emit_levels` does a single relaxed atomic load and returns early when the overlay is off. The audio path no longer touches the Tauri store at ~24 Hz, and for the default Linux config the leak's driver goes silent entirely.
2. **Target the overlay when it *is* enabled.** In Tauri 2 both `AppHandle::emit` and `WebviewWindow::emit` are global broadcasts; the old code produced two `eval_script` dispatches to the overlay per callback. Replacing them with a single `emit_to("recording_overlay", "mic-level", levels)` halves the per-callback WebKit dispatch work.

**Before → after, measured.** I validated with a bisect: the same controlled 50-cycle workload, each fix component behind its own env-var toggle, measuring sustained-phase RSS growth (cycle 10→50) of the overlay's WebKit process.

- **Overlay disabled (`overlay_position: none` -- Linux default):** unfixed control ≈ **4.5 MB/min** of sustained growth; with the fix, growth drops to the measurement-noise floor (effectively flat). This agrees within ~10% with my real-world observation of ~4.0 MB/min sustained (249 MB at launch → 1,230 MB after ~2 h → 9.5 GB / OOM after ~36 h). After applying the fix, this same workload measured 0.0 KB/min of sustained growth with both fixes enabled (and −4.3 KB/min with the skip-when-disabled change alone) -- both statistically indistinguishable from zero. Per recording cycle that's roughly 0 KB, against the control's ~5,585 KB/cycle: the overlay process simply stops growing.
- **Overlay enabled:** the targeted `emit_to` reduced the leak rate by **~33%** (single trial -- weaker evidence than the `none` result, but directionally consistent with halving the dispatch work). This is quite acceptable, as the vast majority of the memory leak was triggered by the overlay disabled scenario.
- **The six-day soak.** Numbers from a lab workload aren't enough for a leak that takes hours to bite, so I ran the patched build as my actual daily driver for 5 days 18.5 hours across 7 launches, with a 30-second `smaps_rollup` poll logging per-process memory to CSV the whole time. It stayed flat, proving (in meatspace) that the fix works.

**Caveats and limitations:**

- This addresses the leak's *driver* (event traffic), not the *mechanism* of WebKit's C++ accumulation, which I did not directly characterize.
- The fix is platform-shared (no `#[cfg]` guards). I tested on **Linux only**; macOS/Windows behavior should be unchanged but I haven't verified -- please flag any platform-specific concerns.
- The `OVERLAY_ENABLED` cache defaults to `false` and is populated in `lib.rs::setup`. This is safe today because recording can't fire before setup runs; a future change that started recording earlier in the lifecycle would need to keep that ordering in mind.
- `hide_recording_overlay` still emits one `hide-overlay` event when the overlay is disabled. That's a bounded residual (one event per recording-stop vs. the hundreds of `mic-level` events the early-return now skips) and not worth addressing here.
- A third candidate fix -- throttling `emit_levels` from ~24 Hz to 20 Hz -- was evaluated and **dropped**: the analysis showed no effect distinguishable from control noise, so shipping it would add complexity for no measured benefit.

N/A -- backend behavior change, no UI surface changes.

## AI Assistance

**If AI was used:**

- **Tools used:** Claude Code (Anthropic) with a heavily customized multi-agent harness.
- **How extensively:** extensively, and disciplined -- not vibe coding. 

**The short version**:

I applied [The Framework](https://vibecodinghangover.com/) -- a comprehensive engineering discipline for AI-accelerated coding -- to diagnose and fix this leak.

- Developed and implemented an agent autonomy harness so that my AI agents could run Handy end-to-end without interrupting my work (this was massive)
- Instrumented the Handy runtime for observability and agent sensory feedback (also massive)
- Used this instrumentation to analyze and narrow systematically to isolate the source of the leak
- Fixed the leak in 3 ways, and tested each bit of the fix on its own, and together, measuring how each contributed (one of the candidate fixes was dropped after seeing its minimal contribution)
- Deployed adversarial agents to scrutinize findings
- Surfaced a patched version of the app for me to run for days to prove, via additional telemetry, that the fix worked (the important meatbag-in-the-loop bit)

Every claim in this description was independently verified against either the source code or the raw measurement data -- agent output was never trusted without recompute, which mattered because I don't have Rust familiarity to fall back on. The investigation produced large local artifacts (instrumentation procedures, per-run captures, statistical analysis, three rounds of adversarial review) that are **not** in this PR -- they're intermediate data, not appropriate for the repo.

**The long version**: I've written the entire process up [here](https://gallon.me/95-gb-in-36-hours-how-i-fixed-handys-memory-leak-without-knowing-rust).

### DIFF
--- a/src-tauri/src/lib.rs
+++ b/src-tauri/src/lib.rs
@@ -542,6 +542,14 @@ pub fn run(cli_args: CliArgs) {
 
             initialize_core_logic(&app_handle);
 
+            // Populate the overlay-enabled cache from initial settings so the
+            // audio path (overlay::emit_levels, called ~24 Hz during recording)
+            // can do a single atomic load instead of reading the Tauri store.
+            // Kept in sync by shortcut::change_overlay_position_setting.
+            overlay::update_overlay_enabled_cache(
+                settings.overlay_position != settings::OverlayPosition::None,
+            );
+
             // Pre-warm GPU/accelerator enumeration on a background thread.
             // The first call into transcribe_rs::whisper_cpp::gpu::list_gpu_devices
             // loads the Metal/Vulkan backend and probes devices, which can take
--- a/src-tauri/src/overlay.rs
+++ b/src-tauri/src/overlay.rs
@@ -1,6 +1,7 @@
 use crate::input;
 use crate::settings;
 use crate::settings::OverlayPosition;
+use std::sync::atomic::{AtomicBool, Ordering};
 use tauri::{AppHandle, Emitter, Manager, PhysicalPosition, PhysicalSize};
 
 #[cfg(not(target_os = "macos"))]
@@ -385,12 +386,41 @@ pub fn hide_recording_overlay(app_handle: &AppHandle) {
     }
 }
 
-pub fn emit_levels(app_handle: &AppHandle, levels: &Vec<f32>) {
-    // emit levels to main app
-    let _ = app_handle.emit("mic-level", levels);
+// Cached "overlay is enabled" flag, kept in sync with the
+// overlay_position setting. Avoids reading the Tauri store on every
+// audio callback (~24 Hz during recording). Defaults to false so the
+// audio path doesn't emit until lib.rs::setup populates the cache from
+// initial settings.
+static OVERLAY_ENABLED: AtomicBool = AtomicBool::new(false);
 
-    // also emit to the recording overlay if it's open
-    if let Some(overlay_window) = app_handle.get_webview_window("recording_overlay") {
-        let _ = overlay_window.emit("mic-level", levels);
+/// Update the cached overlay-enabled flag. Called from `lib.rs` at
+/// startup after settings load, and from `change_overlay_position_setting`
+/// whenever the user changes the overlay position.
+pub fn update_overlay_enabled_cache(enabled: bool) {
+    OVERLAY_ENABLED.store(enabled, Ordering::Relaxed);
+}
+
+pub fn emit_levels(app_handle: &AppHandle, levels: &[f32]) {
+    // Skip emission when the overlay is disabled. The recording_overlay
+    // window is created at boot regardless of overlay_position, so
+    // without this guard a hidden overlay's WebKit subprocess still
+    // processes every event. Each event drives some kind of WebKit
+    // C++ allocation that accumulates without bound (mechanism not
+    // directly characterized; see issue #1279 for the investigation).
+    // For users with `overlay_position: none` (the Linux default) this
+    // skip eliminates the upstream driver of that accumulation.
+    if !OVERLAY_ENABLED.load(Ordering::Relaxed) {
+        return;
     }
+
+    // Target only the overlay window. In Tauri 2 both `AppHandle::emit`
+    // and `WebviewWindow::emit` broadcast to all webviews; Tauri's
+    // listener filter then skips webviews with no registered listener
+    // for the event, so the settings webview never received `mic-level`.
+    // But the previous dual-call pattern still produced two `eval_script`
+    // calls to the overlay per audio callback (one from each .emit()).
+    // `emit_to` with the overlay's window label produces a single
+    // eval_script call per callback, cutting the per-callback WebKit
+    // dispatch work in half.
+    let _ = app_handle.emit_to("recording_overlay", "mic-level", levels);
 }
--- a/src-tauri/src/shortcut/mod.rs
+++ b/src-tauri/src/shortcut/mod.rs
@@ -550,6 +550,10 @@ pub fn change_overlay_position_setting(app: AppHandle, position: String) -> Resu
     settings.overlay_position = parsed;
     settings::write_settings(&app, settings);
 
+    // Keep the cached overlay-enabled flag in sync so emit_levels stops
+    // (or resumes) emitting on the next audio callback.
+    crate::overlay::update_overlay_enabled_cache(parsed != OverlayPosition::None);
+
     // Update overlay position without recreating window
     crate::utils::update_overlay_position(&app);
 


i3130002:
Nice work @captivus you're amazing. I should learn from you.

PR #1465, opened 2026-06-03
Apply paste delay after key press and increase slider range

sboys3:
## Before Submitting This PR

## Human Written Description

The main purpose of this is for people to test if increasing the delay after improves clipboard pasting reliability. This fix applies the existing paste delay config option to after the paste key presses. Both the delay before and after are now controlled by the one setting. The clipboard seems to be reverted before the key presses go through to some applications, resulting in previous clipboard content being pasted instead of the new content. This slider on the debug page has also been extended out to 500ms, because on my system when there is heavy load, it is required to be that high. The only change this should have for normal users is it changes the second delay from 50 to 60 milliseconds by default.

## Related Issues/Discussions

Discussion: https://github.com/cjpais/Handy/issues/502

## Testing

I have used this fix for several days now, during which some of the time my system was under heavy load, and I have not had a paste failure since. Although 1000 milliseconds in total(500ms for the setting) does block the main thread for a while. Previously, it was a common occurrence to get incorrect content pasted.

## AI Assistance

- [X] No AI was used in this PR


cjpais:
Is it possible to make these two separate sliders @sboys3 

Where: `std::thread::sleep(std::time::Duration::from_millis(50));` was not previously using any variable

hbenthow:
Is there an ETA of when this pull request might be implemented? I keep checking for updates to Handy and there haven't been any in a long time.

cjpais:
I am in the process of releasing transcribe.cpp and integrating into handy in parallel. After this is done I will be reviewing issues and merging appropriate features.

peterpeter7500-droid:
Hi - just checking that the fix above would address an odd issue (that I think others may have hit in a slightly different form).

Steps to reproduce:

Copy some text in another application (e.g. copy/paste for a separate task).
Switch to a text field and press the Handy push-to-talk hotkey to dictate.
On this first attempt after the copy, instead of my spoken words, the last thing on the clipboard gets pasted into the field.

Subsequent recordings in the same session work fine it only seems to happen on the first dictation after a copy operation. I've seen a few reports that sound related but not identical to this, so wanted to raise it in case it's a distinct edge case or a variant of the same root cause. Happy to provide more detail (OS, version, logs) if useful...just let me know what would help track it down.

cjpais:
It's largely this issue, and you can definitely use the debug menu to increase things. Though there are some other issues that are also related to this and that we're going to be addressing, please give more details with the OS version and logs if appropriate. But certainly on Windows and macOS, we are going to try to have an improved methodology coming soon

PR #618, opened 2026-01-18
Wake-Word

tachyonicbytes:
## Before Submitting This PR

**Please confirm you have done the following:**

- [X] I have searched [existing issues](https://github.com/cjpais/Handy/issues) and [pull requests](https://github.com/cjpais/Handy/pulls) (including closed ones) to ensure this isn't a duplicate
- [X] I have read [CONTRIBUTING.md](../CONTRIBUTING.md)

**If this is a feature or change that was previously closed/rejected:**

- [ ] I have explained in the description below why this should be reconsidered
- [ ] I have gathered community feedback (link to discussion below)

## Human Written Description

Hi! This is a Draft PR for the discussion of the WakeWord implementation. I chose a draft PR because the code is not production ready, and most likely won't be until we sort out all the details.

I started from the [openWakeWord](https://github.com/dscripka/openWakeWord) project, that seems to train wakewords, as well as provide an inference engine based on onnx to run them.

In order to understand how openWakeWord works, I stripped the entire logic and one example into a single file, very easy to understand, that I provide in this [github gist](https://gist.github.com/tachyonicbytes/850ffd8c7a02d7d87a44521592e46541) here.

Basically, there are three models, the melspectrogram model, the embedding model from google and the actual wakeword model. And all that's needed is chopping up the audio into suitable pieces (resampling if needed), feeding the input from one model into the other up until the wakeword model, which gives us the prediction. The "Hey Mycroft" model seemed to have the best activation from all the models included.

Secondly, I translated the code to Rust, as a standalone example, provided in this [github gist](https://gist.github.com/tachyonicbytes/1e4be48ef5029ad958642a282d406144) here. There are some more goodies than the simple example here. A project that already ports a form of openWakeWord to rust already [exists](https://github.com/skoky/oww_rs), but I couldn't make it work.

Using the functional code from this example, I was able to implement a very rudimentary wakeword function for Handy, using the already present "Always-On Microphone", that's a prerequisite for the wakeword to function.

I posit that the models themselves are small enough and the license is not problematic to distribute them along with the app, not making the user also download them, but closer inspection of the openWakeWord repo should be done in order to do so. Otherwise, we can just download them like we do for the transcription models.

There is an additional question of stopping the recording after the wakeword was uttered. For the simple example, I did a 5 second delay, but we should find something more permanent here.

[...]

PR #1469, opened 2026-06-03
feat: hands-free always-on capture with transcript-prefix command gate

jaredirish:
## What

Adds an opt-in **hands-free always-on capture** mode: the mic stays open, every utterance is logged to local history, and an utterance is pasted at the cursor **only when it starts with a wake word** (default "dude"), which is stripped before paste. Everything else is captured to history but never typed. No shortcut press.

Toggle lives in **Settings → Advanced → Experimental** ("Hands-Free Always-On Capture").

```mermaid
flowchart LR
    A["🎤 Always-on mic<br/>(no shortcut)"] --> B["Auto-segment<br/>Silero VAD"]
    B --> C{"Starts with<br/>wake word?"}
    C -->|"&quot;dude hello world&quot;"| D["Strip wake word<br/>paste: <b>hello world</b>"]
    C -->|"&quot;just thinking out loud&quot;"| E["ambient"]
    D --> G["⌨️ Text cursor"]
    D --> F[("Local history<br/>everything logged")]
    E --> F
    classDef cmd fill:#0d2a14,stroke:#2EA043,color:#fff;
    classDef amb fill:#21262d,stroke:#8B949E,color:#c9d1d9;
    class D,G cmd;
    class E amb;
```

Green path pastes at the cursor; grey path is logged only. Everything lands in local history; only wake-word phrases reach the cursor.

## Relationship to #618 (Wake-Word)

This is a **lighter-weight, complementary** take on #618. Where #618 uses an acoustic openWakeWord ONNX model, this gates on the **transcript prefix** — no extra model to load, any wake word works, and it reuses the in-tree Silero VAD for segmentation. It deliberately reuses #618's `with_speech_frame_callback` scaffolding idea. Happy to align with #618 if you'd prefer a single path.

## How it works

- New serde-defaulted settings: `hands_free_capture`, `wake_word`, `wake_word_required_for_paste`, `capture_all_to_history`.
- Continuous Silero-VAD segmentation drives a single serialized worker; never blocks the audio callback.
- `wake_word_required_for_paste: false` → pure always-on dictation (pastes everything).
- Whole-word leading match: a mid-sentence "okay dude…" does **not** trigger paste.

## Tests

73 unit tests pass, including 8 for `strip_wake_word` (the no-paste-spray guarantee) and 3 for the VAD segmenter.

## Notes

- Backend validated on Linux (Docker, CI-style); `.app` built and live-tested on macOS (Apple Silicon).
- Includes a small `build.rs` escape hatch (`HANDY_DISABLE_APPLE_INTELLIGENCE=1`) so Command-Line-Tools-only toolchains (no full Xcode → no `FoundationModels` macro plugin) can still build the stub. Trivial to drop if unwanted.

---

Draft for discussion — happy to reshape around #618.


koloved:
DUDE, Do you have a build for Windows? I can test it. What is the additional CPU load as a result of using this function?

cjpais:
I can kick off a build later for you

koloved:
@cjpais Yes, that would be great. Of course, I can try to put this together in my branch along with my functionality. But I think it's better to test the clean version without my vibe coding

jaredirish:
Dude, I have the Mac Build, do you need me to post it? works pretty well!

koloved:
@cjpais @jaredirish I check the box in the settings for this feature as described in the description, but the recording still doesn't start on the keyword. No matter how I say it: loudly, quietly, close to the microphone, far away, in different variations, whatever I say, nothing happens.

jaredirish:
I'm not sure the bug, if I boot Handy, make Parakeet active, and then hit the hands free capture, it does both hands free and DUDE functions.   What was the situation 

> @cjpais @jaredirish I check the box in the settings for this feature as described in the description, but the recording still doesn't start on the keyword. No matter how I say it: loudly, quietly, close to the microphone, far away, in different variations, whatever I say, nothing happens.

koloved:
@jaredirish I have no idea what's wrong, buddy. But I did everything exactly as you said. No matter how I said it, nothing happened.

DomenickD3:
I also had the same experience of it not working. I'm on Windows 11 x64.

I forked the fork, and applied suggested fix by codex in src-tauri/src/commands/audio.rs: https://github.com/DomenickD3/Handy/commit/1c50f6a60a8d3884851a9ba2e742be369ff3f810#diff-758f6da1961e1c3c36b9dff244edfbe4c7f341ee1fda61fd2cfd5af2ccf43447.

This resolved the issue for me. I haven't had time to read through the rest code to be knowledgeable if this is an appropriate fix.

koloved:
@jaredirish Can you add this fix or look into it so we can test everything? And if it works, then we can roll out the updates too.

jaredirish:
I'm not sure the bug, if I boot Handy, make Parakeet active, and then hit the hands free capture, it does both hands free and DUDE functions.   What was the situation 

> @cjpais @jaredirish I check the box in the settings for this feature as described in the description, but the recording still doesn't start on the keyword. No matter how I say it: loudly, quietly, close to the microphone, far away, in different variations, whatever I say, nothing happens.

jaredirish:
Update: found the actual bug behind this. `start_hands_free()` was silently swallowing a failed microphone-open — the toggle would flip to "on" and the setting would persist even when the mic never actually opened, so the wake word had nothing to listen to. Fixed in ee07d26: the start command now only marks the setting enabled once the loop has genuinely started, and surfaces a real error to the UI otherwise.

Also fixed a separate frontend bug where the toggle always showed "off" on load regardless of actual backend state (28c85af), which made it look like it needed re-enabling on every restart.

@DomenickD3 — your fix targeted `update_microphone_mode`, which doesn't exist anymore after upstream's transcribe-cpp refactor, but it was the same root-cause shape, just in a different function.

Branch is rebased on current `main` and passing `cargo test --lib` (282 tests) + a live macOS test — VAD segmentation, wake-word gating, and capture-all-to-history all confirmed working end to end. @koloved / @DomenickD3 — if either of you get a chance to re-test (Windows especially, since I've only verified on macOS), that'd help a lot before this comes out of draft.

PR #1487, opened 2026-06-09
fix(paste): release the modifier keys after pasting with wtype

Yusyuriv:
## Before Submitting This PR

## Human Written Description

I'm using `Ctrl+Shift+V` paste method. I noticed that when I press the hotkey, dictate the text, and release the hotkey, the next time I press the hotkey, it does nothing, Handy doesn't pop up. However, if I press the hotkey, wait for the popup, and release it without speaking, it keeps working. Pressing Super after transcription also fixes this bug until the next transcription. This made me think the problem is in the pasting code. I found out that when using `wtype`, the code correctly holds the modifier keys when pasting (`-M ctrl -M shift -k v`), but never releases the modifier keys after. I made it release the modifiers after pasting: `-M ctrl -M shift -k v -m shift -m ctrl`. [Related documentation](https://man.archlinux.org/man/extra/wtype/wtype.1.en#OPTIONS)

## Related Issues/Discussions

I haven't found any.

## Testing

I ran Handy before this change and after. Before this change, the bug was present. After this change, Handy keeps working as expected, and the bug is no longer present.

## AI Assistance


cjpais:
thanks

PR #1488, opened 2026-06-09
Add setting for mic fallback when in clamshell mode 

anchpop:
Note: This PR is only intended to change the behavior for MacOS. The same principles apply to windows and linux, but I don't have testing environments for those two.

## Problem

**For clarity, I'll restate the reason we want a clamshell mode in the first place:**

I use airpods with my mac, but it's annoying when the the airpods' mic is enabled because it lowers the audio quality. So I set handy to use the built-in mic.

But I also sometimse use my mac in clamshell mode (lid closed + external display). In that case, I have no choice but to use the airpods, beacuse the built-in mic can't pick up anything when the lid is closed. 

This basically meaens I have to keep going in and out of the settings. 

**But we already have a settng for this**

I found out while implementing this that there's already a debug option that does the same thing as what I implemented. But I didn't know that before, and I would assume most handy users don't either. You have to enable debug mode (which I didn't know about), then go to the debug tab and enable it. 

So the real problem solved by this PR is that the settings for clamshell mode are difficult to find 

## Community interest

Clamshell settings were requested in #326, implemented in #329, and the implementation was tweaked in #348. Given this, I think there is a moderate amount of interest in this.

## Solution

A new collapsable menu in the main settings page, only visible on macs, with clamshell settings.

Menu closed: 
CleanShot 2026-06-09 at 01 31 54@2x

Menu open: 
CleanShot 2026-06-09 at 01 32 03@2x

With this, we can also remove the the clamshell mic picker in the debug settings.

## Before Submitting This PR

## Human Written Description

(This entire PR description was written by me or adapted from the template, no AI was used to write it.)

## Testing

TBD

## AI Assistance

- [x] AI was used

**If AI was used:**

- Tools used: Claude, Codex
- How extensively: Implementation written by Claude and reviewed by Codex. I have looked at the code, but not yet as extensively as I plan to, which is why the PR is a draft. I also want to refine the UI a bit.


anchpop:
Closing because I'm thinking this is probably too niche a feature to justify including in the front page

PR #1497, opened 2026-06-10
Add configurable models storage directory

### LINKED ISSUE #1496: Feature Request: Configurable models storage directory (default to install folder instead of AppData on C:)
## Problem

On Windows, language models are always stored in the user profile AppData directory:

```
C:\Users\{username}\AppData\Roaming\com.pais.handy\models\
```

This is hardcoded via `app_data_dir()` in `ModelManager::new()` (`src-tauri/src/managers/model.rs`).

For many users this is inconvenient because:
- The system drive (C:) may have limited free space, while models are large (hundreds of MB to ~1.1 GB each)
- Users may prefer to keep models on another drive/partition (D:, external SSD, etc.)
- Portable installs on non-C: drives still end up storing models on C: unless portable mode was chosen at install time

## Current workaround (partial)

Handy already supports **portable mode** (`src-tauri/src/portable.rs`), where all user data (settings, models, recordings, logs) is stored in `{install_dir}/Data/` when a `portable` marker file exists next to the executable.

However, this is not a full solution for this use case:
1. It must be selected at install time (or manually by creating the marker file)
2. It relocates **all** app data, not just models
3. Normal installs have no UI setting to change the models path after installation

## Proposed feature

Add a **Settings → Models → Storage location** option that lets users choose where downloaded models are stored.

Suggested behavior:
- **Default for normal installs:** `{install_dir}/models/` (next to the executable), not AppData on C:
- **Allow custom path:** folder picker to choose any directory (e.g. `D:\AI\Models\Handy\`)
- **Migration:** when changing the path, offer to move existing downloaded models to the new location
- **Validation:** check free disk space and write permissions before saving the new path

Portable mode could continue to use `{install_dir}/Data/models/` as today, or respect the same setting within the portable data root.

## Why this helps

- Saves space on C: for users with small system SSDs
- Makes Handy easier to use on multi-drive setups without requiring a full portable reinstall
- Aligns with user expectations that large ML assets live near the application or on a chosen data drive

## Environment

- OS: Windows (also relevant for Linux/macOS users who want models outside `~/.config` / `~/Library`)
- App: Handy (Tauri)
- Affected area: model download, discovery, and custom model auto-detection

Thanks for this great offline speech-to-text tool!

presetslrdev:
## Summary
- Adds **Settings → Models → Storage location** with the current models path and a **Choose Folder** button
- Defaults to the installation folder (`{install_dir}/models`, or `{install_dir}/Data/models` for portable installs)
- Offers optional migration when picking a new folder (moves models and removes them from the old location)

Closes #1496

## Motivation
On Windows, models are stored on drive C: under AppData by default. Users with small system SSDs or multi-drive setups need models stored next to the application install directory.

## Test plan
- [ ] Open Settings → Models and verify the default path is the installation folder
- [ ] Use **Choose Folder** to pick another path and verify downloads land there
- [ ] Confirm the migration prompt moves models and removes them from the old folder
- [ ] Verify portable installs resolve to `{install_dir}/Data/models`


cjpais:
duplicate #874

Issue #660, opened 2026-01-24
[BUG] state lost: mute while recoding turns off mute when recording finished instead of restoring state


Morriz:
## Before You Submit

**Please search [existing issues](https://github.com/cjpais/Handy/issues) to avoid duplicates.** Your bug may already be reported! Right now it's just me maintaining this project so many issues can be overwhelming! Help me out by checking first.

## Bug Description

mute while recoding turns off mute when recording finished, instead of restoring previous state

## System Information

**App Version:**

v0.7.0

**Operating System:**

macOS latest

**CPU:**
Aplple M3 Max

**GPU:**

cjpais:
I'm sorry I don't understand the bug, can you walk through the flow step by step, along with your expectation? I just want to know clearly and explicitly what I should be reproducing

cjpais:
@Morriz what is the issue? I need more explanation otherwise this will be closed

PR #1417, opened 2026-05-20
fix: replace osascript with CoreAudio for macOS mute

t2o2:
## Before Submitting This PR

- [x] I have searched [existing issues](https://github.com/cjpais/Handy/issues) and [pull requests](https://github.com/cjpais/Handy/pulls) (including closed ones) to ensure this isn't a duplicate
- [x] I have read [CONTRIBUTING.md](https://github.com/cjpais/Handy/blob/main/CONTRIBUTING.md)

## Human Written Description

double initiation of recording when `Mute While Recording` is enabled on MacOS.

- Handy is able to mute immediately, however the recording is paused and then restarted again.
- Painful UX, I have to check the toolbar and make sure I only start speaking when the recording icon flashed up for the second time.

## Related Issues/Discussions

Other users are experiencing the same thing [issue 1101](https://github.com/cjpais/Handy/issues/1101)

## Testing

Built locally correctly and tested fine.

## Screenshots/Videos

Tested on macOS 15.5

Settings:
- Mute While Recording: enabled
- Audio Feedback: enabled
- Mode: toggle / push-to-talk

Before: microphone indicator blinks/reinitializes and first word is clipped.
After: no double initialization observed; first word is captured.

before:
https://github.com/user-attachments/assets/072af370-c202-42c6-88a1-93bd3b6f84ff

after:
https://github.com/user-attachments/assets/f1e1f5fc-5c4d-4f4f-95e6-e488fa720d5e

## AI Assistance

- [ ] No AI was used in this PR
- [x] AI was used (please describe below)

- Tools used: pi + gpt 5.5 + opus 4.6 (advisor)
- How extensively: one-shot

PR #672, opened 2026-01-26
fix: race conditions

joshribakoff:
## Before Submitting This PR

**Please confirm you have done the following:**

- [x] I have searched [existing issues](https://github.com/cjpais/Handy/issues) and [pull requests](https://github.com/cjpais/Handy/pulls) (including closed ones) to ensure this isn't a duplicate
- [x] I have read [CONTRIBUTING.md](https://github.com/cjpais/Handy/blob/main/CONTRIBUTING.md)

## Human Written Description

Implements state machine, single source of truth, declarative UI pattern, testing pattern, fixes race conditions at multiple layers. Trades off concurrency for simplicity, laying the foundation for future extensibility (and concurrency). Replaces jQuery style imperative UI management with React style declarative updates.

I am new to Rust and tried to do my best with AI, open to feedback / feel free to close and "steal" the idea but I would love feedback and/or credit!

## Related Issues/Discussions

Fixes #641, #462

## How to Reproduce

Three ways to trigger / different issue(s):

1. **During transcription**: Record something, release hotkey, then while transcribing trigger the hotkey again., global lock solves it
2. **Rapid toggle**: Quickly press and release the hotkey multiple times in succession capturing small empty audio packets rapidly, causes "deadlock" (not responding), global lock solves it.
3. **Hold and spam**: Hold down the hotkey key rapidly (same as 2) but UI panel disappears but audio cues indicates its still recording (which it is). Global lock does not solve it, required declarative UI.

## Testing

Tested locally by triggering all three reproduction scenarios above.

## AI Assistance

- [x] AI was used (please describe below)

**If AI was used:**

- Tools used: Claude Code
- How extensively: Pair programming to implement the global lock and declarative UI sync

---

The app has multiple sources of state that can get out of sync - different mutex locks at different layers, imperative UI updates with delays that don't get cancelled. This PR demonstrates a pattern that fixes the UI overlay and tray icon: a single source of truth state machine where UI syncs declaratively when the phase changes (similar to React's model).

This also fixes the "Not Responding" freeze by adding a global lock that prevents concurrent operations.

**Intentionally pragmatic**: I've kept this PR minimal to avoid overwhelming reviewers with huge changes. It focuses on fixing the immediate issue and demonstrating a pattern that could be extended to other parts of the app in future PRs.

Other state could be consolidated into this pattern in follow-up PRs:

- Audio feedback timing (stop playing one sound, play the latest sound instead, etc)
[...]

PR #1501, opened 2026-06-11
fix: stop 'Mute While Recording' from stranding system audio muted

newideas99:
## Before Submitting This PR

## Human Written Description

I listen to music a lot while I work, so I've been messing with the mute-while-recording behavior in Handy on my Mac. While testing it we found the audio could get stuck muted if you cancel a recording or tap the shortcut really quickly — this PR fixes those two cases so your audio always comes back.

## Related Issues/Discussions

Related (closed) report of the same symptom family: #660 — mute state not restored correctly after recording.

Fixes #
Discussion:

## Community Feedback

These are pre-existing bugs rather than a feature; happy to open a discussion if preferred.

## Testing

Tested on macOS 26 (Apple Silicon), both microphone modes (always-on and on-demand), with "Mute While Recording" enabled:

- **Cancel path**: start a recording with music playing, press the cancel shortcut → audio now unmutes immediately. Before this fix, in always-on mode (or with lazy stream close) audio stayed muted until the *next* completed recording, because only `TranscribeAction::stop()` called `remove_mute()`.
- **Quick press/release race**: tap the transcribe shortcut for <100ms. The delayed audio-feedback thread used to apply the mute *after* `remove_mute()` had already run, leaving the system muted on a dead session. With the generation counter, the stale apply is skipped (debug log: "Skipping mute: stream closed or session already ended").
- Normal recordings: mute applies after the start sound and is removed on key release, unchanged.
- `cargo fmt --check`, `cargo clippy`, `cargo test`, `bun run lint`, `bun run format:check` all pass.

N/A (audio behavior).

## AI Assistance

**If AI was used:**

- How extensively: Wrote most of the code under my direction; I identified the symptoms, tested the behavior on real hardware, and reviewed the changes.

### DIFF
--- a/src-tauri/src/actions.rs
+++ b/src-tauri/src/actions.rs
@@ -419,11 +419,12 @@ impl ShortcutAction for TranscribeAction {
             debug!("Always-on mode: Playing audio feedback immediately");
             let rm_clone = Arc::clone(&rm);
             let app_clone = app.clone();
+            let mute_generation = rm.mute_generation();
             // The blocking helper exits immediately if audio feedback is disabled,
             // so we can always reuse this thread to ensure mute happens right after playback.
             std::thread::spawn(move || {
                 play_feedback_sound_blocking(&app_clone, SoundType::Start);
-                rm_clone.apply_mute();
+                rm_clone.apply_mute(mute_generation);
             });
 
             if let Err(e) = rm.try_start_recording(&binding_id) {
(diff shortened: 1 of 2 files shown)


cjpais:
Have you looked at or reviewed other PR's which do the same thing? It's much better to contribute or test than just submit PR's which may already be implemented.

newideas99:
Fair point, I went back and did a much more thorough pass over open and closed PRs to double check (my first search was honestly not good enough). Closest things I found:

- #341 (merged): Windows/Linux mute stability, different platform and different failure mode
- #672 (closed in favor of #824): race conditions, but in the overlay/shortcut handling, not the mute path
- #1417 (open): replaces osascript with CoreAudio for the mute mechanism itself, doesn't touch these two bugs
- issue #660: same symptom family (mute state lost), but that was about the toggle state being cleared rather than restored

The two specific paths this PR fixes are: (1) cancelling a recording never restores audio in always-on mode or with lazy stream close, since only the normal stop path called remove_mute, and (2) a very quick press/release can have the delayed mute apply land after remove_mute already ran, leaving the system muted with nothing left to restore it. As far as I can tell no open or closed PR covers either path, but if I missed one please point me at it and I'll close this.

Also happy to rebase this onto #1417 if that's landing first, the fixes are mechanism-agnostic (they're about when mute/unmute is called, not how).

Issue #1431, opened 2026-05-25
[Feature] Smooth fade in/out option for "Mute While Recording"


booandrew:
### Feature request

Currently "Mute While Recording" mutes and unmutes the system output abruptly (a binary mute toggle in `set_mute()` — e.g. on macOS `osascript "set volume output muted true/false"`). When music or other audio is playing, it cuts off sharply and snaps back at full volume when recording stops.

It would be great to have an **optional toggle** to ramp the system volume **smoothly** instead:
- **fade out** the output volume to 0 (or to a configurable level) when recording starts
- **fade back in** to the previous level when recording stops

This makes the experience much less jarring, especially when listening to music through speakers.

### Possible approach
- Add a setting (e.g. `mute_fade_while_recording` + a duration, default ~150–300 ms) next to the existing `mute_while_recording` option.
- Instead of a single mute call, step the system output **volume** down/up over the configured duration (small interpolated steps). On macOS via `set volume output volume <n>`, on Windows via `IAudioEndpointVolume::SetMasterVolumeLevelScalar`, on Linux via `wpctl set-volume` / `pactl set-sink-volume`.
- Remember and restore the user's original volume level (related to the state-restore fix in #660).

Happy to keep it behind the existing experimental flag if preferred. Thanks for the great app!

cjpais:
Features are Discussions not issues.

Issue #1328, opened 2026-04-23
Pause background media playback when transcription starts


kas-cor:
## Feature Description
When the user starts a voice recording (toggle transcription), Handy should automatically pause any background media playback (e.g., Spotify, browser audio, system media). Once recording stops, playback should resume if it was previously active.

## Motivation
Background audio interferes with voice activity detection (VAD) and degrades transcription accuracy. It also creates a poor user experience — users typically don't want music or podcasts playing while they are dictating text. This is a common behavior in other dictation and voice assistant apps.

## Proposed Behavior
1. On `toggle-transcription` (recording start):
   - Detect active media playback across the system.
   - Pause playback and remember the "was playing" state.

2. On `toggle-transcription` (recording stop):
   - If media was playing before recording started, resume playback.
   - Otherwise, leave it paused.

## Technical Notes / Ideas
- Cross-platform media control could be handled via:
  - **macOS:** Media key events or AppleScript (`tell application "Spotify" to pause`).
  - **Windows:** `SendInput` for media keys or Windows Media APIs.
  - **Linux:** MPRIS D-Bus interface (`org.mpris.MediaPlayer2`).
- Alternatively, sending a global media-pause key event might be a simple cross-platform MVP.

## Related
- This pairs well with the existing `--toggle-transcription` CLI and global hotkey workflows.

cjpais:
I believe this already exists. It's currently mute audio while recording

Also feature requests belong in discussions

Issue #642, opened 2026-01-21
[BUG] muting the system audio has delay during recording


edwinencomienda:
Bug Description

When recording or dictating, the mute system audio feature has a slight delay that feels a bit off. If you try WhiprFlow it's instant.

System Information
App Version:

0.7.0

Operating System:

macOS Tahoe
Version 26.2

CPU:

Apple M1 Max

GPU:

Built-In

cjpais:
do you have 'audio feedback' also enabled? there's definitely a delay when that is there, otherwise it would mute the audio feedback too

we probably should evaluate the code path for if audio feedback is disabled and make sure it runs as fast as possible

zhukovgeorge:
@edwinencomienda @cjpais  
What is your average delay between pressing record and being able to start speaking?

On my machine, there is ~5 seconds of latency before audio is actually captured.  
Additionally, the visual input indicator is not very sensitive it only reacts when I’m very close to the microphone, which makes it unclear when it’s safe to start talking.

### Environment
- OS: Windows 11 Pro (Build 26200)
- Architecture: x64
- CPU: Intel (Family 6 Model 141)
- RAM: 32 GB
- Machine: Dell Precision 5760
- Virtualization: Hyper-V enabled

### Settings
![settings screenshot](https://github.com/user-attachments/assets/8bc2c285-7be5-4078-b223-83ca55c6f73c)

cjpais:
@zhukovgeorge on macos it's less than 500ms for me, I did just get a windows machine yesterday and it's very long, maybe 3 seconds. We need to fix this.

adrian-rpf:
I also experienced this on my MacBook M4 Pro - 14 inch - 24GB Memory - Nov 2024. 

Handy is so good this is one of the most annoying bugs when compared to Whispr.

cjpais:
@adrian-rpf which microphone are you using?

adrian-rpf:
@cjpais normally I'm using my builtin `macbook pro microphone`. 
I've also just tested it with out external USB mic, monitor mic, and using the system `Default` input (which is the build in macbook mic)

Tested on v0.9.6
MacOS Golden Gate 27.0

cjpais:
ty @adrian-rpf are you able to provide a log as well? about > log directory

NikhilVerma:
Yeah I just noticed this too and turning off "Mute While Recording" option makes it faster. However I still need to wait around 1s or so before I speak, otherwise the first words of my sentences are gone

PR #1503, opened 2026-06-11
feat: 'Volume While Recording' slider — duck audio instead of hard-muting (works on USB interfaces)

### LINKED ISSUE #998: [BUG] macOS 26.3 - 'Mute While Recording' doesn't work when using USB Audio Interface. Can handy have 'Pause While Recording' as an alternative?

newideas99:
## Before Submitting This PR

## Human Written Description

I tend to listen to music a lot especially while coding and I always found it annoying that I had to pause or turn down the volume of whatever I'm playing. This adds a Volume While Recording slider on macOS — set it above 0% and while you're recording it ducks your Mac's audio to that level, and as soon as you let go and it starts processing the transcript the audio goes right back to where it was, and everyone is happy. At the default 0% it behaves like the existing mute, except it now actually works on USB audio interfaces too.

## Related Issues/Discussions

Fixes #998 ('Mute While Recording' doesn't work on USB audio interfaces — the process-tap strategy works on exactly those devices, both for muting at 0% and ducking above it)
Related: #1431 (abrupt mute/unmute), #1328 (pause background media), #642 (mute delay — the tap engages without AppleScript latency)
Complements PR #1028 (pause-while-recording): pause suits resumable media; ducking covers live streams/calls and users who want music to keep playing. They can coexist.
Builds on the stranded-mute fixes in #1501.

Discussion: https://github.com/cjpais/Handy/discussions/1502

## Community Feedback

Discussion: https://github.com/cjpais/Handy/discussions/1502 (opened alongside this PR per CONTRIBUTING). #998 already has multiple affected users reporting exactly the problem this solves.

## Testing

Tested on macOS 26 (Apple Silicon), with a Native Instruments Traktor Kontrol S4 MK2 (USB interface with NO software volume/mute — the #998 scenario) and standard outputs:

- **Process tap tier**: verified the tap captures and replays all processes' audio at the configured gain through the aggregate device; buffer layout verified against a device that contributes its own input streams (tap buffer index = device input stream count). Verified the inert-tap case (permission not granted): the signal probe detects it and falls through to the AppleScript tiers instead of silently doing nothing. Handy's own feedback sounds are excluded from the tap and stay at full volume.
- **System volume tier**: at 0% uses the original mute flag; above 0% sets/saves/restores the output volume; skipped when output is muted or already at/below the duck level (never raises volume).
- **Per-app tier**: Spotify duck/restore verified live (89 → 10 → 89); browser-tab scripts verified against Chrome (skipped silently when "Allow JavaScript from Apple Events" is off).
- Quick press/release, cancel, and always-on/on-demand modes all restore audio.
- Slider UI: shown on macOS only (Windows/Linux always hard-mute), disabled until the toggle is on, 0% renders as "Muted". Default 0 = no behavior change for existing users. New i18n keys added to all 19 locales.
- macOS x86_64: the Swift bridge compiles for `CARGO_CFG_TARGET_ARCH` (not the host arch — the issue that bit #1028 earlier). I don't have Intel hardware; would appreciate a CI test-build confirmation.
- `cargo fmt --check`, `cargo clippy`, `cargo test`, `bun run lint`, `bun run format:check`, `bun run check:translations` all pass. Windows/Linux code paths unchanged.

Behavior notes:
- The system-audio-recording permission prompt is surfaced at startup (when the setting is on), never mid-dictation. If declined, the AppleScript strategies still work like before.
- `bindings.ts` regenerated via the debug build's specta export.

Can add a short screen recording on request — music playing, hold shortcut, music dips to the slider level, release, music returns.

## AI Assistance

**If AI was used:**

- How extensively: Wrote most of the code under my direction; I defined the behavior, tested everything on real hardware (USB interface + standard outputs), and reviewed the changes.


cjpais:
Similar to #1501 Have you looked at or reviewed other PR's which do the same thing? It's much better to contribute or test than just submit PR's which may already be implemented.

newideas99:
You're right, and I should have engaged with the existing PRs before opening this. I spent today actually reviewing and testing them, here's where each stands relative to this PR:

**#547 (closed):** volume-based ducking, cross platform. You closed it because you didn't want to risk regressing the mute path across three OSes without verification. This PR is scoped to avoid exactly that: macOS only, Windows/Linux untouched, and the default (0%) keeps full mute so existing setups don't change behavior.

**#1414 (open):** I reviewed the code. get_output_volume/set_output_volume are only implemented for Windows (they return None / no-op on everything else), and it removes the mute_while_recording setting entirely. So as written, macOS and Linux users lose mute and the new volume reduction does nothing for them. The fade idea is nice though and would compose well with either PR.

**#1028 (open):** built and tested it today on my Traktor (one of the #998 interfaces), full results posted in that thread. Pause works there (0.6s to pause, 0.3s to resume) and it's a good solution for resumable media. Ducking covers the cases pause can't (live streams, calls, and people who just want the music to keep going), so I see them as complementary rather than competing.

**#1417 (open):** I saw you want to move the AppleScript stuff to proper CoreAudio plumbing. That's actually what the tap tier in this PR is (pure CoreAudio process tap + aggregate device, in Swift), so it fits that direction and could ride on top of #1417's mute replacement.

The reason none of the volume-based approaches (including #547, #1414, or even #1417's CoreAudio mute) can fully fix #998: those USB interfaces literally don't expose volume or mute controls to macOS. I probed my Traktor Kontrol S4 MK2 directly with CoreAudio: no VirtualMainVolume, no per-channel VolumeScalar, no mute property. Hardware knob only. A process tap is the only software path that works on that class of device, which is the unique thing this PR adds. At 0% it even gives those devices a mute that actually works.

Totally understand if you'd rather see this reduced in scope (for example just the tap tier as a fix for #998, without the slider). Happy to restructure however is most useful, or to keep testing the other PRs in the meantime.

cjpais:
Thanks for looking things over, I'm mostly asking to reduce spam. I only have so much time and solving the same problem the x time is not that helpful unless there is a clear rationale

Issue #1446, opened 2026-05-28
[BUG] Intermittent freeze during post-recording transcription and missing segments in longer speech inputs


aaliyandev:
# Bug Report: Intermittent Transcription Freeze and Missing Segments

## Bug Description

This issue involves intermittent failures during post-recording transcription processing and occasional missing segments in longer continuous speech inputs.

The problem manifests in two related behaviors:

---

## 1. Post-recording transcription freeze during processing

The application records audio normally without interruption. However, after stopping the recording, during the transcription / merging / post-processing phase, the application sometimes freezes or stops unexpectedly.

### Observed Behavior

- Recording phase works normally
- Issue occurs only after stopping recording
- During transcription/post-processing, the application may freeze or stop
- No error message or crash is shown
- Application becomes unresponsive or stops processing the recorded audio
- Issue occurs intermittently (~5–6 times in ~25–30 uses)

---

## 2. Missing or skipped segments in longer speech input

During longer continuous speech, parts of the transcription output are sometimes missing.

### Observed Behavior

- Earlier segments of speech are occasionally not present in the final output
- Output may contain only later portions of the spoken input
- Appears as if some audio chunks are dropped or not merged correctly
- Occurs intermittently (~3-4 times observed)
- More likely during longer continuous utterances or combined sentences

---

## Reproduction Conditions

The issue is non-deterministic, but occurs more frequently under the following conditions:

- Continuous speech without long pauses
- Longer recordings (~20-60+ seconds)
- Multiple sentences spoken in one uninterrupted flow
- Higher system load during transcription
- Combined or extended speech increases likelihood of missing segments

However:

- Some long recordings work correctly
- Some short recordings occasionally fail
- Behavior is inconsistent and timing-dependent

---

## Observed Pattern

The issue is not strictly length-based. It appears probabilistic and dependent on runtime/system state.

Most failures occur during post-recording processing phase, suggesting a potential issue in:

- streaming buffer handling
- chunk merging logic
- async inference completion handling
- post-recording state transition

---

## Model Observation (Testing Note)

During testing, different transcription models were compared:

- **Moonshine Base** showed better English recognition accuracy and voice consistency, correctly understanding speech more reliably across most inputs. However, it occasionally failed during transcription processing, resulting in interruptions or shutdowns (~3-4 times out of ~8-9 tries)
[...]

PR #1515, opened 2026-06-16
feat: eager segmented transcription (transcribe while recording)

prmsregmi:
## Before Submitting This PR

## Human Written Description

Right now, with any model, Handy records the full clip and only sends it to the model once you stop. So transcription waits until the very end, and depending on how long you spoke it takes ~1–5 seconds before anything shows up. And when you speak for a really long time (say over 2 minutes), it sometimes drops part of the text.

The fix is simple: instead of waiting until the end, transcribe at natural pauses *while* you're still talking, then paste the whole thing at once when you stop. I deliberately did **not** paste in fractions/partials, because:

1. It feels less seamless and more jittery.
2. It makes post-processing harder, since the LLM cleanup wants the full text at once.

It's opt-in (off by default) and I tested it across a few different models. In my testing the wait drops drastically — to under ~1 second — and long recordings stopped dropping text.

## Related Issues/Discussions

Discussion: [#272](https://github.com/cjpais/Handy/discussions/272) (Type-as-you-talk option?), [#179](https://github.com/cjpais/Handy/discussions/179) (Chunking option for almost live processing)
Related: #1446 (less work at stop-time should reduce dropped text on long recordings)

## Community Feedback

This implements functionality the community has explicitly asked for:

- **[#272](https://github.com/cjpais/Handy/discussions/272) — "Type-as-you-talk option?"** (8 upvotes) — maintainer noted intent to support this eventually.
- **[#179](https://github.com/cjpais/Handy/discussions/179) — "Chunking option for almost live processing"** (6 upvotes) — same "process at pauses while speaking" mechanism.

On the quality trade-off raised in [#179](https://github.com/cjpais/Handy/discussions/179): this is exactly why the feature is **opt-in (an experimental toggle, default off)** — default behaviour is unchanged, and users who want lower latency can turn it on.

Note on direction: in [#272](https://github.com/cjpais/Handy/discussions/272)/[#179](https://github.com/cjpais/Handy/discussions/179) the preference was to land streaming/chunking in the upstream `transcribe-rs` library first. This PR instead works at Handy's pipeline layer by re-using the existing per-buffer transcribe call on VAD-delimited segments, so it is **model-agnostic** and works with every model available today (Whisper, Parakeet, Moonshine, etc.) without waiting on upstream streaming support. If/when `transcribe-rs` gains native streaming, this can be revisited.

## Testing

- **Unit tests:** `cargo test` passes (72 tests), including 7 new tests covering the VAD-segment boundary logic — segment-per-pause, trailing-segment flush on stop, sub-minimum-length merge, max-length safety cut, and the invariant that the union of emitted segments equals the full recording buffer (so the saved WAV/history audio is unchanged).
- **Lint/format:** `cargo clippy` and `cargo fmt --check` clean for the new code; frontend `tsc`, ESLint, and Prettier all clean.
- **Manual (macOS, Apple Silicon, release build):**
  - Verified across multiple models (Parakeet, Whisper) — speaking with natural pauses; post-stop wait dropped to roughly under a second vs. the previous whole-buffer wait.
  - Toggle **off** (default): behaviour is identical to current `main` (no worker spawned, no segmentation) — regression check.
  - Confirmed the full audio is still saved to history/WAV, post-processing (Chinese/LLM cleanup) still runs once on the joined transcript, and cancelling mid-recording discards work cleanly.

https://github.com/user-attachments/assets/8a3652db-d704-4676-9d43-72bd5f3e5c01

## AI Assistance

**If AI was used:**

- Tools used: Claude Code (Claude Opus)
- How extensively: AI implemented the backend audio-pipeline changes (VAD-boundary segmentation in the recorder, a per-segment transcription worker), the settings plumbing, and the unit tests. I directed the design and the decisions (opt-in toggle, no partial pasting, single post-processing pass), did the manual testing across models, and wrote this description myself. The change was also run through automated code review before opening.


cjpais:
Hey, I do want to say I appreciate you submitting this PR. The latest version does include native streaming support with streaming models, and I think its general overall better user experience. At least from my experience with chunking, there was a lot of awkwardness in the transcriptions. I haven't tried your particular version of it, but I was unable to get a version that I was satisfied shipping. So for now, I'm closing this partially because it needs to be rebased and partially because I would love evaluation on the streaming models that we now ship and see if this is even really necessary in the application. I would like to avoid adding new features unless fundamentally there is a reason to do so.

Issue #1509, opened 2026-06-14
Cannot cancel an in-progress model download during onboarding


adisagar2003:
### Summary
During onboarding, when downloading a model there's no way to cancel an in-progress download. Once a download starts, it's "stuck" and I can't abort it to download a different model instead.

### Steps to reproduce
1. Go through onboarding and reach the model selection / download step
2. Start downloading a model (e.g. Parakeet v3)
3. Try to cancel that download and switch to a different model (e.g. Moonshine v2 Tiny)

### Expected
A way to cancel/abort the current download, then start a different one.

### Actual
No cancel option — the download is stuck and I can't switch models.

### Suggestion
Add a cancel button to in-progress downloads so users can abort and pick another model. Minor UX nitpick, but it leaves you stuck mid-onboarding.

PR #1522, opened 2026-06-18
fix: surface download errors during onboarding

kirilan:
## Before Submitting This PR

## Human Written Description

When initializing the application I noticed that the model download was stuck because my adguard was blocking requests to blob.handy.computer. There was no error so troubleshooting was not easy.

Once I found out what the issue was the fix was fairly trivial. There is no logic change the fix uses a single toaster element above the branching and renders it in every visible branch (accessibility, model, and main). Safe because sonner renders through a portal, so tree position doesn't affect layout. 

The "fix" only helps provide feedback to the user to allow for quick error identification. 

## Related Issues/Discussions

Related but not a duplicate: [PR #997](https://github.com/cjpais/Handy/pull/997) covers model loading notifications, not onboarding download failures.

## Testing

**Reproduction (before the fix):**

1. Make `blob.handy.computer` unresolvable — e.g. add `0.0.0.0 blob.handy.computer` to
   the hosts file, or use a DNS server/sinkhole that blocks it.
2. Start with no models installed so the first-run onboarding wizard reaches the
   model-selection step.
3. Click a model to download it.

- **Before:** the model card spinner appears for a moment, then clears with no message —
  the wizard appears to just "blink" and stall, giving the user no indication of what went wrong.
- **After:** an error toast is shown (surfaced from the existing `model-download-failed`
  event), so the user knows the download failed.

**Root cause:** `<Toaster />` (sonner) was only rendered in the main app view. The
onboarding steps return early before it, so any `toast.error(...)` call during onboarding
had no Toaster mounted and rendered nowhere. This affected the model-download failure path
as well as the other `toast.error` calls in the onboarding/accessibility screens. The fix
hoists a single `<Toaster />` above the onboarding branches in `src/App.tsx`.

**Static checks (all passing):** the following were run against the change on the frontend
toolchain:

- `tsc --noEmit` — passes
- `eslint src/App.tsx` — passes
- `prettier --check` — passes (Prettier code style)
- `vite build` (`tsc && vite build`) — production build succeeds

The change is frontend-only TSX (it reuses the already-imported `Toaster`), so the Rust
backend was not affected or rebuilt.

**Still help-wanted — live runtime check:** I have not been able to do an end-to-end run of
the reproduction on a running build (blocking `blob.handy.computer` and confirming the toast
actually appears in the onboarding wizard). I'd appreciate a maintainer running through the
steps above on a real `bun run tauri dev`/build to confirm the toast shows as expected.

## AI Assistance

**If AI was used:**

- Tools used: Claude Opus 4.8 High
- How extensively: Detected and wrote the fix, human validated and tested


cjpais:
Thanks for this I will try it and merge, this is needed.

koloved:
As far as I understand, this should solve the problem, right? https://github.com/cjpais/Handy/issues/1509

kirilan:
> As far as I understand, this should solve the problem, right? #1509

not really, as it does not add a cancel button it just shows the error

PR #1537, opened 2026-06-23
fix: skip post-processing when transcription is empty

### LINKED ISSUE #1409: [BUG] post processing should not be activated on empty transcription
## Bug Description

Sometimes I start a transcription holding down the hotkeys but then realize I'm not ready to speak I'll release the keys even though I haven't said anything. In this situation the post-processing model is still activated, it ends up typing an error response like "you need to provide the transcription". This feels like unexpected behavior, handy should probably not enter post-processing phase if the output from basic transcription is either blank (or whitespace?). 

## System Information
v0.8.3

AhmedDlshad007:
## Before Submitting This PR

## Human Written Description

I was looking through Handy's open bug reports for something small to pick up, and this one stood out. It's a bit of a papercut. If you tap the hotkey but don't actually end up saying anything, instead of just doing nothing it still fires off the post-processing model, which then types out a confusing error like "you need to provide the transcription." That felt wrong for a no-op, so the fix just skips post-processing when there's nothing to process. Seemed like a clean, contained change to start with.

## Related Issues/Discussions

Fixes #1409

## Community Feedback

The bug is confirmed by the maintainer in #1409: "Yeah, we should not post-process if there's an empty transcription."

## Testing

Added unit tests for the new `is_blank_transcription` helper covering empty and whitespace-only input. The fix is a minimal early-return guard at the top of `post_process_transcription()`, which is the single function all post-processing paths route through, mirroring its existing "no provider" and "empty prompt" early-returns. I did not run the full app locally, so CI covers fmt/clippy/tests.

## AI Assistance

**If AI was used:**

- Tools used: Claude Code
- How extensively: Claude Code found the issue, wrote the guard plus unit tests, and drafted this PR. I reviewed the change and wrote the description above.

### DIFF
--- a/src-tauri/src/actions.rs
+++ b/src-tauri/src/actions.rs
@@ -63,7 +63,21 @@ fn build_system_prompt(prompt_template: &str) -> String {
     prompt_template.replace("${output}", "").trim().to_string()
 }
 
+/// Returns `true` when a transcription has no meaningful content to
+/// post-process (empty or whitespace-only). Used to skip the post-processing
+/// LLM call when nothing was actually transcribed, which would otherwise make
+/// the model reply with an error message such as "you need to provide the
+/// transcription".
+fn is_blank_transcription(transcription: &str) -> bool {
+    transcription.trim().is_empty()
+}
+
 async fn post_process_transcription(settings: &AppSettings, transcription: &str) -> Option<String> {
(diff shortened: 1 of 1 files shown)


cjpais:
Thanks for the fix, simple enough and works

PR #1548, opened 2026-06-28
[codex] Fix compressed LLM API responses

ArtemiyShisha:
## Summary
- Enable reqwest decompression support for LLM API responses.
- Add a regression test that serves an OpenAI-compatible chat completion response with `Content-Encoding: gzip`.

## Why
Some OpenAI-compatible APIs can return compressed JSON responses. Without reqwest compression support, Handy fails while decoding the response body and falls back to the raw transcription instead of using the post-processed output.

## Validation
- `cargo fmt -- --check`
- `cargo test llm_client --no-default-features`

### DIFF
--- a/src-tauri/Cargo.lock
+++ b/src-tauri/Cargo.lock
@@ -232,6 +232,18 @@ dependencies = [
 ]
 
 [[package]]
+name = "async-compression"
+version = "0.4.43"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "3976abdc8fe7d1133d43d304afd42abdf5bc3e1319d263d223bde07b5efc4be8"
+dependencies = [
+ "compression-codecs",
+ "compression-core",
+ "pin-project-lite",
+ "tokio",
+]
+
+[[package]]
 name = "async-executor"
 version = "1.14.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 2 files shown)


cjpais:
Can you give a model provider which does this so I can test before and after. Like, you definitely need to specify how this can be tested specifically to make it easier for maintainers to test your work

ArtemiyShisha:
> Can you give a model provider which does this so I can test before and after. Like, you definitely need to specify how this can be tested specifically to make it easier for maintainers to test your work

I reproduced this with Handy's built-in OpenAI provider against `https://api.openai.com/v1/chat/completions`.

Manual before/after test:

1. Configure Handy post-processing with the OpenAI provider, an OpenAI API key, and any Chat Completions-compatible model available for that key.
2. Enable post-processing.
3. Dictate text where the post-processing prompt should visibly change the output.
4. On the failing build, Handy can receive a `Content-Encoding: gzip` response and logs `Failed to parse API response: error decoding response body`, then falls back to the raw transcription.
5. On this branch, the same flow decodes the compressed response and post-processing succeeds.

For a deterministic check without depending on a live provider response, this PR also adds a regression test that serves an OpenAI-compatible gzip-encoded response:

```sh
cargo test llm_client::tests::send_chat_completion_decodes_gzip_response --no-default-features

cjpais:
ive reviewed this and it's fine. ive not been able to replicate the failure against openai's api, but its fine for hardening and reducing bandwidth even if small

PR #1553, opened 2026-06-29
perf: use tmpfs (/tmp) for recordings when available to reduce disk I/O

D3nis:
## Before Submitting This PR

- [x] I have searched existing issues and pull requests

## Human Written Description

On systems with slow HDDs or flash storage (eMMC, SD cards), writing temporary WAV recordings to the app data directory causes unnecessary disk wear and latency. Since these files are short-lived (deleted after transcription), tmpfs is a much better location. This change checks if `/tmp` is mounted as tmpfs via `/proc/mounts` and uses `/tmp/handy-recordings/` if available, falling back to the original app data directory otherwise.

## Testing

- Fedora 43, `/tmp` mounted as tmpfs (16GB)
- WAV files confirmed written to `/tmp/handy-recordings/`
- Fallback to app data dir works when tmpfs unavailable

## AI Assistance

- [x] AI was used
- Tools used: Claude (Anthropic)
- How extensively: helped write the Rust code. Testing and decision made by me.

### DIFF
--- a/src-tauri/src/managers/history.rs
+++ b/src-tauri/src/managers/history.rs
@@ -73,9 +73,26 @@ pub struct HistoryManager {
 
 impl HistoryManager {
     pub fn new(app_handle: &AppHandle) -> Result<Self> {
-        // Create recordings directory in app data dir
+        // Create recordings directory — prefer tmpfs (/tmp) if available to avoid
+        // unnecessary writes to slow HDD or flash storage wear.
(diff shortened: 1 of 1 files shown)


koloved:
It would be great to do it as platform-based then and write to the RAM. While that amount of disk entries is unlikely to scare anyone.

D3nis:
this is closed. missing some checks for write permissions. please see https://github.com/cjpais/Handy/pull/1554

PR #1554, opened 2026-06-29
perf: use tmpfs (/tmp) for recordings when available to reduce disk I/O

D3nis:
## Before Submitting This PR

- [x] I have searched existing issues and pull requests

## Human Written Description

On systems with slow HDDs or flash storage (eMMC, SD cards), writing temporary WAV recordings to the app data directory causes unnecessary disk wear and latency. Since these files are short-lived (deleted after transcription), tmpfs is a much better location. This change checks if `/tmp` is mounted as tmpfs via `/proc/mounts` and uses `/tmp/handy-recordings/` if available, falling back to the original app data directory otherwise.

## Testing

- Fedora 43, `/tmp` mounted as tmpfs (16GB)
- WAV files confirmed written to `/tmp/handy-recordings/`
- Fallback to app data dir works when tmpfs unavailable

## AI Assistance

- [x] AI was used
- Tools used: Claude (Anthropic)
- How extensively: helped write the Rust code. Testing and decision made by me.

### DIFF
--- a/src-tauri/src/managers/history.rs
+++ b/src-tauri/src/managers/history.rs
@@ -73,9 +73,32 @@ pub struct HistoryManager {
 
 impl HistoryManager {
     pub fn new(app_handle: &AppHandle) -> Result<Self> {
-        // Create recordings directory in app data dir
+        // Create recordings directory — prefer tmpfs (/tmp) if available to avoid
+        // unnecessary writes to slow HDD or flash storage wear.
         let app_data_dir = crate::portable::app_data_dir(app_handle)?;
-        let recordings_dir = app_data_dir.join("recordings");
+        let recordings_dir = {
+            let tmp_dir = std::path::Path::new("/tmp/handy-recordings");
+            // Check if /tmp is tmpfs by reading /proc/mounts
+            let is_tmpfs = std::fs::read_to_string("/proc/mounts")
+                .map(|mounts| mounts.lines().any(|line| {
+                    let parts: Vec<&str> = line.split_whitespace().collect();
+                    parts.len() >= 3 && parts[1] == "/tmp" && parts[2] == "tmpfs"
+                }))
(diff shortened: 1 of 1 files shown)


koloved:
I'm certainly not against new features in the app, but I would prefer to first have a stable application with all the current functions. Without implementing new ones. We need to ensure that the Linux application works perfectly in its current state across distros/environments. The more new features are implemented, the harder it will be to fix old problems that should have been corrected at the beginning.

cjpais:
Yeah, I'm pretty sure this is completely unnecessary to be honest. You can just turn off recordings if you don't want them. The recordings are actually not necessary at all for the application itself to work, and nothing is written to disk to actually do the transcription anyhow. So you should just turn recordings off if you don't want them.

It's possible that we write files, even they might just get deleted immediately. I haven't checked this code path in a while, so it's possible we are writing unnecessarily, but that's a different problem and should be resolved in a different PR. And not writing them in the first place would be generally a much better solution than a platform-specific one.

D3nis:
@cjpais  I have to admit it wasn't immediately obvious to me that setting `history_limit: 0` means the app doesn't use disk as an intermediate storage layer before sending audio to the neural network for transcription. I had to look at the code to understand this.
So you're right — the proper fix is to skip WAV writes entirely when `history_limit == 0` or recordings are disabled. That would be a cleaner solution than tmpfs.

cjpais:
@D3nis I agree, it's one of the worst bits of UX in the app I think. There probably should be a clear toggle enable/disable history. The default of 5 is also probably too low. And auto delete recordings is awkward with both time mechanisms and a history limit. The whole thing is a mess. I'm not really sure what the best thing to do with it is.

Anyhow, if you submit a fix to skip .wav writes when limit is 0, I certainly will accept it. This is a bug and unintended behavior and certainly an oversight

Issue #1483, opened 2026-06-08
[BUG] Parakeet ORT failure on long recording drops all artifacts


mullen3:
## Before You Submit

Searched existing issues for `Parakeet ORT error`, `long recording`, `no wav`, and `history`. I did not find an obvious duplicate for this specific failure mode.

## Bug Description

A long Handy recording failed during Parakeet transcription with a generic ORT error. Because the failure happened before Handy persisted the recording, the app dropped all user-facing artifacts: no `.wav`, no history row, no transcript result in the log, and no clipboard contents.

Handy recorded from roughly `2026-06-05 13:32:04` to `13:41:37` local time, about 9.5 minutes. It successfully captured `8,566,560` samples, then failed during transcription.

The larger bug is that audio persistence appears coupled to transcription success. Even if transcription fails, Handy should preserve the captured audio so the user can retry or recover it.

## System Information

**App Version:**

0.8.3

**Operating System:**

macOS 26.4, build 25E246

**CPU:**

Apple M1 Max, 10 cores: 8 performance, 2 efficiency

**GPU:**

Apple M1 Max GPU, 32 cores

## Logs

Relevant log excerpt from `~/Library/Logs/com.pais.handy/handy.log`.

Note: Handy’s log timestamp appears one hour behind local time. The log window `12:32:04` to `12:41:37` corresponds to local `13:32:04` to `13:41:37`.

```text
[2026-06-05][12:32:04][handy_app_lib::actions][DEBUG] TranscribeAction::start called for binding: transcribe
[2026-06-05][12:32:04][handy_app_lib::actions][DEBUG] Microphone mode - always_on: false
[2026-06-05][12:32:04][handy_app_lib::actions][DEBUG] On-demand mode: Starting recording first, then audio feedback
[2026-06-05][12:32:04][handy_app_lib::managers::audio][INFO] Microphone stream initialized in 125.497375ms
[2026-06-05][12:32:04][handy_app_lib::managers::audio][DEBUG] Recording started for binding transcribe
[2026-06-05][12:32:04][handy_app_lib::audio_toolkit::audio::recorder][INFO] Using device: Ok("Connor’s AirPods Pro #3 - Find My")
Sample rate: 24000
Channels: 1
Format: F32
[2026-06-05][12:41:37][handy_app_lib::actions][DEBUG] TranscribeAction::stop called for binding: transcribe
[...]

Issue #1332, opened 2026-04-25
[BUG] Handy silently drops long audio recordings (~5min)


VjiaoBlack:
## Before You Submit

**Please search [existing issues](https://github.com/cjpais/Handy/issues) to avoid duplicates.** Your bug may already be reported! Right now it's just me maintaining this project so many issues can be overwhelming! Help me out by checking first.

## Bug Description

Handy silently drops long audio recordings.

Push to talk off. Short recordings transcript. Large recordings are silently lost - recording is not saved, text is not outputted. Very frustrating especially for long recordings where Handy seems to be doing fine, but then ends up deleting or losing the whole speech. 

## System Information

**App Version:**

v0.8.2

**Operating System:**

Tahoe 26.4.1

**CPU:**

Apple M1 Pro

**GPU:**

Apple M1 Pro?

## Logs

Work device; cannot attach logs.

cjpais:
No audio is showing up in history? Can you give a log? You can censor confidential info.

aspectrr:
I'm seeing an issue where I did a recording for 8 minutes and then it fails to transcribe with Parakeet V3. Supposedly Parakeet V3 should allow for up to 24 minute long translation with full attention. Also an M1 Macbook Pro with Tahoe macOS. The recording is saved but then it is not transcribed.

Here is the error from the logs:
`[2026-05-11][15:03:08][handy_app_lib::actions][DEBUG] Global Shortcut Transcription error: Parakeet transcription failed: inference error: Non-zero status code returned while running Add node. Name:'/layers.0/self_attn/Add_2' Status Message: /Users/runner/work/ort-artifacts/ort-artifacts/onnxruntime/onnxruntime/core/providers/cpu/math/element_wise_ops.h:540 void onnxruntime::BroadcastIterator::Init(ptrdiff_t, ptrdiff_t) axis == 1 || axis == largest was false. Attempting to broadcast an axis by a dimension other than 1. 1192 by 6192`

cjpais:
The onnx implementation does not allow for this, a future ggml version may

chipsenkbeil:
I have an 8 minute audio recording that won't transcribe, but is in my history, on a Mac M1 (15.7.3) using Handy 0.8.3. Tried a couple of Parakeet models and just shows that the transcribe failed. Retrying doesn't help, but I at least have the .wav file.

[...]

PR #1529, opened 2026-06-20
introduce transcribe.cpp

cjpais:
bring in transcribe.cpp as the primary transcription engine in handy

adds streaming transcription and a bunch of new models

PR #1563, opened 2026-06-30
fix(transcription): chunk long GigaAM audio to prevent crash on recordings over ~3 min

thriw:
## Before Submitting This PR

- [x] I searched existing issues and pull requests to make sure this isn't a duplicate
- [x] I read CONTRIBUTING.md

## Human Written Description

I use the GigaAM v3 model and noticed that any recording longer than a few minutes just gave me no text at all, while short ones worked fine. Losing several minutes of speech with nothing to show for it is the worst case, and long dictations are exactly when I need it most, so I looked into why it happens and fixed it.

## Related Issues/Discussions

Fixes the GigaAM case of #1332 (Handy silently drops long audio recordings around 5 min). The same cause probably affects other ONNX models like Parakeet (#1483), but I only changed and tested the GigaAM path. There is also an unmerged `chunking` branch in the repo doing a broader streaming version; this is a smaller targeted fix that reuses the same transcribe-rs VadChunked.

## Community Feedback

#1332 is open with 10 comments from people hitting the same lost long recordings.

## What was wrong

GigaAM crashes on recordings longer than about 3 minutes. The ONNX model has a fixed limit of 5000 encoder frames, which is roughly 200 seconds of audio. Past that it errors out in the attention layer ("5000 by N"). Quality also drops well before that. GigaAM's own README says the model is only meant for audio up to 25 seconds, and for longer audio you are supposed to split it with a VAD first (their transcribe_longform does exactly this).

## The fix

If the audio is longer than 40 seconds, split it into chunks of up to 24 seconds at silent points using the Silero VAD that already ships with Handy, transcribe each chunk, and join the text. Audio under 40 seconds goes through the same code as before, so short recordings are untouched.

Changes:
transcription.rs, route long GigaAM audio through VadChunked.
Cargo.toml, enable the vad-silero feature (Cargo.lock is unchanged).

## Testing

Built it on a Mac (M4, GigaAM v3 int8) and recorded 3.6 minutes. Before the fix this length crashed. After the fix it split into 17 chunks, transcribed the whole thing, and pasted the text. No errors.

Then I ran a benchmark comparing the old behavior (whole audio in one pass) against the fix (chunked), both against a known transcript, measuring word error rate (lower is better). Test files were built from the open GOLOS Russian dataset at fixed lengths, three files each.

Results, old vs new, WER %:
30 sec: old 25.0 / 18.2 / 21.7, new 25.0 / 18.2 / 39.1
120 sec: old 89 / 93 / 91, new 25 / 21 / 23
180 sec: old 100 / 98 / 100, new 19 / 26 / 29
600 sec: old crashes, new 26 / 33 / 27

Clips that fit a single chunk come out identical (25.0 vs 25.0, 18.2 vs 18.2), so there is no quality loss on audio the model can already handle. Past 2 minutes the old path returns near garbage before it crashes after about 200 seconds. The one 30 sec file that differs is actually 34 seconds, so it gets split.

How to reproduce: the dataset builder and the benchmark tool are here: https://gist.github.com/thriw/24dc2059790532b7fddd7421d89fc58a . No audio is committed, the script regenerates it from GOLOS.

### DIFF
--- a/src-tauri/Cargo.toml
+++ b/src-tauri/Cargo.toml
@@ -69,7 +69,7 @@ rusqlite = { version = "0.37", features = ["bundled"] }
 tar = "0.4.44"
 flate2 = "1.0"
 sha2 = "0.10"
-transcribe-rs = { version = "0.3.8", features = ["whisper-cpp", "onnx"] }
+transcribe-rs = { version = "0.3.8", features = ["whisper-cpp", "onnx", "vad-silero"] }
 handy-keys = "0.2.4"
 ferrous-opencc = "0.2.3"
 clap = { version = "4", features = ["derive"] }
--- a/src-tauri/src/managers/transcription.rs
+++ b/src-tauri/src/managers/transcription.rs
@@ -502,6 +502,19 @@ impl TranscriptionManager {
             }
         };
 
+        // Resolved once here so it can be moved into the panic-guarded closure.
+        // Used by GigaAM to VAD-chunk long audio that would otherwise exceed the
+        // model's encoder limit.
+        let vad_path = self
+            .app_handle
+            .path()
+            .resolve(
+                "resources/models/silero_vad_v4.onnx",
+                tauri::path::BaseDirectory::Resource,
+            )
+            .ok();
+        let vad_path = vad_path.as_ref().and_then(|p| p.to_str());
+
         // Perform transcription with the appropriate engine.
         // We use catch_unwind to prevent engine panics from poisoning the mutex,
         // which would make the app hang indefinitely on subsequent operations.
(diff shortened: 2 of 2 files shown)


cjpais:
This is going to be deprecated by #1529 tomorrow,

thriw:
Thanks for the quick look, and that makes sense given the transcribe.cpp work. Since #1529 is still WIP and ONNX stays cross-compatible for now, I mainly meant this as a small stopgap so current GigaAM users stop losing long recordings (#1332) until the new runtime ships. Totally fine to close it if it is not worth carrying, your call. Out of curiosity, will the new runtime need long-audio VAD chunking too, or does it handle arbitrary lengths natively?

cjpais:
Right now I'm not accepting stop gaps for something which will be deprecated. Please move the effort over to the new engine. Likely it will still require chunking, but please test it

PR #1566, opened 2026-07-01
test build to dynamically link onnxruntime and also remove DirectML support

cjpais:


koloved:
I tested this build. And what's interesting, upon initial startup...  had the Overlay window. Sometimes it happened to me that I would start speaking, and the microphone sensitivity graph would show that something was recording. Although in the end, part of the initial text was skipped.

cjpais:
@koloved probably more an issue with the #1529, this was just a test build to try and fix the FMA crash

I am releasing 0.9.0 and it will have some bugs right now, I really just want to get the release out and discover the bugs in prod. This is a major release and has some risk before v1 comes, I want to really spend the time to stabilize things and clean up the last UX issues

Issue #1555, opened 2026-06-29
[META] Wayland: text insertion, hotkeys, and Flathub publication


koloved:
## 🎯 Goal

Consolidate all open issues and PRs related to Wayland and text insertion via hotkeys into a single tracker. Simplify the maintainer's (`@cjpais`) process for merging solutions and tracking problems caused by major differences between Linux distributions.

**Core problem:** Wayland has no single API for programmatic text insertion or registering global hotkeys. Each compositor (GNOME/Mutter, KDE/KWin, wlroots-based compositors) implements different protocols. This leads to fragmented solutions and duplicated effort.

---

### Rationale

| Problem | Solution via Flatpak |
|---|---|
| **Atomic systems** (Bazzite, Fedora Atomic, SteamOS) — cannot install ydotool/wtype/dotool via package manager | Flatpak includes all dependencies inside the sandbox |
| **Different library versions** across distributions (Ubuntu 22.04 vs Arch vs Fedora 43) | Flatpak uses a single runtime (GNOME Platform) |
| **GlobalShortcuts portal** requires a correct app_id | Flatpak guarantees a stable `com.pais.handy` |
| **Different package managers** (apt, dnf, pacman, rpm-ostree) | One format for all |
| **Auto-updates** | Flathub + x-checker-data automatically updates versions |

### Current status

- PR #548 (@GabeDuarteM) — Flatpak packaging support added (open, awaiting merge)
- PR #689 (@petit-aigle, @D3nis) — Remote Desktop portal for Wayland (required for Flatpak)
- PR #1287 (@nmcbride) — GlobalShortcuts portal backend (required for hotkeys in Flatpak)
~- PR #1395 (@vetmiker) — Wayland paste & post-process fix~ (seems not works and not needed in Flatpak)
- testing Flatpak builds on Bazzite (KDE Wayland), finding issues (set_resizable workaround for title bar buttons, pure Wayland support)

---

### Wayland + text insertion

| Issue | Description | Environment |
|---|---|---|
| [#1549](https://github.com/cjpais/Handy/issues/1549) | GNOME Wayland: keycombo paste picks wtype then fails with no fallback; ExternalScript paste hangs via wl-copy | GNOME Wayland |
| [#1282](https://github.com/cjpais/Handy/issues/1282) | Post-processing never triggers on Linux KDE Wayland (v0.8.2) | KDE Wayland |
[...]

Issue #1549, opened 2026-06-28
[BUG] GNOME Wayland: keycombo paste selects wtype then fails with no fallback; ExternalScript paste hangs via wl-copy


JayDoubleu:
## Summary

On GNOME Wayland, two separate problems make paste-back unreliable. Both have
straightforward fixes. Reporting together because they surface in the same flow
(getting paste working on GNOME) and the second is the workaround for the first.

## Environment

- Fedora Workstation 44, GNOME Shell 50.2, Wayland session
- Handy 0.8.3 (AppImage)
- Paste tools available: `wl-clipboard`, `ydotool`; `wtype` installed initially

## Problem 1: keycombo paste selects `wtype` on GNOME, then fails with no fallback

On a clipboard keycombo paste method (e.g. Shift+Insert / Ctrl+V), the Wayland
keycombo path tries `wtype -> dotool -> ydotool` and picks the first available.
With `wtype` installed it logs:

```
[clipboard][INFO]  Using paste method: ShiftInsert, delay: 60ms
[clipboard][INFO]  Using wl-copy for clipboard write on Wayland
[clipboard][INFO]  Using wtype for key combo
[actions][ERROR] Failed to paste transcription: wtype failed: Compositor does not support the virtual keyboard protocol
```

GNOME/Mutter does not implement `zwp_virtual_keyboard_v1`, so `wtype` can never
send the keystroke there. Because selection is by availability and there is no
runtime fallback, an installed-but-nonfunctional `wtype` makes paste fail even
though `ydotool` is installed and works. (Enigo on this build uses the X11/XTEST
backend, which also does not reach native Wayland clients.)

**Suggested fix:** treat `wtype` as unavailable on GNOME (the same way it is
already excluded on KDE), and/or fall back to the next tool when the chosen one
returns a runtime error rather than aborting the paste.

## Problem 2: ExternalScript paste hangs the app after the first paste

`PasteMethod::ExternalScript` runs the script with `Command::output()`:

```rust
let output = Command::new(script_path).arg(text).output()?;
```

`output()` waits for the child to exit AND for stdout/stderr to reach EOF. The
canonical Wayland clipboard tool, `wl-copy`, forks a background daemon that
serves the selection and inherits the script's stdout/stderr. That daemon keeps
the fd open for the lifetime of the selection, so `output()` never sees EOF and
Handy blocks indefinitely. Symptom: paste works exactly once, then the app
appears frozen (the transcribe action never completes; `--cancel` frees it).

[...]

PR #1276, opened 2026-04-11
fix(linux): Clipboard paste methods (ctrl_v / ctrl_shift_v / shift_insert) fail on GNOME Wayland

mrinterweb:
Related to #429

### What

Adds `is_gnome()` and `is_gnome_wayland()` helpers in `utils.rs` and extends the wtype guards in `try_send_key_combo_linux` and `try_direct_typing_linux` to skip wtype on GNOME the same way it's already skipped on KDE.

### Why

I was also encountering this issue.  Direct input worked with the remote desktop app in gnome, but it was terribly slow. 

Mutter (GNOME's compositor) deliberately doesn't implement `zwp_virtual_keyboard_manager_v1`, the same protocol `wtype` relies on, and the same reason the existing code skips wtype on KDE. On GNOME Wayland, Handy's key-combo dispatch was picking wtype first, failing, and propagating the error instead of falling through to dotool/ydotool, which broke the `ctrl_v`, `ctrl_shift_v`, and `shift_insert` paste methods entirely.

This addresses the clipboard paste portion of #429. The broader Direct mode/enigo issues discussed in that thread (dropped capital letters, input-method-v2 support) remain separate.

### How

Minimal, targeted change that mirrors the existing KDE detection pattern. No new dependencies, no behavior changes for KDE/X11/macOS/Windows.

## Before Submitting This PR

- [x] I have searched [existing issues](https://github.com/cjpais/Handy/issues) and [pull requests](https://github.com/cjpais/Handy/pulls) (including closed ones) to ensure this isn't a duplicate
- [x] I have read [CONTRIBUTING.md](https://github.com/cjpais/Handy/blob/main/CONTRIBUTING.md)

## Related Issues/Discussions

Related to #429 — this addresses the clipboard paste portion (ctrl_v, ctrl_shift_v, shift_insert failing on GNOME Wayland). The broader Direct mode/enigo issues discussed in that thread remain separate.

## Testing

Verified locally on CachyOS + GNOME Wayland (Mutter):
- **Before:** `Using wtype for key combo` → `wtype failed: Compositor does not support the virtual keyboard protocol` — clipboard paste methods fail entirely, falling back to enigo's RemoteDesktop portal path (~40ms/keystroke)
- **After:** wtype is skipped, cascade falls through to ydotool which sends the paste key combo. Text pasted successfully via clipboard with no errors.

Also tested direct typing still works correctly (falls through to enigo's portal path, same as before). No regression observed on KDE/X11 paths (guard logic unchanged for those).

[...]

PR #1568, opened 2026-07-01
fix: support global shortcuts and paste on GNOME/Wayland (Ubuntu 26.04)

EmanoelDionisio:
## Before Submitting This PR

## Human Written Description

I run Handy on a few machines and it's been solid on macOS and Ubuntu 24.04, but on a fresh Ubuntu 26.04 install (GNOME on Wayland) it was basically unusable: the global shortcut never started a recording, and when I did get audio through, the transcribed text was never pasted into the focused app. The logs matched what others reported in #1549 — `wtype failed: Compositor does not support the virtual keyboard protocol` — even with `ydotool` installed. After digging in, both problems come down to Wayland/GNOME specifics, not the transcription pipeline itself. This fix keeps current behavior on X11/macOS/Windows and only changes what's needed on Wayland. It matters to me because 26.04 is what ships on new machines now, and the app looks broken out of the box there.

## Related Issues/Discussions

Addresses #1549 (Problem 1: GNOME Wayland keycombo paste picks `wtype`, fails, and previously had no runtime fallback)

Related to #1555 (Wayland meta tracker — text insertion and hotkeys)

Related to #429 (GNOME Wayland paste reliability; clipboard keycombo portion)

Related to #949 (Tauri global shortcuts don't work on Wayland compositors — defaulting to `HandyKeys`/evdev on Wayland)

Discussion: #1555

**Note:** #1549 Problem 2 (ExternalScript paste hanging via `wl-copy`) is **not** addressed here.

**Overlap with other PRs:** #1276 tackles the same GNOME/`wtype` paste guard; this PR goes further with runtime per-tool fallback (`attempt_tool`), explicit GNOME ordering (`ydotool` → `dotool` → `wtype`), default `HandyKeys` on Wayland for hotkeys, and packaging/docs/CI for Ubuntu 26.04. Happy to coordinate if the maintainer prefers a narrower scope.

## Community Feedback

This is a bug fix, not a feature. #1549 documents the exact failure mode (including logs) on GNOME Wayland with Handy 0.8.3. #1555 tracks Wayland paste/hotkey work across the project. #522 (closed) reported the same `wtype` error being surfaced only in logs with no fallback — the `attempt_tool` helper and GNOME tool ordering address that behavior.

## Testing

Tested on Ubuntu 26.04 (GNOME / Wayland):

- Built the `.deb` locally and installed it clean.
- Global shortcut now triggers recording on first launch. Before this change the Tauri global-shortcut backend simply never fired under Wayland.
- Transcribed text is pasted reliably. Before, `wtype` failed with "Compositor does not support the virtual keyboard protocol" and nothing landed in the field.
- Environment: `ydotool` + `wl-clipboard` installed, `ydotoold` running, user in the `input` group.

**System information:**

- App version: 0.8.3 (local build)
- OS: Ubuntu 26.04, GNOME on Wayland (`XDG_SESSION_TYPE=wayland`)
- CPU: AMD Ryzen 7 5700X
- GPU: AMD Radeon RX 6700 XT (Navi 22)

To confirm I didn't regress the other targets:

- Everything I added on the runtime side is gated behind `#[cfg(target_os = "linux")]`, so macOS and Windows compile and behave exactly as before.
- On Linux/X11 nothing changes: the keyboard backend stays `Tauri` and paste still prefers `xdotool`.
- `cargo fmt --check`, `cargo clippy`, and the lib tests pass with no new warnings.

### What changed and why

**Global shortcut.** On Wayland the compositor doesn't hand out a system-wide key-grab API, so the Tauri global-shortcut plugin can't see the hotkey (#949). The `handy-keys` backend already reads input via evdev (`/dev/input`), which works globally on Wayland as long as the user is in the `input` group. The default `KeyboardImplementation` now resolves to `HandyKeys` on Linux **Wayland** sessions and stays `Tauri` on X11.

**Paste / typing.** `wtype` depends on the virtual keyboard protocol that GNOME doesn't implement (#1549, #522). On GNOME Wayland the fallback chain now prefers uinput-based tools (`ydotool`, then `dotool`) before `wtype`, and `attempt_tool` logs failures and falls through instead of aborting on the first error. If the clipboard paste combo still fails on Wayland, it falls back to direct typing.

**Packaging & docs.** The `.deb` recommends `ydotool` + `wl-clipboard`, the CI gets a dependency step for Ubuntu 26.04 builds, and README/BUILD document 26.04 support plus the extra build deps (`clang`, `libclang-dev`, `libevdev-dev`).

## AI Assistance

- Tools used: Cursor (Claude).
- How extensively: used it to diagnose the Wayland/GNOME behavior, search related issues/PRs, and draft the Linux fallback logic and docs. I reviewed every change and tested the build on Ubuntu 26.04 myself.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -135,6 +135,12 @@ jobs:
           sudo apt-get update
           sudo apt-get install -y libwebkit2gtk-4.1-dev libappindicator3-dev librsvg2-dev patchelf libasound2-dev libopenblas-dev libx11-dev libxtst-dev libxrandr-dev libgtk-layer-shell0 libgtk-layer-shell-dev
 
+      - name: install dependencies (ubuntu 26.04 x64)
+        if: contains(inputs.platform, 'ubuntu-26.04') && !contains(inputs.platform, 'arm')
+        run: |
+          sudo apt-get update
+          sudo apt-get install -y libwebkit2gtk-4.1-dev libayatana-appindicator3-dev librsvg2-dev patchelf libasound2-dev libopenblas-dev libx11-dev libxtst-dev libxrandr-dev libgtk-layer-shell0 libgtk-layer-shell-dev libevdev-dev xdg-utils
+
       - name: Verify gtk-layer-shell runtime dependency (Ubuntu)
         if: contains(inputs.platform, 'ubuntu')
         run: |
--- a/BUILD.md
+++ b/BUILD.md
@@ -52,9 +52,9 @@ ORT_LIB_LOCATION=$(brew --prefix onnxruntime)/lib ORT_PREFER_DYNAMIC_LINK=1 bun 
 - Install with:
 
   ```bash
-  # Ubuntu/Debian
+  # Ubuntu/Debian (22.04, 24.04, 26.04)
   sudo apt update
-  sudo apt install build-essential libasound2-dev pkg-config libssl-dev libvulkan-dev vulkan-tools glslc spirv-headers glslang-tools libgtk-3-dev libwebkit2gtk-4.1-dev libayatana-appindicator3-dev librsvg2-dev libgtk-layer-shell0 libgtk-layer-shell-dev patchelf cmake
+  sudo apt install build-essential clang libclang-dev libasound2-dev pkg-config libssl-dev libvulkan-dev vulkan-tools glslc spirv-headers glslang-tools libgtk-3-dev libwebkit2gtk-4.1-dev libayatana-appindicator3-dev librsvg2-dev libgtk-layer-shell0 libgtk-layer-shell-dev libevdev-dev patchelf cmake libopenblas-dev libx11-dev libxtst-dev libxrandr-dev
 
   # Fedora/RHEL
   sudo dnf groupinstall "Development Tools"
--- a/README.md
+++ b/README.md
@@ -137,7 +137,9 @@ For reliable text input on Linux, install the appropriate tool for your display 
 | Display Server | Recommended Tool | Install Command                                    |
 | -------------- | ---------------- | -------------------------------------------------- |
 | X11            | `xdotool`        | `sudo apt install xdotool`                         |
-| Wayland        | `wtype`          | `sudo apt install wtype`                           |
+| Wayland (GNOME/Ubuntu) | `ydotool` + `wl-clipboard` | `sudo apt install ydotool wl-clipboard` then `sudo systemctl enable --now ydotoold` and `sudo usermod -aG input $USER` (log out/in) |
+| Wayland (Sway/Hyprland) | `wtype`          | `sudo apt install wtype`                           |
+| Wayland (KDE)        | `ydotool` or `dotool` | `sudo apt install ydotool`                          |
 | Both           | `dotool`         | `sudo apt install dotool` (requires `input` group) |
 
 - **X11**: Install `xdotool` for both direct typing and clipboard paste shortcuts
@@ -232,7 +234,7 @@ The following are recommendations for running Handy on your own machine. If you 
 - **macOS**: M series Mac, Intel Mac
 - **Windows**: Intel, AMD, or NVIDIA GPU
 - **Linux**: Intel, AMD, or NVIDIA GPU
-  - Ubuntu 22.04, 24.04
+  - Ubuntu 22.04, 24.04, 26.04
 
 **For Parakeet V3 Model:**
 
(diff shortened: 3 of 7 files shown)


EmanoelDionisio:
Hey — I ran into this on Ubuntu 26.04 (GNOME/Wayland): shortcut didn't fire and paste failed with the wtype virtual keyboard error even with ydotool installed. Built and tested locally and both work now with ydotool + wl-clipboard and the user in the input group. Happy to help test anything else if needed.

koloved:
Buddy, could you test this first? https://github.com/cjpais/Handy/pull/1560 . This build was made two days ago. The goal of this build is precisely to implement the most advanced methods for using hotkeys and text insertion for Wayland. Methods that would work regardless of the external packages being used. And the use of additional tools presented here causes issues with different configurations. Therefore, if it's not too much trouble, please test this build. https://github.com/cjpais/Handy/pull/1560

The primary goal is to utilize these two methods and test them on the maximum number of Wayland devices, covering both GNOME and KDE. There is no point in implementing a bunch of secondary methods that will infinitely layer upon each other; this significantly complicates the process.
Push-to-talk with press/release via `org.freedesktop.portal.GlobalShortcuts`
Text input via `org.freedesktop.portal.RemoteDesktop`

EmanoelDionisio:
Quick update from my side: I also tested @koloved's Flatpak build (#1560) on the same machine (Ubuntu 26.04 / GNOME Wayland) and the portal approach is really good — shortcuts and paste both work natively, accents included, no external tools. Genuinely liked it and I'd recommend anyone on Wayland to try that build.

That said, I'm keeping this PR open because I think both paths have a place. #1560 is the better experience when you're on Flatpak / portal permissions. This one targets the native `.deb` install where people may not want Flatpak or can't use the portal the same way — HandyKeys + ydotool/wl-clipboard got me unblocked on a plain Ubuntu 26 install without any sandbox.

So not competing so much as covering different install paths. Happy to adjust scope if the maintainer prefers one approach over the other.

cjpais:
@koloved opinions here? and can you wrangle some of the other people who've been interested or contributed to the linux side of things. I will take a deeper look when I get a chance but I don't have the strongest opinions on Linux

EmanoelDionisio:
@koloved @cjpais — just to clarify on the cross-distro / "does this break other setups?" question, since that came up in the thread:

From what I can tell testing on my side, neither approach is meant to change behaviour on X11 or on non-Linux platforms. The problems we hit are specific to **GNOME on Wayland** (Mutter doesn't expose the virtual keyboard protocol wtype needs, and Tauri's global-shortcut plugin doesn't grab keys on Wayland in general).

**My PR (#1568):** everything runtime-related is behind `#[cfg(target_os = "linux")]`. On **X11** the keyboard backend stays `Tauri` and paste still prefers `xdotool` — Ubuntu 22.04/24.04 on X11, Fedora X11, etc. should behave as before. The new defaults/chains only kick in when `WAYLAND_DISPLAY` / `XDG_SESSION_TYPE=wayland` is detected, with GNOME-specific ordering on top of that. macOS and Windows are untouched.

**#1560 (portal/Flatpak):** same idea in practice — portals are the Wayland-native path. On X11 or distros without portal support it falls back to what was already there.

So this isn't "Ubuntu 26 changes Handy for everyone" — it's fixing a Wayland/GNOME gap that shows up hard on newer Ubuntu (26.04 ships Wayland by default), but the same Mutter limitations exist on GNOME Wayland elsewhere too. Other distros and sessions shouldn't be affected unless they're already on Wayland and hitting the same compositor limits.

Happy to be wrong here — if anyone sees a regression on X11 or a non-GNOME setup, please shout. That's exactly the kind of feedback we need.

koloved:
@cjpais @EmanoelDionisio First, should be used `org.freedesktop.portal.RemoteDesktop` If the user has installed a Atomic distr that does not have all of this tools "ydotool", "wl-clipboard", "wtype", "xdotool" .  

For users who on regular distributions, it might be the case so It's better to use these tools, although I'm not sure about that because on my side these tools (I don't remember which one) caused symbol issues.
That is why [this PR was created with DoTool](https://github.com/cjpais/Handy/pull/1299) which supposedly solves some kind of problems with these tools, but we haven't fully tested it yet. 

I think we need to first test all possible configurations on Wayland using the first remote desktop option.

tim-seoss:
You might consider testing a build against Tauri with this PR https://github.com/tauri-apps/global-hotkey/pull/172 (an enhancement which allows using `org.freedesktop.portal.GlobalShortcuts` on Wayland automatically).

PR #1576, opened 2026-07-02
Disable update checks for Nix/NixOS packaging

xilec:
## Before Submitting This PR

## Human Written Description

On Nix/NixOS, a downloaded/updated binary from the app's self-updater wouldn't actually work — packages there are immutable under `/nix/store`, and update management is expected to be handled centrally through the system configuration instead. So I disabled the auto-updater for Nix/NixOS builds: showing an update mechanism that can never succeed isn't just useless, it's actively misleading to users. This is implemented via a `HANDY_DISABLE_UPDATER` environment variable, which the Nix package sets automatically.

## Related Issues/Discussions

Fixes #
Discussion:

## Community Feedback

No discussion opened yet — this addresses a packaging-level constraint (Nix/NixOS) rather than a general user-facing feature request.

## Testing

- `cargo check` / `cargo fmt --check` / `cargo clippy --no-deps` all pass
- Manual GUI testing on Linux (debug build): with `HANDY_DISABLE_UPDATER=1`, the Debug Settings toggle is locked with an explanatory description, and the tray's "Check for Updates" item is absent from the menu; without the env var, the existing manual toggle behavior (greyed-out tray item when disabled) is unchanged
- Not tested on macOS/Windows

Screenshot from 2026-07-02 12-51-26
Screenshot from 2026-07-02 12-52-42

## AI Assistance

**If AI was used:**

- Tools used: Claude Code (Sonnet 5)
- How extensively: Implementation (Rust backend, frontend, `flake.nix`), and manual test logs verification were done with Claude Code assistance.


xilec:
@y0usaf @kakapt — could you take a look when you have a minute?

xilec:
@y0usaf @kakapt could you take another look when you have a moment?

src-tauri/src/settings.rs
```diff
@@ -980,6 +981,10 @@ pub fn get_settings(app: &AppHandle) -> AppSettings {
         store.set("settings", serde_json::to_value(&settings).unwrap());
     }
 
+    if update_checks_forced_disabled() {
+        settings.update_checks_enabled = false;
```
y0usaf: This is meant to be a runtime override, but it can leak into the persisted setting. Most setting commands call `get_settings()`, change an unrelated field, then pass the whole value to `write_settings()`. With `HANDY_DISABLE_UPDATER` set, that writes `update_checks_enabled = false` to disk, so update checks stay disabled if the user later switches to a non-Nix build. Could we keep the stored preference unchanged and apply this only when deriving the effective updater behavior?

xilec: Good catch, thanks. You're right — most commands round-trip the whole settings object through write_settings() (set_log_level, audio/shortcut changes, ...), so the forced-off value was leaking to disk and stayed disabled after moving to a non-Nix build. It's technically recoverable by toggling it back on in Settings, but that default behavior is unintuitive, so it's worth fixing.

Fixed by keeping the stored preference untouched and deriving the effective state at the point of use via update_checks_effectively_enabled() — the backend update triggers and the frontend UpdateChecker now treat HANDY_DISABLE_UPDATER as an overlay rather than writing it through.

xilec:
@y0usaf, I believe PR is ready for another look whenever you have a moment. Thanks!

y0usaf (changes requested):
If `isUpdateChecksLocked` rejects, `updateChecksLocked` stays null (settingsStore.ts:594) and UpdateChecker reads null as still-loading, so update checks never start on non-Nix builds. `set({ updateChecksLocked: false })` in the catch would cover it.

xilec:
@y0usaf Fixed in 348260c1 — the catch now falls back to `set({ updateChecksLocked: false })`.

xilec:
@y0usaf, the fix for your last point landed in 348260c1 — the `catch` now falls back to `updateChecksLocked: false`, so a failed lock-state query no longer keeps update checks from starting on non-Nix builds.

Do you see anything else worth addressing, or could you approve the PR if it looks good to you now? Thanks!

y0usaf:
Sorry for letting this hang, candidly I'm not making use of Handy regularly but am happy to test here and there. As I last recall this had no issues though :)

xilec:
@cjpai I believe this PR is ready to merge.

PR #1582, opened 2026-07-02
faster mic initialization

cjpais:
attempt at fixing #1283

* dont enumerate devices on recording start if device can be resolved
* cache device config per device
* only probe ioreg on macOS if clamshell mic is configured

### DIFF
--- a/src-tauri/src/actions.rs
+++ b/src-tauri/src/actions.rs
@@ -443,6 +443,7 @@ impl ShortcutAction for TranscribeAction {
         let rm = app.state::<Arc<AudioRecordingManager>>();
 
         // Load ASR model and VAD model in parallel
+        let kickoff_started = Instant::now();
         tm.initiate_model_load();
         let rm_clone = Arc::clone(&rm);
         std::thread::spawn(move || {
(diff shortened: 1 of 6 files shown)


xilec:
Tested this locally by building the branch (`nix develop -c bun run tauri dev`) and comparing mic-init latency against production 0.9.0, using the `Microphone stream initialized in Xms` debug log line.

Ran the comparison twice on the same machine (Linux, ALSA `default` device, no pinned microphone), once under KDE Plasma and once under niri, to rule out compositor-specific effects:

| | KDE (prod → PR) | niri (prod → PR) |
|---|---|---|
| avg mic-init | ~190ms → ~12ms | ~156ms → ~10.4ms (warm) |
| speedup | ~15x | ~15x |

The win comes from `config_cache` in `recorder.rs` (`fetch_config` drops to a few µs once cached; `build_stream` itself, ~9-13ms, isn't cacheable). The `cached_device` cache wasn't exercised in my setup since I don't have a specific mic pinned in settings.

Also confirmed the dropped-first-chunk fix (`Cmd::Start` processed before the chunk instead of after) is working — the new instrumentation shows the first captured chunk is available ~350-400µs after `Cmd::Start`, instead of being silently dropped.

And honestly — I can feel the difference just by using it, without even looking at the logs.

cjpais:
@xilec This is great news, especially that it works nicely across platforms and that is significantly faster. I will probably be merging this later today after a second review over the code

PR #1586, opened 2026-07-03
fix hf downloads + custom models in models dir

cjpais:
fix for #1579


cjpais:
Mostly this is already resolved on the main branch. So for now, I'm closing this

PR #1232, opened 2026-04-05
feat: add accent color picker to customize app theme

a81x:
## Summary

Adds a color picker under **Advanced > Appearance** that lets users choose any accent color for the app. The default pink theme is preserved — users who don't touch the setting see no change.

- Pick any color via native color picker or paste a hex value
- Full palette (buttons, icons, sidebar, overlay, hover states) is derived automatically from a single color via HSL manipulation
- Reset button restores the original pink theme
- Setting is persisted in `AppSettings` and validated on both frontend and backend

## How it works

A single `accent_color` string field stores the hex value (empty = default pink). On the frontend, `applyAccentColor()` converts the chosen color into 5 CSS custom properties (`--color-background-ui`, `--color-logo-primary`, `--color-logo-stroke`, `--color-accent-bar`, `--color-accent-hover`) applied to `:root`. When reset, inline styles are removed and CSS defaults take over.

Hardcoded `#FAA2CA` references in icons and AudioPlayer now use `var(--color-logo-primary)` with fallback, so they respond to theme changes automatically.

## Changes

**Backend (3 files, +22 lines):**
- `settings.rs` — `accent_color: String` field with serde default
- `shortcut/mod.rs` — `change_accent_color_setting` command with hex validation
- `lib.rs` — command registration

**Frontend (14 files, +211 lines):**
- `accentColor.ts` — hex→HSL palette derivation + DOM application (69 lines)
- `AccentColorPicker.tsx` — color input + hex text field + reset button (86 lines)
- `AdvancedSettings.tsx` — new "Appearance" settings group
- `App.tsx` / `RecordingOverlay.tsx` — apply accent from settings
- Icons + AudioPlayer — `var(--color-logo-primary)` with fallback
- CSS — default `--color-accent-bar` / `--color-accent-hover` variables
- i18n, bindings, settingsStore — minimal wiring

## Screenshots

The color picker sits alongside other settings in the familiar horizontal layout:

image

PR #1599, opened 2026-07-04
feat: add Light / Dark / System appearance selector

kud:
## Before Submitting This PR

## Human Written Description

I noticed Handy already ships both a light and a dark palette, but it always follows the OS — there's no way to choose. To me this isn't really a "community" ask, it feels like an obvious, expected behaviour: loads of apps let you pick. My own case is that my macOS is set to light, but I keep a lot of apps in dark because I prefer them that way. I'd like Handy to be one of them.

To be clear about the earlier rejected PRs (#1232, #551): those asked to **change Handy's brand colours**. This does not. It doesn't add or alter a single colour — it only lets me pick which of the two palettes you already designed is used, exactly like the macOS "Appearance" setting itself. `System` stays the default and keeps today's behaviour.

## Related Issues/Discussions

Discussion: https://github.com/cjpais/Handy/discussions/1598

## Community Feedback

Opened a discussion to gauge support: https://github.com/cjpais/Handy/discussions/1598 — happy to gather more before this is considered.

## Testing

Static checks (all green): `tsc --noEmit`, `eslint src`, `prettier --check`, `cargo fmt`.

Expected behaviour of the new **Appearance** dropdown (Settings → About):

- `System` — removes the `data-theme` attribute; app follows OS `prefers-color-scheme` (unchanged default behaviour).
- `Light` — forces the light palette even while the OS is in dark mode.
- `Dark` — forces the dark palette even while the OS is in light mode.
- The choice persists (stored in `AppSettings`) and is applied synchronously before React mounts, so there's no flash of the wrong palette on launch.

> Note: this still needs a manual click-through in a normal `bun run tauri dev` session; CI also compiles the backend. Marking the PR as **draft** until that's done — flagging honestly rather than claiming a run I couldn't complete on my Intel Mac (unrelated `ort`/ONNX Runtime prebuilt-binary issue for `x86_64-apple-darwin`).

## What changed (technical)

- **Backend:** new `Theme` enum (`System`/`Light`/`Dark`) in `settings.rs`, a `theme` field on `AppSettings` (defaults to `System`, `serde` default keeps existing settings files valid), and a `change_theme_setting` command registered in `collect_commands!`.
- **CSS:** added `:root[data-theme="light"]` / `:root[data-theme="dark"]` overrides in `App.css`. Their higher specificity beats the existing `prefers-color-scheme` media query, which is left untouched for `System`.
- **Frontend:** `theme.ts` util applies the mode to `document.documentElement.dataset.theme` and mirrors it to `localStorage` for flash-free boot; a `ThemeSelector` dropdown next to the existing language selector; wired through the settings store like every other setting. English i18n keys added (other locales fall back to English).
- **Scope:** the recording overlay is intentionally untouched — it has its own deliberate styling.

## AI Assistance

**If AI was used:**

- Tools used: Claude Code
- How extensively: AI wrote the implementation code following the existing `sound_theme` / `app_language` patterns in the codebase; the description and reasoning above are my own.


kud:
Thanks @cjpais !

kud:
Follow-up: the appearance selector shipped here missed two spots, fixed in #1659.

1. **Recording overlay didn't follow the setting.** The overlay is a separate webview. The `:root[data-theme=…]` override CSS lived in `App.css` (only the settings window loads it), and `overlay/main.tsx` never set the `data-theme` attribute — so the overlay stayed on `prefers-color-scheme` and kept following the OS regardless of the choice. Fix moves the override into the shared `theme.css` and applies/syncs the theme in the overlay entry, plus a `theme-changed` event for live updates.

2. **macOS native title bar stayed light.** `window.set_theme()` (the only thing that themes the native chrome) was gated to `#[cfg(target_os = "windows")]`, deliberately — to avoid dragging the overlay along, since it's app-wide on macOS. But that's exactly the behaviour we now want, so the guard is extended to macOS: title bar + overlay move together.

cjpais:
@kud did you notice #2 on macOS or is this just from a code review? on my actual machine it switches just fine

Issue #956, opened 2026-03-04
[BUG] Modifiers as shortcut activate Handy when held


raine:
## Bug Description

Keyboard implementation: Handy Keys

1. If you have modifier such as fn or Command key as shortcut in Handy
2. And try to press some unrelated hotkey such as Command + S to save in another app
3. Handy will activate, as soon as you press down the Command key

Expected behavior as that Handy does not activate on keydown, and doesn't activate if I use the modifier key as combination with some other key. It should activate only if I just press (not hold) the modifier key.

## System Information

**App Version:**

v0.7.7

**Operating System:**

macOS 26.3 (25D125)

cjpais:
Uhhh, can you explain this further? If you assign command, and you press command it will trigger. That's expected behavior

raine:
Basically there's a difference between tapping a modifier key and holding it as part of a combo.

When you assign Command as the Handy shortcut, it currently fires on keydown - the instant the key is pressed. This means:

- You press Cmd+S to save in Safari -> Handy activates the moment Cmd goes down, before you even hit S
- You press Cmd+Tab to switch apps -> same thing, Handy fires immediately
- Any hotkey involving Cmd becomes unusable because Handy intercepts the modifier

What would be better, IMO: Handy activates only when you tap Command (press and release without combining it with another key). 

1. On keydown: remember the modifier was pressed, but don't activate yet
2. If another key is pressed while the modifier is held: cancel - it's a combo (Cmd+S, Cmd+Tab, etc.)
3. On keyup: if no other key was pressed during the hold, then activate Handy

This way, Command still triggers Handy when tapped alone, but doesn't interfere with every Cmd+X shortcut across the entire system.

If this sounds too complicated to you, then maybe the right way is to use Karabiner Elements to work around this on user level.

cjpais:
I'm not really sure 100% about this one. Let me think more. To me you just shouldn't assign command as it's an obvious conflict. But on the other hand people are going to do it because we don't explicitly block it. 

My main concern right now is that this code path affects a lot of users and we really need to know its correctness. I really don't want to have regressions in this path.

With Tauri global shortcut no one cared because you couldn't even do this

raine:
So here's a workaround with Karabiner Elements, in case anyone else runs into this:

1. Add a Karabiner rule that maps fn (tapped alone, could be Cmd as well or any modifier) to F18, while still passing through fn normally when held with other keys:

```json
{
[...]

PR #914, opened 2026-02-28
Start transcription on key up

AlexanderYastrebov:
The change does not affect PTT which starts on key down as before.

Fixes #898

## Before Submitting This PR

**Please confirm you have done the following:**

- [ ] I have searched [existing issues](https://github.com/cjpais/Handy/issues) and [pull requests](https://github.com/cjpais/Handy/pulls) (including closed ones) to ensure this isn't a duplicate
- [x] I have read [CONTRIBUTING.md](https://github.com/cjpais/Handy/blob/main/CONTRIBUTING.md)

**If this is a feature or change that was previously closed/rejected:**

- [ ] I have explained in the description below why this should be reconsidered
- [ ] I have gathered community feedback (link to discussion below)

## Related Issues/Discussions

Fixes #
Discussion:

## Testing

`bun tauri dev`

## AI Assistance

- [x] No AI was used in this PR
- [ ] AI was used (please describe below)

**If AI was used:**

- Tools used:
- How extensively:

PR #1600, opened 2026-07-04
fix: modifier-only shortcuts now activate on a single tap instead of on every press

JereSalo:
## Before Submitting This PR

## Human Written Description

With Handy I like using the Right Option key to toggle recording (Push To Talk disabled) but I don't think that the current approach of starting the recording on press down provides the best experience. The reasoning is that if I use Option to move across words with `Option + Left/Right Arrow` I don't want Handy to start recording, and the simplest way of solving it without being too invasive is to just enable handy on an actual tap and not when holding down a key. I find 250ms to be a reasonable amount of time to differentiate tap from hold. 

I think this is a straight improvement of the app given that starting the recording on press down is valuable when Push To Talk is enabled and not when it's disabled. If we were to be more conservative it could be a toggle in the UI but I don't see the benefits of the current option.

## Related Issues/Discussions

Partially Addresses #956, though this issue mentions two things:
- Handy does not activate on keydown
- Handy doesn't activate if I use the modifier key as combination with some other key

The PR fixes just the first one but the second one is not fully addressed because a proper fix for that would be a bit more invasive. However, in regular conditions I'd say that it solves both because when using more than one key the user generally doesn't just press the modifier-key for less than 250ms, but it's usually held and what's tapped is the other key (for example, in `Option + Left Arrow` I press option and then tap on the arrow. Handy doesn't activate in this branch).

I see that something along the lines has been done in #914 but the approach is not the same and perhaps the current one may be preferred. I understand if this PR is closed because I've seen some comments in issues mentioning this should probably be handled in `handy-keys`. Sorry if opening this was inadequate, I just wanted to give it a try. I'll continue to use my fork of Handy if that's the case. Thanks.

## Testing

- I compiled and installed the app and it's what I'm currently using. Everything working perfectly fine. Tried other shortcuts like `Option + Space` and it also works well.

## AI Assistance

**If AI was used:**

- Tools used: Claude Code (Opus 4.8 xhigh)
- How extensively: Just thought of the idea myself and I told it to apply it in the code. Iterated a bit to make it minimal.

### DIFF
--- a/src-tauri/src/transcription_coordinator.rs
+++ b/src-tauri/src/transcription_coordinator.rs
@@ -8,6 +8,7 @@ use std::time::{Duration, Instant};
 use tauri::{AppHandle, Manager};
 
 const DEBOUNCE: Duration = Duration::from_millis(30);
+const MODIFIER_TAP_MAX_DURATION: Duration = Duration::from_millis(250);
 
 /// Commands processed sequentially by the coordinator thread.
 enum Command {
@@ -49,6 +50,7 @@ impl TranscriptionCoordinator {
             let result = std::panic::catch_unwind(std::panic::AssertUnwindSafe(|| {
                 let mut stage = Stage::Idle;
                 let mut last_press: Option<Instant> = None;
+                let mut modifier_only_press: Option<(String, Instant)> = None;
 
                 while let Ok(cmd) = rx.recv() {
                     match cmd {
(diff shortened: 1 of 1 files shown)


JereSalo:
Solution is still not ideal, this was just a quick fix. A more proper solution can be implemented once this becomes more important.

Issue #1601, opened 2026-07-05
[BUG] Voxtral Small 24B: every transcription fails with 'invalid argument' (Segment timestamps forced for models advertising InitialPrompt)


btoaldas:
## Environment
- Handy v0.9.0, macOS 26 (Apple M5 Pro, 64 GB), Metal backend
- Model: `handy-computer/Voxtral-Small-24B-2507-gguf/Voxtral-Small-24B-2507-Q5_K_M.gguf`

## Symptom
Model downloads and loads fine (`voxtral: using metal backend: MTL0`, load ~4.2s), but **every** transcription attempt fails:

```
[handy_app_lib::actions][ERROR] Transcription failed: transcribe-cpp transcription failed: invalid argument: run: invalid argument (status 1)
```

Voxtral Mini 4B Realtime works fine (streaming path). Whisper models work fine.

## Root cause
In `src-tauri/src/managers/transcription.rs`, the offline run path sets:

```rust
timestamps: if model_takes_initial_prompt {
    TimestampKind::Segment
} else {
    TimestampKind::None
},
```

The comment above it assumes *"Only whisper advertises InitialPrompt"* — but **Voxtral Small 24B also advertises `Feature::InitialPrompt`** (log: `initial_prompt=true`), while it does **not** support segment timestamps. transcribe.cpp then rejects the run.

## Reproduction (transcribe-cli, same model + audio)
```
$ transcribe-cli -m Voxtral-Small-24B-2507-Q5_K_M.gguf -l es --timestamps segment audio.wav
run: unsupported timestamp granularity        # <- Handy's exact failure

$ transcribe-cli -m Voxtral-Small-24B-2507-Q5_K_M.gguf -l es --timestamps none audio.wav
text: Hola, si me escuchas o no me escuchas.  # <- works, 2-3x realtime on M5 Pro
```

Same audio, with/without `--initial-prompt` makes no difference — only the timestamp granularity does.

## Suggested fix
Gate the `TimestampKind::Segment` workaround on the model family actually being whisper (the repetition-loop issue it works around is whisper-specific), not on `Feature::InitialPrompt`:

```rust
timestamps: if family_is_whisper {
    TimestampKind::Segment
} else {
    TimestampKind::None
},
```

(or query a `Feature::SegmentTimestamps`-style capability from transcribe.cpp before requesting them).

Happy to provide full debug logs if useful. Thanks for Handy — great app!

---
[...]

PR #1602, opened 2026-07-05
move to auto timestamps for all models

cjpais:
partial fix for #1601

basically we are removing the earlier hack (which had a bug), and instead use the models default timestamps in transcribe.cpp


cjpais:
for me this works, so merging, any other remaining fixes can be picked up in #1603 if needed

Issue #1358, opened 2026-05-01
[BUG] Push to talk with Super+Space keeps auto-repeating the space key


tredondo:
## Bug Description

1. Set the shortcut to super+space, and check Push to talk

<img width="942" height="508" alt="Image" src="https://github.com/user-attachments/assets/e7ea6614-d183-4595-ac9b-075da584d4a5" />

2. In Advanced, disable the Overlay

<img width="928" height="696" alt="Image" src="https://github.com/user-attachments/assets/28ecf931-e360-4fe5-bc89-c509bf67f7bc" />

3. Hold Super+SPACE and dictate
4. Notice space characters being inserted: `                                                                         ictate, dictate, dictate.`

https://github.com/user-attachments/assets/d4f7201e-0fbb-438c-adc7-9b480aeba8e8

## System Information

**App Version:**

0.8.3

**Operating System:**

Ubuntu 26.04

cjpais:
Thanks for the report, which keyboard implementation are you using? I assume global shortcut

tredondo:
@cjpais Not sure how to answer this... I assume you saw the screenshot with my settings. Are there other keyboard implementations? I'll paste the rest of my settings below.

<img width="993" height="1152" alt="Image" src="https://github.com/user-attachments/assets/31e609e3-4ccd-4dac-aea9-3ae97e45c880" />

cjpais:
Sorry if you turn on experimental features there should be another option

PR #1605, opened 2026-07-05
fix: push-to-talk rapidly toggles recording under X11 auto-repeat (defer synthesized releases)

### LINKED ISSUE #1539: [BUG] Push-to-talk rapidly toggles recording while the key is held (X11 auto-repeat synthesizes KeyRelease/KeyPress)
## Bug Description

In **push-to-talk** mode on X11, holding the shortcut down does not record continuously.
Instead recording starts and stops many times per second for as long as the key is held,
so the audio stream is torn down and rebuilt in a tight loop and no usable transcription
is produced. Toggle mode (press to start / press to stop) is unaffected, because it only
acts on a single press.

The trigger is standard X11 auto-repeat. When a non-modifier key is held, the X server's
default (non-detectable) auto-repeat does **not** send one long press — it synthesizes a
`KeyRelease` immediately followed by a `KeyPress` for every repeat cycle while the key is
still physically down. Handy's global-shortcut listener appears to treat each synthesized
`KeyRelease` as "push-to-talk released" (stop) and the next synthesized `KeyPress` as
"push-to-talk pressed" (start), producing the start/stop loop.

## System Information

- **App Version:** 0.8.3
- **OS:** Fedora 44
- **Session:** X11 (lightweight window manager, no full desktop environment)
- **Mode:** Push-to-talk

## Steps to Reproduce

1. Set a push-to-talk shortcut whose activation key is a normal (auto-repeating) key,
   e.g. `Ctrl+Space`.
2. Hold the shortcut down and speak.
3. Observe recording toggling on and off repeatedly instead of staying on.

## Expected behavior

While the push-to-talk shortcut is physically held, recording should remain active
continuously, and stop once — when the key is physically released.

## AI-suggested fix

**Preferred — request detectable auto-repeat in the X11 listener.** Call
`XkbSetDetectableAutoRepeat(display, True, &supported)` once when the X11 keyboard listener
initializes. With detectable auto-repeat, the server emits a `KeyRelease` only on physical
release, so synthesized release/press pairs disappear and push-to-talk sees a single press
and a single release. Per the `XkbSetDetectableAutoRepeat` documentation this is a
**per-client** setting — it affects only Handy's connection, not other applications'
auto-repeat — so it has no side effects on the user's system and needs no cleanup on crash.

**Alternative without Xkb — debounce synthesized pairs.** A synthesized auto-repeat
`KeyRelease` and the `KeyPress` that immediately follows it share the same serial number
and timestamp. The listener can drop a `KeyRelease` when an identical `KeyPress` for the
same keycode is already pending with a matching serial/time, and only honor a release that
isn't immediately followed by a re-press.

**Also relevant — an evdev backend (see #949).** Raw evdev events distinguish auto-repeat
(`EV_KEY` value `2`) from real press (`1`) and release (`0`), so a push-to-talk
implementation reading evdev can simply ignore repeats. That backend would fix this X11
issue and the Wayland shortcut gap in #949 at the same time.

## Workarounds for affected users (until fixed)

- Use **toggle mode** instead of push-to-talk (single press to start, single press to stop).
- Or disable auto-repeat for the specific activation key at the X level (e.g. `xset -r 65`),
  accepting that auto-repeat is then off for that one key.

## Notes

This issue might have same root cause as #1358

mvanhorn:
## Before Submitting This PR

## Human Written Description

I hold my push-to-talk shortcut down to dictate, and on X11 it kept flickering recording on and off many times a second instead of just recording while I held the key. Digging into it, this is plain X11 auto-repeat: the server does not send one long press, it synthesizes a KeyRelease immediately followed by a KeyPress on every repeat cycle, and the coordinator was honoring each of those synthesized releases as a real "let go" of the key. Fixing it at the coordinator level matters because it makes push-to-talk usable on X11 without touching the keyboard backend, and it also protects against any other source of synthesized release/press pairs.

## Summary

Holding a push-to-talk key under X11 auto-repeat no longer rapidly toggles transcription on and off. `TranscriptionCoordinator` now debounces push-to-talk releases with a short grace window and cancels a pending release when a matching press arrives, so the synthesized release/press pairs collapse into one continuous recording that stops exactly once on the real physical release.

## Related Issues/Discussions

The coordinator thread already debounced rapid press events, but its comment stated "Releases always pass through for push-to-talk" and that pass-through was the bug: each synthesized KeyRelease read as "push-to-talk released" (stop) and the following synthesized KeyPress read as "push-to-talk pressed" (start). This change defers a push-to-talk release for a short grace window (50ms) instead of honoring it immediately. If a press for the same binding arrives within the window, the pending release is cancelled and recording continues uninterrupted; if the window elapses with no matching press, the release is honored and recording stops once. Toggle mode acts only on press and is unchanged. The fix lives entirely in `TranscriptionCoordinator`, so it is backend-agnostic. The decision (honor now / defer / cancel pending) is extracted into a small pure helper so it can be unit tested without a Tauri `AppHandle` or an audio device.

Fixes #1539
Discussion:

## Community Feedback

The bug is confirmed by the maintainer in #1539 (able to replicate on Fedora), and #1358 was closed as a duplicate in favor of it.

## Testing

- Added a `#[cfg(test)] mod tests` covering the pure decision helper: a push-to-talk release while recording defers, a press matching a pending release cancels it, toggle mode passes through, a press for a different binding does not cancel a pending release, and a press matching a pending release cancels regardless of debounce (so a debounced auto-repeat press still cancels the deferred stop).
- `cargo test --lib transcription_coordinator` passes (5 tests).
- `cargo fmt --check` reports no changes for `transcription_coordinator.rs`.

N/A

## AI Assistance

**If AI was used:**

- Tools used: AI coding assistant
- How extensively: Used to help implement and review the coordinator change and the unit tests; the debounce approach follows the maintainer's own suggested alternative in the issue.


mvanhorn:
Thank you @cjpais - X11 auto-repeat no longer strobes push-to-talk.

mvanhorn:
Thanks for taking this, @cjpais! Push-to-talk no longer fighting X11 auto-repeat should feel much better.

cjpais:
Thank you for the fix @mvanhorn!

Issue #1607, opened 2026-07-05
[BUG]  0.9.0 libopenblas.so.0 not found


Christopher761:
## Before You Submit

**Please search [existing issues](https://github.com/cjpais/Handy/issues) to avoid duplicates.** Your bug may already be reported! Right now it's just me maintaining this project so many issues can be overwhelming! Help me out by checking first.

## Bug Description

Installed wtih the package manager.  Run from icon, it simply fails to start.  Running it from the command line, it give an error:
handy: error while loading shared libraries: libopenblas.so.0: cannot open shared object file: No such file or directory

Handy Version:
9.0

Ubuntu 26.4 / Gnome / Wayland

SOLUTION:
=========

sudo apt update
sudo apt upgrade
sudo apt install libopenblas-dev

After that the program starts.

(Still not working with the hot key, and I have to used "handy --toggle-transcription" workaround.)

koloved:
Could you please test this build if you have a flatpak. You don't need any workaround for the hotkey. https://github.com/cjpais/Handy/pull/1560#issuecomment-4886091447

Christopher761:
Yes.  I will put comments in #1560.

cjpais:
#1621

cjpais:
@Christopher761 can you test the build in the #1621 thread and see if it helps, maybe temporarily uninstalling libopenblas-dev on your machine?

cjpais:
Should be resolved in the next release

PR #1621, opened 2026-07-07
add openblas to ci and packaging for linux

cjpais:


cjpais:
Should resolve #1607 as well

Issue #74, opened 2025-08-07
Windows 11, Aesthetic of the GUI hint when listening are off


ABODFTW:
As can be seen in the attached video, when I press the shortcut, I see this nice GUI, but it is off because of the horizontal and vertical scroll bars, is there a way to fix this?

I could work on this if can be pointed into the right direction.

https://github.com/user-attachments/assets/4fa5a4a7-4f53-4d8c-8d30-53ea0b0e9d7e

And thanks for making such a great open-source tool

cjpais:
It's funny I ran into this while testing and thought I had gotten it to go away but I guess not on all machines. There's a width and height param for when the window is created and modifying that alone should help

I know @vladstudio has an open PR which tweaks the styling and lets you choose position, I will probably be reviewing it in a few hours and will try to modify if I can.

OneAbodFTW:
That would be fantastic

cjpais:
If you don't mind, could you give 0.3.6 a shot and see if it fixes this issue? I can't say for certain that it will. It doesn't show up on my machine on Windows, however, it may still be an issue on yours, and if that's the case, I'll tweak the values a little bit

OneAbodFTW:
this works great, it looks great no scroll bars, thank you also that update flow was very seamless

cjpais:
I'm super glad to hear for both. Thank you for testing it!

PR #1633, opened 2026-07-08
Use configurable transcribe.cpp output budget

vladkalinichencko:
Depends on handy-computer/transcribe.cpp#74. Until that is merged and released, the Cargo patch points to the PR branch so this integration can be reviewed and tested against the exact implementation.

This keeps the Handy-side change small:
- request a 4096-token output budget for transcribe.cpp runs
- keep non-empty partial text when transcribe.cpp reports `OUTPUT_TRUNCATED`
- show the backend error message when history re-transcription fails
- update the integration to the current transcribe.cpp 0.2.0 API

Validated with:
- `cargo fmt --check`
- `cargo check`
- 151 Rust tests
- frontend production build
- full Rust build
- headless transcription through Handy with an installed Whisper Medium model on Metal

After #74 is released, the temporary Cargo patch entries can be replaced by the released crate version.

### DIFF
--- a/src-tauri/Cargo.lock
+++ b/src-tauri/Cargo.lock
@@ -7195,9 +7195,8 @@ dependencies = [
 
 [[package]]
 name = "transcribe-cpp"
-version = "0.1.3"
-source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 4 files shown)


cjpais:
Can you add a screenshot for this?

vladkalinichencko:
Yep, here you go:

clipboard

Issue #1213, opened 2026-04-03
[BUG] Windows 11 after sleep mode stuck on "transcribing"


KevinUK:
## Bug Description

Handy works when I reboot the PC but stops working after putting Windows in sleep mode, even if I close and start the app again.

After sleep, when I hold ctrl + space, talk and then let go of the key combo, the UI gets stuck on "transcribing".

In task manager I see "Recording" and it goes from 9MB usage to 691MB

## System Information

**App Version:**

0.8.2

**Operating System:**

Windows 11, all up to date

**CPU:**

AMD Ryzen 7 7700X

**GPU:**

RADEON RX 6800

## Logs

[handy.log](https://github.com/user-attachments/files/26457627/handy.log)

cjpais:
Thanks for the report and log, we need to fix this

@VirenMohindra if you get a chance would love a look, otherwise I'll get to this in a few days or sooner

Avatarsia:
### Confirming this issue — detailed reproduction + log evidence

**System:** Windows 11 Pro, NVIDIA RTX 3090 (30GB VRAM), Parakeet V3 (`parakeet-tdt-0.6b-v3`, CPU), AirPods Pro (Bluetooth)
**Version:** v0.8.x (auto-updated), push-to-talk mode (Ctrl+Space), keyboard_implementation: `tauri`

---

#### Exact reproduction steps

1. **Press Ctrl+Space** to start recording
2. **Long delay** (~1-5 seconds) before the audio feedback signal plays — sometimes it never comes
3. **Release Ctrl+Space** — the "Transcribing..." overlay appears immediately
4. **Stuck forever** — the overlay never resolves, even though nothing was actually recorded
5. App must be force-killed and restarted

This happens **more frequently** after the app has been idle for a while (minutes to hours) and is consistent with Bluetooth audio device sleep/reconnection behavior.

---

#### Root cause analysis from logs

**The core issue: transcription pipeline is triggered with 0 audio samples and never returns.**

Here is the full chain visible in the logs:

**Step 1: Bluetooth device gone → fallback to wrong device**

When AirPods sleep/disconnect, Handy falls back to a different audio device with mismatched sample rates:

```
[2026-04-02][17:24:53][audio_feedback][WARN] Device 'Kopfhörer (AirPods Pro)' not found, using default device
[2026-04-02][17:24:53][audio::recorder][INFO] Using device: Ok("Mikrofon (NVIDIA Broadcast)")
[...]

PR #1642, opened 2026-07-10
debug build for #1213

cjpais:
debug #1213


cjpais:
@KevinLedge I appreciate your debug logs in the last thread, I'm just wondering if you're able to try this debug build and see if we can capture a log from this one. It should have a bit more details so we can figure out exactly where things are getting stuck and try to create the best fix possible

cjpais:
The issue has been closed, so we can close out this debug build

Issue #1349, opened 2026-04-30
[BUG] Handy occasionally locks up Pipewire on Ubuntu


xenotropic:
## Bug Description

Occasionally Handy gets stuck such that it doesn't output any text and will just get stuck either listening or processing (or at least, showing one of those icons in the tray) and will not start working again until I restart Pipewire (```systemctl --user restart pipewire pipewire-pulse wireplumber```) 

## System Information

**App Version:**

0.8.2 (just updated to 0.8.3 when it was in this failed state and that did not fix it; but if you just specifically fixed this bug in 0.8.3 a way that prevents it, then maybe I missed that)

**Operating System:**

Ubuntu 24.04

**CPU:**

Intel® Core™ i7-5600U × 4

**GPU:**

none (I mean "Intel® HD Graphics 5500" but no meaningful GPU)

## Logs

The logs are full of things I said that I do not want to publish in a bug report. So I cannot attach it. But I _will_ relay you what Gemini Pro said to tell you when I put the log into it.

> Starting at 23:56:31, your log shows that Handy is unable to initialize the microphone stream. Every time you attempt to start a transcription after this point, the application hits this specific error: Failed to build input stream: A backend-specific error has occurred: ALSA function 'snd_pcm_prepare' failed with error 'I/O error (5)'.
> 
> Because this is an underlying OS-level or driver-level error, the audio library gets stuck trying to connect. You can see this in the log where TranscribeAction::start completed in 30.423640917s — it is hitting a hard 30-second timeout before giving up on every attempt. Restarting the app does not clear the OS-level lock or deadlocked state of the audio device.
> 
> When you file an issue with the Handy STT developer, you can provide them with the following summary:
> 
> Component: handy_app_lib::audio_toolkit::audio::recorder
> 
> Error: ALSA function 'snd_pcm_prepare' failed with error 'I/O error (5)'
> 
> Behavior: Following a sudden audio backend failure, TranscribeAction::start fails to capture audio and blocks for exactly ~30 seconds per attempt.
> 
> Notes: The app's audio stream fails to recover gracefully from an EIO (Input/Output error 5). The development team likely needs to implement better device re-enumeration or shorter connection timeouts in their audio library (which appears to be cpal based on the error string formatting ) when the ALSA device drops out or the sound server deadlocks.

Happy to try things, or find specific lines from the logs, if those would be helpful.

Issue #1339, opened 2026-04-28
[BUG] Microphone not detected on Linux — sof-hda-dsp (hw:0,7) works in Vocalinux but Handy only shows "Default"


avinashkanaujiya:
## Bug Description

On my Linux system, the microphone is not detected by Handy, preventing any audio input. The same microphone works correctly in [Vocalinux](https://github.com/jatinkrmalik/vocalinux), so this appears to be specific to how Handy enumerates or selects audio input devices.

## Expected Behavior

Handy should detect and allow selection of my system microphone (`sof-hda-dsp: - (hw:0,7)`), or at minimum use it when "Default" is selected, since it is the active system input device.

## Actual Behavior

Handy only shows "Default" as the input device option. When the transcription shortcut is pressed, no audio is captured.

## Steps to Reproduce

1. Launch Handy on Linux
2. Open Settings → Input Device
3. Observe that only "Default" is listed; `sof-hda-dsp: - (hw:0,7)` is not shown
4. Press transcription shortcut — no audio is captured
5. Launch Vocalinux → same microphone works and transcribes correctly

## Environment

- OS: OpenSUSE Leap 16.0 (Host) + OpenSUSE Tumbleweed (Distrobox)
- Handy version: v0.8.2
- Audio stack: PipeWire (PulseAudio-compatible)
- Working device in Vocalinux: `sof-hda-dsp: - (hw:0,7)`

## Related Issues

- #806 — No microphone recording on Pop!_OS 24.04 (PipeWire) – cpal probes OSS but fails to use ALSA input
- #521 — Microphone and Output Device only detected as "Default" on Linux
- #1284 — Mic permission page loading forever
- #1101 — audio recording issue. Microphone initialize twice

## Additional Context

The fact that Vocalinux detects and uses this ALSA hw device suggests the underlying audio subsystem is functional, and Handy's device enumeration (via cpal) may be missing it or defaulting to a non-functional OSS probe.

koloved:
@cjpais The same problem (/dev/dsp problem) as this one https://github.com/cjpais/Handy/issues/806 , it can be closed.

cjpais:
dup of #521

Issue #806, opened 2026-02-12
[BUG] No microphone recording on Pop!_OS 24.04 (PipeWire) – cpal probes OSS but fails to use ALSA input


jnials:
## Bug Description

Handy launches successfully (shortcuts/Enigo init works), but pressing the hotkey to record does nothing useful: no transcription, and raw WAV files (if saved) are silent/empty. The terminal shows repeated ALSA OSS errors even when /dev/dsp does not exist.

## System Information

**App Version:** 0.7.5

**Operating System:** popos-24.04

**CPU:** AMD Ryzen 7 7840HS 

**GPU:** Radeon 780M Graphics

## Details

```bash

env RUST_LOG=cpal=debug,info handy
[2026-02-12][21:54:13][handy_app_lib::managers::history][INFO] Initializing database at "/home/jnials/.local/share/com.pais.handy/history.db"
[2026-02-12][21:54:14][handy_app_lib::commands][INFO] Enigo initialized successfully after permission grant
[2026-02-12][21:54:14][handy_app_lib::commands][INFO] Shortcuts initialized successfully

sudo mv /dev/dsp /dev/dsp.bak
[sudo] password for jnials:

env RUST_LOG=cpal=debug,info handy
[2026-02-12][21:56:20][handy_app_lib::managers::history][INFO] Initializing database at "/home/jnials/.local/share/com.pais.handy/history.db"
[2026-02-12][21:56:21][handy_app_lib::commands][INFO] Enigo initialized successfully after permission grant
[2026-02-12][21:56:21][handy_app_lib::commands][INFO] Shortcuts initialized successfully
ALSA lib pcm_oss.c:397:(_snd_pcm_oss_open) Cannot open device /dev/dsp
ALSA lib pcm_oss.c:397:(_snd_pcm_oss_open) Cannot open device /dev/dsp
ALSA lib pcm_oss.c:397:(_snd_pcm_oss_open) Cannot open device /dev/dsp
ALSA lib pcm_oss.c:397:(_snd_pcm_oss_open) Cannot open device /dev/dsp
```

```bash
wpctl status
PipeWire 'pipewire-0' [1.5.85, jnials@jnials-work-laptop, cookie:1514338208]
 └─ Clients:
        33. WirePlumber                         [1.5.85, jnials@jnials-work-laptop, pid:3264]
        37. pipewire                            [1.5.85, jnials@jnials-work-laptop, pid:3267]
[...]

PR #1644, opened 2026-07-10
fix(audio): bump cpal 0.16→0.17 to fix device-enumeration crash on macOS 26 (Tahoe)

### LINKED ISSUE #1643: [BUG] App segfaults (SIGSEGV) on launch enumerating audio devices on macOS 26 (Tahoe)
## Before You Submit

Searched existing open and closed issues — no report covers the macOS 26 (Tahoe) CoreAudio device-enumeration segfault. Closest issues are Linux-specific (OSS/ALSA/PipeWire: #806, #1339, #1349).

## Bug Description

On macOS 26 (Tahoe), Handy segfaults deterministically shortly after launch. The crash happens while enumerating audio devices (when the frontend requests the microphone/output device list), so the app never becomes usable.

Root cause: `cpal` 0.16 (which pulls `coreaudio-rs` 0.13) segfaults in CoreAudio device enumeration on macOS 26. Bumping `cpal` to 0.17 (`coreaudio-rs` 0.14) resolves it. The `cjpais/rodio` fork pins `cpal` 0.16, so the whole tree needs to move to `rodio` 0.22 (which uses `cpal` 0.17) to unify on a single `cpal` version.

A fix is ready in a linked PR.

## System Information

**App Version:** 0.9.0 (development build from `main`)

**Operating System:** macOS 26.5.1 (Tahoe); crash first reproduced on 26.4

**CPU:** Apple M5

**GPU:** Apple M5 (integrated)

## Logs

Native `SIGSEGV` — no app-level log is produced because the crash occurs inside cpal's CoreAudio device-enumeration path, below Handy's logging layer.

Crash signature: deterministic `SIGSEGV` ~15-22s after launch, triggered by device enumeration (e.g. opening Settings → Audio). Reproduces on every launch on macOS 26.x with `cpal` 0.16; does not reproduce after bumping to `cpal` 0.17.

xronocode:
## Before Submitting This PR

Duplicate search done: no open or closed issue/PR covers the macOS 26 (Tahoe) CoreAudio device-enumeration segfault or a cpal 0.16→0.17 bump. Closest issues are Linux-specific OSS/ALSA/PipeWire (#806, #1339, #1349).

## Human Written Description

Handy crashes hard on launch on macOS 26 (Tahoe) — it segfaults while enumerating audio devices, so the app is unusable there at all. I traced it down to cpal 0.16's CoreAudio path (coreaudio-rs 0.13); bumping cpal to 0.17 (and rodio to 0.22 so the whole tree shares one cpal) makes it survive device enumeration and run normally. This is a straight crash on a current macOS, so I wanted to get it upstream.

## Related Issues/Discussions

Fixes #1643

## Community Feedback

This is a bug fix (top priority per CONTRIBUTING) for a hard crash on macOS 26 (Tahoe). Not previously reported upstream; surfacing it as a fix alongside the bug report.

## Testing

**Root cause verified:** `cpal` 0.16 (via `coreaudio-rs` 0.13) segfaults in CoreAudio device enumeration on macOS 26. `cpal` 0.17 pulls `coreaudio-rs` 0.14, which resolves it. Verified the tree now resolves to a **single** `cpal v0.17.3` (shared by the direct dep and `rodio v0.22.2`) — `cargo tree -i cpal` confirms no duplicate cpal version.

**Build verification (on `main` @ `9b0d8a1`):**
- `cargo check --locked` → 0 errors
- `cargo clippy --locked` → no new lints (only pre-existing `device.name()` deprecation notes — see note below)
- `cargo fmt --check` → clean
- `cargo tree -i cpal` → single `cpal v0.17.3`

**Manual verification needed (hardware/macOS):** launch on macOS 26, open Settings → Audio — the app must survive device enumeration with no `SIGSEGV`. Before this patch the crash was deterministic within ~15–22s of launch.

**Scope:** 4 files — `Cargo.toml` (2 dep lines), `Cargo.lock` (regenerated), `recorder.rs` (1 line: `SampleRate` became a `u32` alias in 0.17), `audio_feedback.rs` (5 lines: rodio 0.22 renamed `OutputStreamBuilder` → `DeviceSinkBuilder`). No recording/enumeration logic changed; Linux host selection (`get_cpal_host()` → ALSA) is untouched.

**Cross-platform note:** `cpal` 0.17 + `rodio` 0.22 are current upstream releases; the change is platform-agnostic at the API level (only the `SampleRate` tuple→alias and the rodio builder rename). The `cjpais/rodio` fork (rodio 0.20, which pinned cpal 0.16) is dropped in favor of upstream rodio 0.22.

**Known follow-up (intentionally NOT in this PR):** `cpal` 0.17 deprecates `Device::name()` in favor of `description()`/`id()`. That migration is **not** a drop-in — `description()` returns richer strings (name + manufacturer) that could stop matching saved `selected_microphone`/`selected_output_device` settings, so it deserves its own PR and separate testing. Left as 4 deprecation warnings for now.

_Native `SIGSEGV` crash produces no in-app visual; repro is "launch on macOS 26 → Settings → Audio → SIGSEGV". Happy to capture a no-crash video on request._

## AI Assistance

**If AI was used:**
- **Tools used:** Claude (Kilo CLI) for analysis, patch porting, and verification; the same fix was independently shipped and verified in production in a downstream fork (TokMo).
- **How extensively:** AI assisted with identifying the cpal 0.16/coreaudio-rs 0.13 root cause, porting the dependency bump and the two API-break fixups onto current `main`, and running the build/clippy/fmt verification. The crash reproduction and the decision to migrate to cpal 0.17 were driven by real macOS 26 hardware testing in the fork.

### DIFF
--- a/src-tauri/Cargo.lock
+++ b/src-tauri/Cargo.lock
@@ -69,9 +69,9 @@ checksum = "683d7910e743518b0e34f1186f92494becacb047c7b6bf616c96772180fef923"
 
 [[package]]
 name = "alsa"
-version = "0.9.1"
+version = "0.11.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "ed7572b7ba83a31e20d1b48970ee402d2e3e0537dcfe0a3ff4d6eb7508617d43"
+checksum = "812947049edcd670a82cd5c73c3661d2e58468577ba8489de58e1a73c04cbd5d"
 dependencies = [
  "alsa-sys",
  "bitflags 2.11.0",
@@ -81,9 +81,9 @@ dependencies = [
 
 [[package]]
 name = "alsa-sys"
-version = "0.3.1"
+version = "0.4.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "db8fee663d06c4e303404ef5f40488a53e062f89ba8bfed81f42325aafad1527"
+checksum = "ad7569085a265dd3f607ebecce7458eaab2132a84393534c95b18dcbc3f31e04"
 dependencies = [
  "libc",
  "pkg-config",
@@ -776,6 +776,17 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "613afe47fcd5fac7ccf1db93babcb082c5994d996f20b8b159f2ad1658eb5724"
 
 [[package]]
+name = "chacha20"
+version = "0.10.1"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "d524456ba66e72eb8b115ff89e01e497f8e6d11d78b70b1aa13c0fbd97540a81"
+dependencies = [
+ "cfg-if",
+ "cpufeatures 0.3.0",
+ "rand_core 0.10.1",
+]
+
+[[package]]
 name = "chrono"
 version = "0.4.43"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -1043,11 +1054,11 @@ dependencies = [
 
 [[package]]
 name = "coreaudio-rs"
-version = "0.13.0"
+version = "0.14.2"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "1aae284fbaf7d27aa0e292f7677dfbe26503b0d555026f702940805a630eac17"
+checksum = "7d5d7dca3ebcf65a035582c9ad4385371a9d9ee6537474d2a278f4e1e475bb58"
 dependencies = [
- "bitflags 1.3.2",
+ "bitflags 2.11.0",
  "libc",
  "objc2-audio-toolbox",
  "objc2-core-audio",
@@ -1057,9 +1068,9 @@ dependencies = [
 
 [[package]]
 name = "cpal"
-version = "0.16.0"
+version = "0.17.3"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "cbd307f43cc2a697e2d1f8bc7a1d824b5269e052209e28883e5bc04d095aaa3f"
+checksum = "d8942da362c0f0d895d7cac616263f2f9424edc5687364dfd1d25ef7eba506d7"
 dependencies = [
  "alsa",
  "coreaudio-rs",
@@ -1072,13 +1083,17 @@ dependencies = [
  "ndk-context",
  "num-derive",
  "num-traits",
+ "objc2",
  "objc2-audio-toolbox",
+ "objc2-avf-audio",
  "objc2-core-audio",
  "objc2-core-audio-types",
+ "objc2-core-foundation",
+ "objc2-foundation",
  "wasm-bindgen",
  "wasm-bindgen-futures",
  "web-sys",
- "windows 0.54.0",
+ "windows 0.61.3",
 ]
 
 [[package]]
@@ -1086,6 +1101,15 @@ name = "cpufeatures"
 version = "0.2.17"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "59ed5838eebb26a2bb2e58f6d5b5316989ae9d08bab10e0e6d103e656d1b0280"
+dependencies = [
+ "libc",
+]
+
+[[package]]
+name = "cpufeatures"
+version = "0.3.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "8b2a41393f66f16b0823bb79094d54ac5fbd34ab292ddafb9a0456ac9f87d201"
 dependencies = [
  "libc",
 ]
@@ -2265,6 +2289,7 @@ dependencies = [
  "cfg-if",
  "libc",
  "r-efi",
+ "rand_core 0.10.1",
  "wasip2",
  "wasip3",
 ]
@@ -3377,9 +3402,9 @@ checksum = "c41e0c4fef86961ac6d6f8a82609f55f31b05e4fce149ac5710e439df7619ba4"
 
 [[package]]
 name = "mach2"
-version = "0.4.3"
+version = "0.5.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "d640282b302c0bb0a2a8e0233ead9035e3bed871f0b7e81fe4a1ec829765db44"
+checksum = "6a1b95cd5421ec55b445b5ae102f5ea0e768de1f82bd3001e11f426c269c3aea"
 dependencies = [
  "libc",
 ]
@@ -3853,6 +3878,16 @@ dependencies = [
 ]
 
 [[package]]
+name = "objc2-avf-audio"
+version = "0.3.2"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "13a380031deed8e99db00065c45937da434ca987c034e13b87e4441f9e4090be"
+dependencies = [
+ "objc2",
+ "objc2-foundation",
+]
+
+[[package]]
 name = "objc2-cloud-kit"
 version = "0.3.2"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -3873,6 +3908,7 @@ dependencies = [
  "objc2",
  "objc2-core-audio-types",
  "objc2-core-foundation",
+ "objc2-foundation",
 ]
 
 [[package]]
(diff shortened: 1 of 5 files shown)


cjpais:
Most likely, this cannot be merged as is largely because there are other changes in the libraries that affect other platforms that we have to be aware of and need to be testing, especially as it goes to device enumeration. So, certainly this can't be pulled in as is and I do still need more information about your specific system

xronocode:
Agreed, and thanks for calling that out. This should be treated cautiously because the dependency bump affects audio backends beyond the specific macOS crash I hit.

The patch itself is intentionally small: cpal 0.16 -> 0.17, moving rodio off the cpal-0.16-pinned fork to rodio 0.22, plus the two required API fixups (`SampleRate` and `DeviceSinkBuilder`). I did not change Handy's device-enumeration or recording logic directly.

What I verified locally:

- `cargo check --locked` passes
- the macOS crash no longer reproduces on my affected machine
- the affected machine has virtual devices installed: BlackHole 2ch and Xronophone Microphone, plus built-in devices

What I have not verified:

- Linux audio backends
- Windows audio backends
- a wider macOS device matrix

So I agree this may not be merge-ready without broader testing. If there is a specific test matrix or smoke test you want run for Linux/Windows/macOS, I can help with the parts I can reproduce locally and keep the PR scoped to the dependency bump only.

xronocode:
Hi @cjpais, gentle nudge on this one. I know you flagged that the cpal 0.16 to 0.17 bump pulls in other library changes that need care. Happy to do that legwork if it helps: I can rebase on current main, run the full build and tests locally, and write up exactly which transitive deps changed and what I verified, so there's less for you to dig through. Let me know if that's useful or if you'd rather take it a different way. No rush.

cjpais:
I mean, if you want to take this up and are serious about getting this in, I need you to basically reach out across platforms and get validation on a wide range of devices from people in the community. It's just unlikely I can make a dependency bump without large amounts of testing because it seems likely to bring in new bugs that at the moment I'm not really willing to because, as far as I can tell, you're one of the only people running into the crash. Either that or I need more reports on the crash itself to make it a higher priority. But right now, this is just not a priority. There are so many other issues that are affecting more users with less risk, so that has to take my priority at the end of the day

xronocode:
Following up on the offer above: I rebased this onto current `main` and ran the local checks. Below is the result and where I would focus review.

## What I did
Rebased `fix/cpal-017-macos26-crash` (1 commit) onto current `upstream/main` (`b428ae4`),
so it is no longer a month old. Then ran the library-level local checks (the exact `--lib` commands in Checks below).

## Checks (on rebased commit `5efa5b4`, base `b428ae4`)
- `cargo build --lib`: success (exit 0).
- `cargo test --lib`: 177 passed, 0 failed.
- `cargo clippy --lib`: 0 errors. 1 pre-existing warning (`transcription.rs:1178`) unrelated to this change. The 7 `Device::name()` deprecation warnings that cpal 0.17 introduces are silenced with `#[allow(deprecated)]` (see below).
- Note: these are the `--lib` checks only. The stricter `cargo clippy --all-targets -- -D warnings` is already red on `upstream/main` from pre-existing warnings unrelated to this change (transcription.rs, lib.rs, gguf_meta.rs, portable.rs, etc.), so this PR neither turns it green nor adds to it.
- Cargo.lock is clean: only the audio/RNG cluster changed; spot-checked tauri, serde, tokio, reqwest, rubato — all identical to `upstream/main` (no accidental downgrades of main's deps).

## The dependency cascade (what you flagged as needing care)
13 packages change in `Cargo.lock`, all resolving to two roots: `cpal 0.17` and `rodio 0.22`. Nothing outside the audio stack moved.

Audio / OS-interop (the part that touches hardware):
| crate | old | new |
|---|---|---|
| cpal | 0.16.0 | 0.17.3 |
| rodio | 0.20.1 (cjpais fork) | 0.22.2 (published) |
| coreaudio-rs | 0.13.0 | 0.14.2 (macOS CoreAudio) |
| mach2 | 0.4.3 | 0.5.0 (macOS Mach, realtime thread priority) |
| alsa | 0.9.1 | 0.11.0 (Linux ALSA) |
| alsa-sys | 0.3.1 | 0.4.0 (Linux ALSA FFI) |
| objc2-avf-audio | — | 0.3.2 (ADDED, macOS AVFoundation Audio) |
| rtrb | — | 0.3.4 (ADDED, rodio mixer ring buffer) |

Plus a small RNG/crypto sub-cascade from rodio 0.22's new `rand_distr` use: `rand 0.9.4 -> 0.10.1`, `rand_core -> 0.10.1`, added `rand_distr 0.6`, `chacha20 0.10`, `cpufeatures 0.2 -> 0.3`. These only affect rodio's white-noise generation, not audio hardware.

## Where to focus review
- The macOS HAL cluster is the real target for the macOS 26 crash fix: `cpal 0.17` + `coreaudio-rs 0.13 -> 0.14` + `mach2 0.4 -> 0.5` + the newly added `objc2-avf-audio`. The new AVFoundation Audio linkage is the most likely place for an unexpected device-enumeration or latency regression on macOS 26.
- Linux ALSA got the largest jump (`alsa 0.9 -> 0.11`), so non-macOS behavior is changing under the same bump even though the PR targets macOS.

## Deprecation handling
cpal 0.17 deprecated `Device::name()` in favor of `description()` (name + manufacturer + type) and `id()` (stable identifier). The existing `.name()` call sites still compile and work; rather than change behavior, I silenced the 7 new warnings with `#[allow(deprecated)]` plus a comment. Migrating to `id()`/`description()` would change both the device strings users have saved in settings and the picker text (and `name` vs `description` vs `id` is a per-call-site decision), so that belongs in a separate PR. Behavior is unchanged.

## Net
The bump is rebased on current main, builds, and passes all 177 lib tests with no dependency downgrades. The cascade is contained to the audio stack; the macOS CoreAudio/Mach/AVFoundation cluster is what actually needs eyes for the macOS 26 fix.

xronocode:
Fair enough, that makes sense. A dependency bump is a big surface to take on for a crash only one person has hit, and I don't want to rush it either. I'll leave the branch rebased on main and parked here rather than pushing further on it. If more reports of the macOS 26 enumeration crash come in, happy to pick it back up. Appreciate the honest call.

rdslw:
A note in case this is picked up again: cpal 0.17.x does not fix the ALSA capture stall I hit on Linux (#2113).

If a bump is ever considered, 0.18.2 is the version that matters for Linux capture, not 0.17.

cjpais:
Thanks @rdslw, @xronocode what do you think of that version?

I do eventually plan to pull this in. It just will take me a while to get to it so we can do validation across operating systems and add whatever other necessary code to do this correctly

xronocode:
0.18.2 is the better target imo. It keeps the CoreAudio enumeration fixes we're riding here and adds the ALSA capture work rdslw described, and the Linux default host moves to PipeWire first. So both platforms land in one bump instead of a second PR later. I bumped it locally to check the scope: the breakage stays within the same three files this PR already touches, mostly mechanical (name() is Display now, unified error type, StreamConfig by value), plus wiring an explicit play() where streams used to auto-start. If you're good with that I can rebase onto 0.18.2 and re-run the local checks. @rdslw once this builds, would you be up for retesting the capture stall on your side?

rdslw:
> If you're good with that I can rebase onto 0.18.2 and re-run the local checks. @rdslw once this builds, would you be up for retesting the capture stall on your side?

Happy to help.

xronocode:
Quick correction after digging deeper into the 0.18 bump: I did run it locally before my last reply, counted 10 compile errors and read them all as mechanical. One wasn't. The real blocker is rodio: the playback path hands a cpal Device to rodio's DeviceSinkBuilder, and the latest rodio (0.22.2, March) still builds against cpal 0.17. No rodio release supports 0.18 yet, it's tracked in RustAudio/rodio#898. A 0.18 bump here would mean either carrying two cpal versions in the tree or dropping rodio from the playback path, and neither belongs in this PR.

So walking my suggestion back: take this one as is on 0.17.3. It fixes the macOS 26 enumeration crash it was opened for, and the 0.18 bump becomes a follow-up once rodio catches up, which is also where the ALSA capture fixes land. I'll watch that issue and do the follow-up bump when it's ready.

@rdslw this does mean your ALSA stall isn't covered by this PR alone, it lands with the later bump. Sorry for the detour, and thanks, your report set the right target version.

xronocode:
Status addendum: the cpal 0.18 support in rodio is already merged on master (a release is pending their audio engine rewrite, tracked in RustAudio/rodio#898). I checked our 0.18 bump against master pinned at 685e057: it compiles apart from the mechanical 0.18 migrations, nothing structural. So the future bump here is de-risked and ready the day rodio ships a release. No change to the suggestion above, 0.17.3 as is still stands.

cjpais:
The macOS minimum version is the main blocker for now and I'm going to see if I can get a change up streamed in cpal

PR #1445, opened 2026-05-28
fix(overlay): stop rendering animations while hidden

koloved:
Stop rendering the overlay React tree while its native window is hidden. This prevents WebView2 from continuously compositing the overlay's CSS animations after recording has stopped.

## Human Written Description

I finally managed to get to the truth, and it seems something came out of it.

## Related Issues/Discussions

Fixes #1418

## Testing

I made the final release build and tested the application many times by repeating its use. No CPU load was observed during idle time.

Maintainer verification after reducing the PR to the minimal fix:

- `bun run lint`
- `bun run build`
- `bunx prettier --check src/overlay/RecordingOverlay.tsx`

## AI Assistance

- Tools used: OpenCode with Deepseek V4 flash free Then checked this code through Qwen-Coder.
- Maintainer cleanup: Pi with OpenAI GPT-5.6 reduced the patch to the minimal hidden-render change.

PR #1645, opened 2026-07-10
fix(windows): refresh overlay bounds after DPI changes

cjpais:
picked the changes from #1445 to separate things out for easier review

@koloved


cjpais:
close in favor of #1733

PR #1662, opened 2026-07-11
fix(models): don't reload the already-active model when its card is clicked

kud:
## Before Submitting This PR

## Human Written Description

I noticed that in **Settings → Models**, clicking the model that's already active kicks off a full reload of that same model. Nothing about the app's state changes — you end up exactly where you started — but you sit through the "Switching…" / "Loading…" spinner for no reason, and on the larger models that's a real wait. It also just feels wrong that the current model looks and behaves like a button when there's nothing to switch to. This makes the active card non-interactive so an accidental click can't trigger a pointless reload.

## Related Issues/Discussions

Fixes #
Discussion:

## Testing

- In **Settings → Models**, the currently-active model card (badged "Active") no longer responds to clicks: no pointer cursor, no hover lift, no `role="button"`, and clicking it does nothing — so it can't re-trigger a reload of the model that's already loaded.
- Switching to a different **downloaded** model (status "available") still selects and loads it as before.
- Clicking a **downloadable** model still starts its download.
- The active card's Delete button and status badge continue to work.
- `bun run lint` passes clean.

**Why the change is a one-liner:** a single `isClickable` flag in `ModelCard.tsx` gates the click handler, the Enter-key handler, `role="button"`, `tabIndex`, the pointer cursor, and the hover styling all at once. It previously included `status === "active"`; dropping that from the flag disables the whole interaction for the active card — the click _and_ every visual cue that it's clickable — while leaving `"available"` and `"downloadable"` untouched.

## AI Assistance

**If AI was used:**

- Tools used: Claude Code
- How extensively: AI located the root cause (the `isClickable` flag including `"active"`) and made the one-line change plus the explanatory comment. The description above is my own; I reviewed and verified the change.

### DIFF
--- a/src/components/onboarding/ModelCard.tsx
+++ b/src/components/onboarding/ModelCard.tsx
@@ -92,8 +92,9 @@ const ModelCard: React.FC<ModelCardProps> = ({
 }) => {
   const { t } = useTranslation();
   const isFeatured = variant === "featured";
-  const isClickable =
-    status === "available" || status === "active" || status === "downloadable";
+  // The active model is already loaded — re-selecting it just reloads it for no
+  // gain, so it is deliberately not clickable.
+  const isClickable = status === "available" || status === "downloadable";
 
   // Get translated model name and description
   const displayName = getTranslatedModelName(model, t);


cjpais:
thanks for this it's simple and makes sense

PR #1666, opened 2026-07-12
feat: add configurable streaming latency for Parakeet and Nemotron models

nub235:
## Before Submitting This PR

## Human Written Description

Currently, the app only lets you use Parakeet Unified and Nemotron models on the highest latency setting, 1120 ms for Nemotron and 2080 ms for Parakeet. I added a slider in each model's settings card so users can choose between four latency options for each model to get their preferred balance of speed and accuracy. I did not expose specific latency in milliseconds in the interface, in order to keep it simple and intuitive. For Nemotron models, it sets the right context value, and for Parakeet Unified, it sets a tuple of left, chunk, and right values, depending on the position of the slider.

## Community Feedback

There is a discussion at [https://github.com/cjpais/Handy/discussions/1654](https://github.com/cjpais/Handy/discussions/1654)

## Testing

I thoroughly tested each of the four latency options for each of the three models, including on long dictations. 

In my testing, I found Parakeet Unified can lag behind on very low latency, so I set the minimum latency for this model to 320 ms. On weak hardware running on CPU only, 320 ms may also lag behind after a while, so I warned about that in the Streaming Latency tooltip. 

The only other issue I found is that for Nemotron Speech Streaming EN, on the second latency setting (160 ms), after around a minute, it sometimes drops sentences, I was not able to figure out why this happens. It's not a super big deal, but I thought it was worth mentioning. 

Screenshot 2026-07-12 at 1 34 28 PM

## AI Assistance

**If AI was used:**

- Tools used: OpenCode
- How extensively: I basically just used it to write the code, everything else was done by me.


nub235:
I tested Parakeet Unified with CPU only, and the first two latency settings still cause it to lag behind, so those only really work on a machine that has good GPU, like a Mac, or high end Windows laptop. Nemotron models still work fine, even at the lowest latency setting, since they use cache aware streaming and therefore do not have to recompute left context every time. It defaults to highest latency setting for these models, so it will not harm the out of the box experience. But it is worth noting, especially if users don't read the tooltip and then think the app is broken when they use lower latency.

cjpais:
I do actually think that this is a good change. I am just a bit skeptical on the UI implementation here. And I do think we should pull this in at some point, but I need to close it just for my own sanity, basically, because I'm not going to be able to get to it in the near term

PR #1670, opened 2026-07-13
feat: allow a separate key to stop recording (opt-in stop shortcut)

shyamkhakharia:
## Before Submitting This PR

## Human Written Description

I found myself struggling to end the transcription the first time I was using this software - I didn't have push-to-talk enabled. It took me time to realise I needed to use the same key bind to end the transcription. Whereas intuitively I would have thought pressing Enter (or similar) would end it and that the key bind is a barrier to only start the recording (you wouldn't want to start it by mistake)

## Related Issues/Discussions

Discussion: https://github.com/cjpais/Handy/discussions/1519 (toggle-mode users are confused about how to *finish* a recording vs cancel it)
Discussion: https://github.com/cjpais/Handy/discussions/818 (request to remap the key that stops the recording)
Discussion: https://github.com/cjpais/Handy/discussions/410 (per-shortcut activation behaviour)

## Community Feedback

Discussions #1519 and #818 both describe the underlying pain point: in toggle mode the only ways to end a recording are pressing the start shortcut again or hitting Escape (which *discards* it, and #1519 shows people find that distinction confusing). This PR lets users opt into a dedicated stop key (default: Enter) that finishes the recording and transcribes it, so start and stop can be different keys.

**What it does**

- Adds an opt-in **"Separate Stop Key"** toggle in Settings → General (hidden in push-to-talk mode and on Linux, same as the cancel shortcut). When enabled, a rebindable **Stop** shortcut appears (default `enter`).
- The stop binding follows the cancel shortcut's existing dynamic-registration pattern: it is **only registered while a recording is active**, so it never swallows the key while idle, and it is **off by default** so existing users see zero behaviour change.
- The key is routed through the `TranscriptionCoordinator`, so it stops whichever transcribe binding started the recording (`transcribe` or `transcribe_with_post_process`) and then transcribes as normal.
- Refactor: the duplicated cancel-specific register/unregister helpers in both keyboard implementations were generalised to `register_dynamic_shortcut(app, binding_id)` and the special-cases now go through a single `is_dynamic_binding()` helper.
- Related hardening: `suspend_binding`/`resume_binding` now skip dynamically-managed bindings. Previously, starting to edit the cancel shortcut in the UI and clicking away could re-register Escape globally while idle; the same would have applied to Enter for the stop key, which would be much worse.

## Testing

- `cargo test --lib`: 122 passed, 0 failed (macOS, Apple Silicon)
- `cargo check --tests` clean (one pre-existing warning in `managers/transcription.rs`, untouched by this PR)
- `cargo fmt --check`, `bun run lint`, `tsc --noEmit`, `bunx prettier --check`, `bun run check:translations` (all 21 non-English locales complete) all pass
- New i18n keys were added to all 22 locales; translations for the non-English locales are best-effort and happy to be corrected by native speakers
- Not yet manually tested end-to-end on a running build (no local model setup on this machine) — happy to do a follow-up pass or take testing help, especially on Windows and with the handy_keys implementation

None yet — the UI addition is a toggle + shortcut row in Settings → General directly under the Cancel shortcut, following the existing components.

## AI Assistance

**If AI was used:**

- How extensively: The implementation was written by Claude Code following the repo's existing patterns (cancel shortcut dynamic registration, settings command plumbing, i18n), directed and reviewed by me. Non-English translations are AI-generated.


cjpais:
Features are not really being accepted at the moment...

PR #1700, opened 2026-07-16
fix(overlay): restore Wayland overlay on Tauri 2.11.5 (Linux)

### LINKED ISSUE #1696: [v0.9.3 regression] Recording overlay invisible on KDE Wayland (KWin) — GTK layer-shell broken after tauri fork → upstream switch
## Bug Description

After upgrading from **v0.9.2 → v0.9.3** on **Fedora 44 KDE Plasma (Wayland, KWin)**, the recording overlay is **no longer visible**. Transcription itself works fine — only the overlay window disappears.

This is a **regression introduced in v0.9.3**. Downgrading back to v0.9.2 restores the overlay immediately (verified).

## Root Cause (investigated)

- v0.9.2 shipped via a **`cjpais/tauri` fork** (branch `handy-2.10.2`) with custom `tauri-runtime`, `tauri-runtime-wry`, and `tauri-utils` that carried **Wayland-specific patches for GTK layer-shell** integration.
- v0.9.3 (PR #1675) **replaced the fork** with **upstream `tauri` 2.11.5** + a pinned `tao` (commit `07f3742b1`, only to fix Wayland CSD).
- The overlay **is created successfully** and `show_overlay_state()` runs to completion (logs confirm `set_size` / `pos_calc` / `set_pos` / `show`). GTK layer-shell **initializes**. But the **KWin compositor never renders the layer surface** — the window is logically shown but not painted on screen.

### Evidence

With `RUST_LOG=debug /usr/bin/handy`:

```
[handy_app_lib::overlay][DEBUG] Recording overlay window created successfully (hidden)
[handy_app_lib::overlay][DEBUG] GTK layer shell initialized for overlay window
[handy_app_lib::overlay][DEBUG] overlay 'recording': set_size=... pos_calc=... set_pos=... show=...
```

Settings confirm `overlay_style: Live`, `overlay_position: bottom`, `start_hidden: true`.

### Workaround that proves the cause

Running with the env var that **skips GTK layer-shell** makes the overlay appear:

```bash
HANDY_NO_GTK_LAYER_SHELL=true /usr/bin/handy
```

→ Overlay **becomes visible** (as a regular window), but **centered**, because on Wayland the compositor blocks client-side window positioning. This confirms the layer-shell surface is the piece that broke between the fork and upstream tauri 2.11.5.

## Reproduction

1. Install `Handy-0.9.3-1.x86_64.rpm` on **Fedora 44 KDE Plasma, Wayland session**.
2. Set `overlay_style: Live` (default), `overlay_position: bottom`.
3. Start a recording (e.g. push-to-talk).
4. **Expected:** overlay shows at the bottom of the screen.
5. **Actual:** no overlay is visible; transcription still works.
6. Set `HANDY_NO_GTK_LAYER_SHELL=true` and repeat → overlay appears (centered).

## System Information

- **App Version:** 0.9.3 (regression vs 0.9.2)
- **Operating System:** Fedora 44 KDE Plasma 6
- **Session:** Wayland (KWin)
- **Architecture:** x86_64 (rpm package)

## Additional notes

- The `cjpais/tauri` fork appears to have **no 2.11.x branch**, so a clean revert via Cargo.toml is not currently possible without porting the Wayland layer-shell patches to upstream tauri 2.11.
- On v0.9.2 the overlay works correctly (positioning + visibility), with only minor/occasional UI lag — that version is the last known-good for KDE Wayland overlay.

xilec:
## Before Submitting This PR

## Human Written Description

The Tauri 2.11.5 update in commit 5a7c0eac1ac250142b9fe58a898cbf79cf9f1814 broke
overlay rendering on my system, and based on issue #1696 it is also broken on some
other systems. This PR narrowly fixes the overlay for Linux while preserving the
overlay behavior on other platforms.

## Related Issues/Discussions

Fixes #1696

## Community Feedback

Bug fix confirmed by the reporter in #1696 (KDE/KWin). Verified independently on
niri (another Wayland compositor) where the overlay was similarly invisible.

## Testing

- Platform: Linux (NixOS, Wayland, niri).
- Build: `cd src-tauri && cargo build` on Tauri 2.11.5 (pinned `[patch.crates-io]`
  tao/tao-macros to rev `07f3742b1833b64be27b1ef991e38d557d4276c9`).
- Manual verification on niri:
  - Overlay appears on the **first** transcription (previously invisible).
  - Overlay grows correctly with streamed text.
  - Runtime Top↔Bottom position switch works; overlay does not disappear.
  - Repeated show/hide cycles keep working.
- Non-Linux platforms (Windows/macOS) are unchanged: all new code is under
  `#[cfg(target_os = "linux")]`; shared code was only moved into
  `#[cfg(not(target_os = "linux"))]` with no logic change.

## AI Assistance

**If AI was used:**

- Tools used: opencode (AI coding assistant, model hy3-free, Kimi K2.7, reviewer Kimi K3)
- How extensively: AI assisted with root-cause analysis, the GTK-surface
  helper refactor, and code review. The final fix and its behavior on
  Wayland were verified manually by the human contributor.


cjpais:
Just to confirm this did work probably on your system?

xilec:
Yes, the fix works on my system. I tested it as described in the PR's testing section, and I've been using Handy from this PR branch daily. I've also added a few small improvements based on further review.

cjpais:
@xilec one quick question did you ever encounter the bug in #1806

xilec:
@cjpais No — the main window has rendered correctly for me on NixOS/niri across all builds after the Tauri 2.11.5 switch. Left a note in #1806: https://github.com/cjpais/Handy/issues/1806#issuecomment-5145658548

xilec:
Hi @cjpais — just checking in on this one. The branch is now up to date with `main` (including the main-thread overlay dispatch from #1810), and I've re-verified the overlay locally on Wayland/niri: recording and streaming modes show, reposition (top/bottom), and hide correctly.

Is there anything currently blocking this? For example:

- just a matter of finding time for review,
- you'd like additional testing from other Linux contributors,
- you're waiting for the transcribe-cpp 0.2.0 merge to land first,
- or you'd like a more detailed description of how the new version works — here in the PR or somewhere else?

Happy to do whatever makes this easier to land.

buz-ch:
FWIW, I built this on Opensuse Tumbleweed and it fixes the issue on KDE Wayland. Been running it since yesterday, it seems perfectly stable.

Thanks a lot!

cjpais:
I'm doing a review over this and there's possibly some changes I want to make. I'm just like a bit concerned about the changes as written

I think the feedback from the issue itself might be worth taking into account, like the other user who offered his perspective on implementation, it may be able to simplify this implementation and maybe be a bit more robust

It's just hard for me to review these Linux things unless I have time to really sit down in front of a Linux machine and make sure that it's solid basically if you could just take a look at the feedback from that other user, I would be grateful to hear some perspective

xilec:
Took a look at @arimxyer's feedback as you asked — and went a bit further: I tested a cherry-pick of  [a7dea9c8](https://github.com/arimxyer/Handy/commit/a7dea9c8) onto current `main` on niri.

Scenarios covered: first show, streaming growth (256x46 → 400x120 on a mapped surface), runtime Top↔Bottom switch including mid-recording, and repeated show/hide cycles. Everything works, so his `set_size_request` + layer-shell margin approach is now verified on two compositors (his KWin + niri here).

His analysis is correct, and it also names the problem this PR was working around: the fixed max-size window and "never resize a mapped layer surface" were my way around the same 0x0 surface he diagnosed properly.

On the extra mechanics in this PR (surface priming, remap on position change, sync before the `show-overlay` emit): I added those in July against failures I was hitting on niri at the time. In this round of testing the same scenarios worked without them — I suppose because #1810 moved all overlay geometry onto the main thread. So I agree the simpler implementation should be sufficient.

One caveat worth flagging from all this testing: there's a pre-existing glitch where the streaming card leaves faded pixels behind the compact overlay — WebKit doesn't clear the now-transparent region on state change. It reproduces on current `main`, on this PR, and on arimxyer's build alike, so it's orthogonal to the sizing approach. I'll investigate it separately.   
 
orca-paste-1786029200011-1f4d7e75-60f9-4e1a-b13c-42fa2698b266

arimxyer:
Happy to also contribute to the investigation! @xilec  and @cjpais . Also happy to send in a PR if it provides any relief, I originally withheld and just focused on leaving the comment since the PR was already created.

cjpais:
okay @arimxyer @xilec I made the changes directly here, can you please test them and let me know if we are good to merge this based on your testing?

ctcac00:
I'm also on niri and I've just built this locally and the overlay is working again for me

xilec:
@cjpais Tested on niri (multi-monitor Wayland). Almost everything works: first show, streaming growth 256x46 → 400x120 via `set_size_request` + `resize(1, 1)`, runtime Top↔Bottom switch including mid-recording, correct 40/4px margins, repeated show/hide cycles.

One issue: the cursor-based `set_monitor` block misplaces the overlay on niri. Handy builds enigo with default features (the x11rb backend), so `location()` reads the pointer through XWayland — which only tracks the pointer while it's over an X11 window. Over native Wayland windows the coordinates are stale: I kept getting the identical point across app restarts and mouse movement, and `monitor_at_point` then picked the wrong output (the overlay landed on the monitor with Firefox, an X11 window, instead of the one holding the text-input focus).

I don't see a reliable client-side way to pick the monitor on Wayland: the global pointer position and the keyboard focus are deliberately not exposed to clients, and enigo's own Wayland backend has `location()` [unimplemented](https://github.com/enigo-rs/enigo/blob/main/src/linux/wayland.rs#L1033-L1039) for that reason. The alternatives are all dead ends — foreign-toplevel protocols say *which* app is active but not on *which* output, and compositor IPC (niri/sway/hyprland) would need per-compositor adapters.

With the `set_monitor` block removed, the output is left to the compositor — and per the [layer-shell protocol](https://wayland.app/protocols/wlr-layer-shell-unstable-v1#zwlr_layer_surface_v1.request.set_output), when no output is set the compositor maps the surface on the output the user most recently interacted with. Verified on niri: the overlay then appears on the monitor holding the current text-input focus, which is exactly where a dictation overlay belongs. So this looks like the intended mechanism rather than a fallback.

I'd suggest dropping the `set_monitor` block (keeping `set_size_request` + `resize(1, 1)` and the anchors + margins). Happy to push that change here — or feel free to adjust directly, whatever is easier for you.

cjpais:
@xilec ive applied the fix, can you give it a build and test on your machine please?

buz-ch:
Just built it, works on KDE/Wayland.

Can test multi monitor on Saturday.

xilec:
@cjpais Built and tested on commit 709fb68 on niri (multi-monitor Wayland) — everything works:

- the overlay now lands on the monitor holding the text-input focus (the case that was broken with `set_monitor`)
- first show, streaming growth 256x46 → 400x120, runtime Top↔Bottom switch including mid-recording, repeated show/hide cycles — all good

Good to merge from my side. Thanks for the quick turnaround!

cjpais:
Thanks yall for the confirmation, merging

PR #381, opened 2025-11-25
Added feature of transcription of local files (WAV, MP3 and M4A) along with progressbar

Signal46:
Hello! This is my first contribution to an open-source project on GitHub. I’m a big fan of Handy and saw a need for a feature to handle long-form recordings safely.

I realized after working on this that there is an existing PR (https://github.com/cjpais/Handy/pull/371) for file uploads, but I thought I would make this contribution anyway since this has a progress bar.

**Motivation & Context**
There is a significant need for this feature in the public sector, specifically for municipalities and government agencies (e.g., here in Sweden).

Compliance & GDPR: Many organizations cannot use cloud-based transcription (which often costs ~$150/user/month) due to strict data compliance laws regarding sensitive meetings.

Local Processing: By keeping the file and transcription 100% local, Handy solves major legal hurdles regarding data transfer.

Efficiency: Estimates in a small Swedish municipality suggest transcription (including sensitive meetings) could save around 5000 hours administrative hours annually. Yes, 5000 hours, not 500... for a municipality of around 60,000 citizens.

**Technical Details**
Implementation: This feature was built with the assistance of AI coding agents but has been manually reviewed and tested.

Testing: Validated on a standard work laptop using the Parakeet V3 model. Successfully transcribed 20 and 40-minute files without memory spikes or crashes.

I am happy to make changes or discuss how this might be merged or combined with existing efforts!

_________________________________________________________________________________________________________________________________

### Summary
Adds the ability to import existing audio files (MP3, M4A, WAV) into Handy's history with automatic transcription. This feature includes real-time progress tracking, system notifications, and robust handling of long audio files using smart chunking (VAD).

Successfully transcribed files are automatically moved to the "Recordings" folder for organization.

Motivation
Users requested the ability to transcribe existing audio files, not just live recordings. This feature enables batch processing of pre-recorded audio while maintaining the same quality and privacy guarantees as live transcription.

Changes
Core Features
File Import: Native file picker for selecting MP3, M4A, and WAV files.
Smart Chunking: Uses Voice Activity Detection (VAD) to split audio on silence, preventing words from being cut in half and improving transcription quality.
Progress Tracking: Real-time progress bar showing transcription progress (0-100%).
System Notifications: Desktop notifications on completion or failure.
Long File Support: Prevents crashes on files >20 minutes by processing in chunks.
Auto-Model Loading: Automatically loads the transcription model if it's not currently loaded.
File Management: Moves imported files to the 
recordings
 folder after successful transcription.
Technical Implementation
[...]

PR #1704, opened 2026-07-17
feat: transcribe local audio files through the existing streaming path

### LINKED ISSUE #154: [FR] file transcription?
I have some files I need to transcribe, and I’d prefer to keep using only this tool instead of switching to others—unless it’s too much work.
Another thing: I can’t use the Fn shortcut on my MacBook. Also, Parakeet v3 sometimes transcribes in languages other than English, so I’d love to be able to lock it to English only.

### LINKED ISSUE #1494: Feature Request: Audio file transcription (drag & drop / file picker)
## Summary

Handy is excellent at real-time speech-to-text, but it would be incredibly useful to also support **transcribing audio files** (WAV, MP3, M4A, OGG, etc.) — not just live microphone input.

## Use Case

I often have audio recordings (voice memos, meeting recordings, podcast clips, interview snippets) that I'd like to transcribe locally with the same quality Whisper/Parakeet models provide in Handy. Currently I have to use cloud services or separate tools for this, even though Handy already has all the transcription infrastructure built in.

## Proposed UX

- A new **"Transcribe File"** tab/section in the sidebar (alongside existing settings)
- **Drag & drop** an audio file onto the window, or use a **file picker** button
- Show transcription progress (model loading, processing percentage)
- Display the resulting text with a **copy to clipboard** button
- Optionally save to transcription history (same as mic recordings)

## Why This Fits Handy Well

The backend already has most of the infrastructure:
- `read_wav_samples()` in the audio toolkit for reading WAV files
- `TranscriptionManager.transcribe()` for running inference
- `retry_history_entry_transcription` command that re-transcribes stored recordings
- `rubato` for resampling to 16kHz
- Tauri's `dialog` plugin is already available

What's missing is mainly:
1. A Tauri command that accepts an arbitrary file path and runs it through the transcription pipeline
2. Multi-format decoding (MP3/M4A/OGG — could use `symphonia` crate)
3. A React UI component for file selection and result display

## Benefits

- **Privacy**: Same offline, local-only processing — no cloud upload needed
- **Consistency**: Same Whisper/Parakeet models users already trust for live transcription
- **Low effort**: Most of the backend pipeline already exists
- **Natural extension**: Turns Handy from "speech-to-text input tool" into "complete local transcription toolkit"

## Alternatives Considered

- Using Whisper CLI separately — works but no GUI, no model management, separate tool
- Cloud transcription services — defeats Handy's privacy-first philosophy
- Other desktop apps — none combine the model quality + local processing + clean UX that Handy has

Would love to hear if this is something on the roadmap or if a PR would be welcome! Happy to help with implementation.

bygriffin:
## Before Submitting This PR

## Human Written Description

I produce online courses and transcribe hour-long call recordings every week, and I didn't want to ship them to a cloud service. So I tried to make Handy do it. My first attempt just decoded the file and handed the whole buffer to the engine. An 85 minute mp3 ate 86 GB of RAM and froze my Mac. Turned out the right way is to feed the existing streaming session frame by frame, the same way the mic path already does. That's what this PR is.

## What it does

Adds a "File" tab in settings: pick an audio or video file, the currently selected model transcribes it, you copy the text out. It reuses the streaming path the live mic preview already uses, no new engine code.

Why not just call `transcribe()` with the decoded file? Attention memory in the conformer models is quadratic in input length. An 85 minute mp3 decodes to 81.8M samples at 16 kHz, and passing that as one tensor allocated 86 GB on my machine before macOS gave up.

So instead:

- transcribe-cpp models that support streaming (whisper family, Nemotron): a new `TranscriptionManager::transcribe_buffer_streaming()` drives `session.stream()` with 512-sample frames, same cadence as the recorder. Memory stays flat no matter how long the file is. Feeding is synchronous, so a slow engine gets backpressure instead of a growing frame queue.
- models that can't stream (Parakeet and friends): fall back to 30 second windows through the existing batch `transcribe()`, so no single call ever sees the whole file.
- decoding is ffmpeg to 16 kHz mono wav, then the existing `read_wav_samples`. ffmpeg is probed at absolute paths first because GUI apps launched from Finder don't get the Homebrew PATH.

The concurrency part took the most iterations. The file run claims the same worker/lease tokens as the live mic stream worker, in the same order, with checked CAS on both, so hitting the dictation hotkey mid-file gets you a "busy" error instead of corrupted state. Model loading goes through a new `ensure_model_loaded()` serialized on the existing `is_loading` slot, so a hotkey press while the file's model is still loading can't start a second multi-GB load. Engine calls are wrapped in `catch_unwind` like `transcribe()` already does. `ModelUnloadTimeout::Immediately` is honored: the chunked fallback reloads per chunk, the streaming path unloads once after the run.

The UI is one sidebar tab, about 230 lines: file picker, a progress bar with an ETA fed by a `transcribe-file-progress` event, result in a textarea with a copy button. i18n keys are added for all 23 locales. The result survives switching tabs.

Numbers from a real run (85 min / 79 MB mp3, Apple Silicon, Nemotron Streaming 0.6B GGUF): peak RSS 1.5 GB and flat for the whole run, 6.9 minutes wall clock (12.3x real time), 144k characters out. `cargo check`, `tsc`, `bun run check:translations` and prettier/cargo fmt are all clean.

What I left out on purpose: there is no cancel button yet (the job runs to completion or errors, the UI stays responsive throughout), and if a streaming model fails to open a stream the run quietly falls back to the chunked path, which is the same recovery the mic path uses. ffmpeg stays an external dependency for now. symphonia is already in the tree via rodio and could replace it later, but ffmpeg swallows every container people will actually throw at this.

## Related Issues/Discussions

Fixes #154, where @cjpais wrote: "I think this would be a good thing to have in the future. Would love to have a PR implement it in a clean way."
Fixes #1494 (the same request, a year later)

On prior attempts: #371 was closed with "not sure what the best UI for it is". My answer to that is one settings tab and nothing else. #381 is open and builds a whole Studio batch workflow, about 4000 lines. This one is about 450 lines of code on top of the existing streaming engine. If the Studio direction wins, the memory-safe core here (`transcribe_buffer_streaming`, the chunked fallback, the engine lease discipline) is the part it needs anyway, and I'm happy to strip this down to a backend-only PR.

## Community Feedback

#154 has 11 comments of people asking for this or working around it by hand (one user cuts recordings into 2-3 minute chunks in Audacity and feeds them through the mic path). #1494 is the same ask a year later.


cjpais:
This has been submitted multiple times and the thing blocking it is not the backend, it's having good UI

You didn't include any screenshots so...

cjpais:
No response from the author, so I'm closing this for now because there are other PR's doing this and we don't really need another

PR #1711, opened 2026-07-18
feat: add ${custom_words} placeholder to post-processing prompts

djurcola:
## Summary

- add `${custom_words}` placeholder support to post-processing prompt templates
- render custom words consistently for structured-output, Apple Intelligence, and legacy provider paths
- use an explicit `(none provided)` fallback when the custom-word list is empty
- preserve placeholder-like text inside transcription and custom-word values without recursive expansion
- document `${output}` and `${custom_words}` in the English prompt UI

## Motivation

Custom words already help transcription correction, but post-processing prompts cannot currently reference the configured list. This lets users include the same domain-specific vocabulary in their LLM instructions without duplicating it manually.

## Testing

- `cargo fmt --check --manifest-path src-tauri/Cargo.toml` — passed
- focused pure-Rust renderer harness — 5 passed
- `npm run lint` — passed
- `npm run build` — passed
- Prettier check for `src/i18n/locales/en/translation.json` — passed
- JSON parse validation — passed
- `git diff --check` — passed

The full Cargo test target could not compile on the local Linux host because `pkg-config`/GLib development tooling is unavailable; it failed while building `glib-sys`, before compiling Handy. The focused renderer was therefore compiled and exercised independently with `rustc --test`.

## Compatibility

- existing `${output}` behavior is preserved
- templates without either placeholder are unchanged
- existing fuzzy custom-word transcription correction is untouched
- no settings schema, generated bindings, CI, or dependency changes

## AI assistance disclosure

Yes. OpenCode using the Codex-backed GPT-5.6 Sol model assisted with adapting the existing fork implementation to current upstream, and Hermes Coder reviewed the diff and ran verification. The final changes were constrained to the two feature-relevant files.


cjpais:
This is probably a good change, I will try and merge it when I get a chance

djurcola:
I am running it locally and it makes post processing almost perfect, would be great to get it into the main branch!

Issue #1228, opened 2026-04-05
[BUG] sometimes trascribing is stuck forever(?)


sakhdevel:
## Bug Description

On Linux, the transcribing process is never ending sometimes. I don't know why it happens, but sometimes I it looks like it happens on the amount of text that being transcribed. I mean, if I talk too much, of course the transc transcription process will get longer, but sometimes it is just stuck. So, in this situation, I can use cancel from the menu from system tray icon. And in this case, the push to talk button doesn't work. Anyway, I need to close the application entirely and open it again. 

## System Information

**App Version:**

0.8.2 (but i noticed the same at 0.8.1 and may be earlier)

**Operating System:**

Ubuntu 24.04.4 LTS

**CPU:**

Intel Core Ultra 5 125H

**GPU:**

Intel Arc Graphics

## Logs

[handy.log.zip](https://github.com/user-attachments/files/26487903/handy.log.zip)

cjpais:
Possible duplicate of #1213? Is this a laptop? Also curious if this ever happened on 0.7.x. Do you have a laptop by chance? Or just generally put the computer to sleep?

I believe we made a change in 0.8.0 which might have caused this regression. We didn't get any reports previously

Avatarsia:
> Possible duplicate of [#1213](https://github.com/cjpais/Handy/issues/1213)? Is this a laptop? Also curious if this ever happened on 0.7.x. Do you have a laptop by chance? Or just generally put the computer to sleep?
> 
> I believe we made a change in 0.8.0 which might have caused this regression. We didn't get any reports previously

it happened also on 0.7.x in windows

sakhdevel:
@cjpais I'm using laptop. I can confirm that it still happening sometimes at 0.8.2. In my case it does not relate to sleep mode. It can work well, and then (without going to sleep mode) seldom stuck. It seems it does not depend on how long the record is. I thought before that it stuck only for too long records. But for small it happens too. Don't know what addittional info I can provide you with

cjpais:
@sakhdevel thanks for now this is good info

sakhdevel:
Confirm that this sometimes happening in 0.8.3. 

May be could at least show some progress (in percent) in this small window? so that we know that it is not stuck
<img width="563" height="183" alt="Image" src="https://github.com/user-attachments/assets/15fbfc89-7dbb-4e32-970a-03c48714f584" />

sakhdevel:
Hi guys!

[...]

PR #1713, opened 2026-07-18
fix: transcription pipeline hangs forever with audio_feedback enabled (WASAPI stall + unbounded recorder waits)

### LINKED ISSUE #1712: [BUG] Transcription pipeline hangs forever with audio_feedback=true — per-chime WASAPI stream stall + mic worker never sees Stop when capture callbacks die (fixes field-tested)
## Bug Description

On Windows, with **Audio Feedback enabled**, the transcription pipeline intermittently hangs forever after the stop hotkey: the `Transcribing...` overlay never resolves, Cancel does not help, further presses log `Ignoring press for 'transcribe': pipeline busy`, and only restarting Handy recovers. Before root-causing this I hit the hang every 2–30 transcriptions; disabling Audio Feedback was the only reliable workaround. Reproduced on v0.8.3 and the relevant code is unchanged in v0.9.2.

After building from source, instrumenting, and field-testing fixes for several days, this turns out to be a **chain of two independent defects**:

### Defect A — audio_feedback opens a fresh WASAPI output stream on every chime

`src-tauri/src/audio_feedback.rs` (`play_audio_file`) calls `OutputStreamBuilder::from_default_device()` — or, for an explicit device, `host.output_devices()` enumeration — on **every** feedback sound, i.e. twice per transcription, concurrently with other WASAPI session activity (music playing, the app's own mic stream being stopped on the same hot path). This per-chime stream creation is fragile: after a device state change (observed with a wireless headset; likely also sleep/wake, cf. #1213) the stream can stall with a **frozen audio clock and no error** — `sleep_until_end()` then never returns and the playback thread is wedged inside the audio stack forever.

Chime playback is already fire-and-forget on the stop path (a detached thread), so defect A alone does not hang the pipeline — which is why the hang is intermittent and survived the async refactor. The pipeline dies because of:

### Defect B — the mic worker services commands only while capture callbacks deliver audio

In `src-tauri/src/audio_toolkit/audio/recorder.rs` the worker's main loop is:

```rust
    // commands (Cmd::Start / Cmd::Stop / Cmd::Shutdown) are polled
    // only after a chunk arrives
```

and both `stop()` (`resp_rx.recv()`) and `close()` (`worker_handle.join()`) wait **unbounded**.

When the audio engine is disturbed by the wedged render stream from defect A, the capture callback stops delivering chunks. The worker then sleeps in `sample_rx.recv()` forever, `Cmd::Stop` is never read, the async transcription task blocks inside `stop_recording`, and the pipeline stays `busy` until the process is killed. Any other condition that silences the capture callback (dead device, engine stall — cf. #1228, #1349 on PipeWire) reaches the same dead end through defect B, with or without audio feedback.

This chain is directly visible in the logs below: on a build where chime playback already had a bounded timeout, the wedged chime thread produced 90 seconds of contained "audio feedback did not finish" warnings (six transcriptions kept working, silently), after which a stop request hung exactly at the mic-stream stop and never returned.

## Fixes (implemented and field-tested)

I have three fixes running on my machine as a patched v0.9.2 build. Happy to submit any/all as PRs.

1. **Persistent chime output stream** (defect A): a dedicated long-lived playback worker opens the output stream once — pre-warmed at startup, while no transcription is running — and reuses it; it is recreated only on device change or playback error. Blocking callers (start-chime → mute sequencing, test sound) wait with a bounded timeout. Per-transcription WASAPI stream creation is gone entirely.
2. **Bounded waits in the recorder** (defect B): the worker's main loop uses `recv_timeout(250ms)` so commands are serviced even when the capture callback delivers nothing; `stop()` waits at most 5 s (losing one utterance is recoverable, a forever-busy pipeline is not); `close()` joins via a reaper thread with a 3 s bound. Streams suspected of being wedged are dropped on throwaway threads, never on the worker.
3. **Default-device following**: since the persistent stream no longer re-resolves the default on every play, a "Default" stream re-checks what the OS default resolves to before each use (a cheap device-name query, not a stream creation) and recreates itself when the user switches headphones ↔ speakers.

**Field validation:** before the fixes the hang occurred every 2–30 transcriptions with Audio Feedback on. With the fixes: several days of daily use (hundreds of transcriptions, GigaAM v3 ONNX and Whisper Large v3 on the Vulkan backend), zero hangs. Notably, one real audio-stack stall did occur during testing — it degraded to a few missing chimes and a WARN in the log, exactly as designed, and recovered without restart. Worst-case degradation is now one lost chime or one lost utterance, never a stuck app.

## System Information

**App Version:** v0.9.2 (originally reproduced on v0.8.3; the affected code paths are identical)

**Operating System:** Windows 10 Home 10.0.19045

**CPU:** AMD Ryzen 7 9800X3D

**GPU:** NVIDIA RTX 5070

**Audio:** USB fifine microphone (capture); ROG Theta wireless headset / speakers (render, both as OS default and as explicit selection — the hang reproduced in both configurations, explicit device only reduced the frequency)

## Logs

Original hang, stock v0.8.3 (`handy.log`, sanitized). The async task reaches the chime WAV probe and then produces no further entries — no `Microphone stream stopped`, no inference, nothing:

```
[18:59:15][handy_app_lib::actions][DEBUG] TranscribeAction::stop called for binding: transcribe
[18:59:15][handy_app_lib::actions][DEBUG] TranscribeAction::stop completed in 6.1668ms
[18:59:15][handy_app_lib::actions][DEBUG] Starting async transcription task for binding: transcribe
[18:59:16][symphonia_core::probe][DEBUG] found the format marker [52, 49, 46, 46] @ 0+2 bytes.
[18:59:22][handy_app_lib::shortcut::handy_keys][DEBUG] handy-keys event: binding=transcribe, state=Pressed
[18:59:22][handy_app_lib::transcription_coordinator][DEBUG] Ignoring press for 'transcribe': pipeline busy
```

The two-defect chain, captured on a build where chime playback already had a bounded blocking timeout (fix 1 partially applied). The wedged chime thread is contained for ~90 s — six transcriptions complete normally, only the sound is missing — then a stop request hangs at the mic-stream stop (defect B) and never completes:

```
[04:57:25][handy_app_lib::audio_feedback][WARN] Audio feedback did not finish within 3s; continuing without it
[04:57:28][handy_app_lib::managers::transcription][INFO] Transcription completed in 0.04s ...   <- pipeline still fine
[04:57:37][handy_app_lib::audio_feedback][WARN] Audio feedback did not finish within 3s; continuing without it
[04:57:38][handy_app_lib::managers::transcription][INFO] Transcription completed in 0.06s ...   <- still fine
[04:58:49][handy_app_lib::audio_feedback][WARN] Audio feedback did not finish within 3s; continuing without it
[04:58:54][handy_app_lib::actions][DEBUG] TranscribeAction::stop called for binding: transcribe
[04:58:54][handy_app_lib::actions][DEBUG] Starting async transcription task for binding: transcribe
                                          <- nothing ever follows: no "Recording stopped and samples retrieved"
[04:59:26][handy_app_lib::utils][INFO] Initiating operation cancellation...
[04:59:26][handy_app_lib::managers::audio][DEBUG] Cancellation requested while recording is stopping
[04:59:30][handy_app_lib::transcription_coordinator][DEBUG] Ignoring press for 'transcribe': pipeline busy
```

## Related issues

- #1213 — Windows 11 stuck on "transcribing" after sleep (sleep/wake is a device-state trigger for the same chain)
- #1228 — transcribing stuck forever (defect B dead end)
- #1349 — PipeWire lockup on Ubuntu (consistent with the same per-chime stream-creation fragility on the cpal/PipeWire side)
- #1415 (closed), #1199 (closed) — same symptom

If useful, this issue can serve as the consolidated root-cause report for the family above.

---

*Disclosure: the analysis and patches were developed with AI assistance (Claude Code) and were built, instrumented and field-validated by me on the machine described above.*

kydechuk01:
## Before Submitting This PR

**If this is a feature or change that was previously closed/rejected:** N/A — bug fix requested by the maintainer in #1712.

## Human Written Description

> Thanks for a great product! But on my setup transcription kept hanging all the time, so I ended up making a few patches with Claude — it really helped, and I want to share.

*(author's own words, translated from Russian)*

## Related Issues/Discussions

Fixes #1712
Likely the same underlying family: #1213, #1228, #1349 (analysis in #1712)

## Community Feedback

The maintainer asked for this patch in #1712: "If you have a patch would love to see it and pull it in". #1213/#1228/#1349 contain independent confirmations of the symptom.

## Testing

Field-tested on Windows 10 (Ryzen 7 9800X3D, RTX 5070, USB fifine mic, wireless headset + speakers):

- **Before:** with Audio Feedback enabled the pipeline hung every 2–30 transcriptions (stuck on "Transcribing…", restart required).
- **After:** several days of daily use — hundreds of transcriptions, GigaAM v3 (ONNX) and Whisper Large v3 (Vulkan) — zero hangs.
- One real audio-stack stall did occur mid-test: it degraded to a few missing chimes with `WARN` log lines and recovered on its own, which is exactly the designed worst case (one lost chime / one lost utterance, never a stuck app).
- Verified the "Default" output selection now follows a headphones ↔ speakers default-device switch at runtime.
- `cargo clippy` clean on the touched files, `cargo fmt` applied.

Commit-by-commit rationale and full log forensics are in #1712. The three commits are independent and can be cherry-picked separately if you want only part of this.

I'd appreciate additional testing on macOS/Linux — the recorder changes (bounded waits, command tick) are cross-platform, while the WASAPI-specific observations are Windows-only.

N/A — sanitized log excerpts are in #1712.

## AI Assistance

**If AI was used:**

- Tools used: Claude Code
- How extensively: root-cause analysis (code reading + log forensics) and the patch code were written with Claude Code under my direction and review; I built the patched app and validated it on my machine over several days of daily use. Commits carry a `Co-Authored-By: Claude` trailer.

### DIFF
--- a/src-tauri/src/audio_feedback.rs
+++ b/src-tauri/src/audio_feedback.rs
@@ -141,6 +141,11 @@ fn wait_for_play(settings: &AppSettings, path: PathBuf) {
     }
 }
 
+/// If a chime hasn't finished within this bound, its output stream is
+/// considered a zombie (frozen audio clock after a device state change —
+/// observed with wireless headsets) and gets scrapped.
+const PLAYBACK_STALL_TIMEOUT: Duration = Duration::from_secs(5);
+
 fn playback_worker(rx: mpsc::Receiver<Request>) {
     // The output stream is created once and kept open across transcriptions.
     // Recreating it per chime is what raced concurrent WASAPI session state
@@ -161,11 +166,11 @@ fn playback_worker(rx: mpsc::Receiver<Request>) {
                 if let Some((_, stream)) = ensure_stream(&mut cached, device) {
                     if let Err(e) = play_on_stream(stream, &path, volume) {
                         error!(
-                            "Failed to play sound '{}': {}; dropping cached output stream",
+                            "Failed to play sound '{}': {}; scrapping output stream",
                             path.display(),
                             e
                         );
-                        cached = None;
+                        scrap_stream(&mut cached);
                     }
                 }
                 if let Some(done) = done {
@@ -176,14 +181,22 @@ fn playback_worker(rx: mpsc::Receiver<Request>) {
     }
 }
 
+/// Dispose of the cached stream on a throwaway thread. Dropping a cpal
+/// stream whose device wedged can block indefinitely — that must park a
+/// disposable thread, never this worker.
+fn scrap_stream(cached: &mut Option<(Option<String>, rodio::OutputStream)>) {
+    if let Some((_, stream)) = cached.take() {
+        thread::spawn(move || drop(stream));
+    }
+}
+
 fn ensure_stream(
     cached: &mut Option<(Option<String>, rodio::OutputStream)>,
     device: Option<String>,
 ) -> Option<&(Option<String>, rodio::OutputStream)> {
     let stale = cached.as_ref().map(|(d, _)| d != &device).unwrap_or(true);
     if stale {
-        // Drop any previous stream before opening the replacement.
-        *cached = None;
+        scrap_stream(cached);
         match create_stream(device.as_deref()) {
             Ok(stream) => *cached = Some((device, stream)),
             Err(e) => error!("Failed to open audio feedback output stream: {}", e),
@@ -230,6 +243,20 @@ fn play_on_stream(
     let file = File::open(path)?;
     let sink = rodio::play(stream.mixer(), BufReader::new(file))?;
     sink.set_volume(volume);
-    sink.sleep_until_end();
+    // Bounded wait instead of sleep_until_end(): a stream whose device
+    // changed state underneath it stops consuming samples without erroring,
+    // and an unbounded wait would wedge the worker on it forever.
+    let started = std::time::Instant::now();
+    while !sink.empty() {
+        if started.elapsed() > PLAYBACK_STALL_TIMEOUT {
+            sink.stop();
+            return Err(format!(
+                "playback did not finish within {:?} (zombie output stream?)",
+                PLAYBACK_STALL_TIMEOUT
+            )
+            .into());
+        }
+        thread::sleep(Duration::from_millis(25));
+    }
     Ok(())
 }
--- a/src-tauri/src/audio_feedback.rs
+++ b/src-tauri/src/audio_feedback.rs
@@ -55,9 +55,7 @@ fn get_sound_base_dir(settings: &AppSettings) -> tauri::path::BaseDirectory {
 
 enum Request {
     /// Ensure the output stream for `device` exists (startup pre-warm).
-    Warm {
-        device: Option<String>,
-    },
+    Warm { device: Option<String> },
     Play {
         path: PathBuf,
         device: Option<String>,
@@ -230,8 +228,7 @@ fn ensure_stream(
         None
     };
     let stale = cached.as_ref().is_none_or(|c| {
-        c.selection != device
-            || (is_default_selection(&device) && c.default_name != default_name)
+        c.selection != device || (is_default_selection(&device) && c.default_name != default_name)
     });
     if stale {
         scrap_stream(cached);
--- a/src-tauri/src/audio_feedback.rs
+++ b/src-tauri/src/audio_feedback.rs
@@ -6,13 +6,21 @@ use rodio::OutputStreamBuilder;
 use std::fs::File;
 use std::io::BufReader;
 use std::path::{Path, PathBuf};
+use std::sync::mpsc;
+use std::sync::OnceLock;
 use std::thread;
+use std::time::Duration;
 use tauri::{AppHandle, Manager};
 
 pub enum SoundType {
     Start,
     Stop,
 }
+
+/// How long a caller that needs the chime to finish (to sequence muting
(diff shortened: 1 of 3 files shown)


cjpais:
So I'm not super happy with the changes as they were submitted. Mostly they didn't compile which is a pretty bad sign. I'm not sure how this was tested, at least on MacOS this does not compile as submitted 

I've reverted the audio feedback related changes. Please give the branch a try in its current state and see if the bug is fixed. I would rather have more targeted changes towards the bug than rewriting things at the moment.

kydechuk01:
Thanks for reverting the part that didn't build — that one's on me, I only compiled on Windows so I missed that cpal's stream is `Send` on WASAPI but not on CoreAudio/ALSA.

The recorder changes you kept are the actual fix for the hang. I've been running exactly these on Windows daily for over a week (audio feedback on, hundreds of transcriptions) with zero hangs, where before it was every 2–30. So this branch as-is fixes it for me.

I redid the audio-feedback half properly and cross-platform as a separate PR #1740 (green Linux CI on my fork, and I verified the macOS build compiles and links — it only fails at codesign because my fork has no Apple cert) so this one can stay focused on the recorder fix. Up to you whether it's worth taking.

Issue #1661, opened 2026-07-11
[BUG] KDE Wayland: paste_method "direct" silently drops ALL capital letters (Shift+key lost in injection)


IcezX2K:
# [Linux/KDE Wayland] paste_method "direct" silently drops ALL capital letters

## Summary
With `paste_method: direct` on KDE Plasma (Wayland session), every character requiring Shift is silently dropped from the typed output. Transcriptions are correct in history.db — only the keystroke-injection stage loses the characters.

## Environment
- Handy 0.9.1, CachyOS (Arch), KDE Plasma 6 Wayland, kernel 7.1.3
- typing_tool: Auto (enigo → X11/XTEST path via XWayland per debug logs)

## Reproduction
1. KDE Wayland session, `paste_method: direct`
2. Dictate: "That's interesting because I also gave it time before I started speaking."
3. Typed output: `hat's interesting because  also gave it time before  started speaking.`

Every capital letter is missing — sentence-initial letters AND mid-sentence ones ("Windows" → "indows", standalone "I" vanishes entirely). Lowercase, digits, and punctuation arrive fine. 100% reproducible across models (whisper-medium, large-v3) — `history.db` shows perfectly capitalized transcripts, so the loss is purely in output injection.

## Why it's nasty
The symptom masquerades as a speech-recognition problem ("the model drops the first letter"), sending users down mic/model debugging paths. We burned an hour on exactly that before diffing history.db against the pasted text.

## Workaround
`paste_method: ctrl_v` (or `ctrl_shift_v` for terminals) works perfectly.

## Suggested fix
On Wayland sessions, either default away from `direct`, or have enigo's Shift handling verified/fixed for the XWayland XTEST path; at minimum a docs warning.

## Related issues
- #429 looks like a milder cousin (first character missing, GNOME): same injection path, but on KDE Wayland the loss is ALL Shift+letter characters, not just the first.
- #1555 (Wayland text-insertion meta).

cjpais:
This is a duplicate issue

PR #1722, opened 2026-07-19
fix(linux): X11 keymap resets silently drop characters in Direct paste (stale enigo keycode bindings)

MeeskaMooska:
## Before Submitting This PR

## Human Written Description

I was using Handy and noticed that the first letter of various words were being truncated consistently. Upon investigating, I found that this had to do with an X11 quirk. It was a quick fix.

## Related Issues/Discussions

Related but distinct root causes: #1661 (KDE Wayland Shift+key race), #429 (Wayland first-char loss).

PR #809 was closed for replacing enigo's `.text()` with manual Shift+key typing. This PR keeps `.text()` unchanged — it only changes when the `Enigo` instance is created.

**Root cause.** On X11, enigo types characters that aren't on the first level of the active layout (capital letters included) by binding their keysyms to spare keycodes, and caches those bindings for the lifetime of the `Enigo` instance. Handy keeps a single `Enigo` in managed state for the whole process. When the X server reloads its keymap (suspend/resume, layout re-apply, device hotplug), the bindings are reverted behind enigo's back, and every affected character is silently dropped from typed output until Handy restarts — while the transcript in history stays correct.

Debug log from the session where I caught it: keycode 184 was bound to `XK_S` at 17:05; after an idle gap containing a keymap reset, a 20:27 dictation dropped every `S`/`I`/`A` (bound before the reset) while `H` survived (first bound at 19:28, after the reset):

```
[17:05:14] enigo::platform::keymap: mapped keycode 184 to keysym XK_S
[19:28:17] enigo::platform::keymap: mapped keycode 219 to keysym XK_H
[20:27:50] Transcription result: So, right now I'm just using Handy and I'm ...
           typed output:         o, right now 'm just using Handy and 'm ...
```

**Fix.** Hold the managed `Enigo` in `Mutex<Option<Enigo>>` and drop-then-recreate it at the start of each paste on X11 (`refresh_enigo` in `input.rs`), so bindings are always made against the server's current keymap. Wayland and other platforms are untouched.

The drop-before-create order matters: enigo's `Drop` (`Drop for Con`, `linux/x11rb.rs` in enigo 0.6.1) reverts its own bindings, and an instance created before the old one is dropped would capture those bindings in its keymap snapshot and keep firing keycodes that die moments later.

A longer-term fix belongs in enigo itself (invalidate the cache on `MappingNotify`), as suggested on #809. This change is complementary: it stays within enigo's public API and keeps Handy correct either way.

Fixes #
Discussion:

## Testing

- Reproduced on Ubuntu 24.04, GNOME on X11 (aarch64), paste method "Direct" with no xdotool installed (enigo fallback path): capitals dropped after suspend/resume, matching the stale-binding timeline above.
- Built the patched arm64 deb and have been running it as my daily dictation driver; capitals type correctly, including after the previously failing suspend/resume condition.
- Sanity-checked the per-paste lifecycle cost: 30 consecutive `Enigo::new` → query → drop cycles against a live X server with no failures; connection setup is milliseconds against a multi-second XTEST typing pass.
- `cargo fmt` clean; release build succeeds.

n/a — the log excerpt above shows the before/after.

## AI Assistance

**If AI was used:**

- Tools used: AI coding assistant (Claude Code)
- How extensively: Helped trace the root cause through Handy's debug logs and the enigo 0.6.1 source, and drafted the patch. I hit the bug in daily use and verified the fix by dictating with the patched build on my machine.

### DIFF
--- a/src-tauri/src/clipboard.rs
+++ b/src-tauri/src/clipboard.rs
@@ -611,10 +611,23 @@ pub fn paste(text: String, app_handle: AppHandle) -> Result<(), String> {
     let enigo_state = app_handle
         .try_state::<EnigoState>()
         .ok_or("Enigo state not initialized")?;
-    let mut enigo = enigo_state
+    let mut enigo_slot = enigo_state
         .0
         .lock()
         .map_err(|e| format!("Failed to lock Enigo: {}", e))?;
+
+    // On X11 a long-lived Enigo instance can hold keycode bindings the server
+    // has since discarded, which silently drops those characters when typing
+    // (see input::refresh_enigo). Rebuild it so this paste runs against the
+    // current keymap.
+    #[cfg(target_os = "linux")]
+    if paste_method != PasteMethod::None && !is_wayland() {
+        input::refresh_enigo(&mut enigo_slot)?;
+    }
+
+    let enigo = enigo_slot
+        .as_mut()
+        .ok_or("Enigo instance is not available")?;
 
     // Perform the paste operation
     match paste_method {
@@ -623,7 +636,7 @@ pub fn paste(text: String, app_handle: AppHandle) -> Result<(), String> {
         }
         PasteMethod::Direct => {
             paste_direct(
-                &mut enigo,
+                enigo,
                 &text,
                 #[cfg(target_os = "linux")]
                 settings.typing_tool,
@@ -631,7 +644,7 @@ pub fn paste(text: String, app_handle: AppHandle) -> Result<(), String> {
         }
         PasteMethod::CtrlV | PasteMethod::CtrlShiftV | PasteMethod::ShiftInsert => {
             paste_via_clipboard(
-                &mut enigo,
+                enigo,
                 &text,
                 &app_handle,
                 &paste_method,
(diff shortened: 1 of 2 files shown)


cjpais:
I'll probably pull this in soon. I've taken a brief look at the code and it seems reasonable, but I want to do a bit further investigation and if possible it would be great to contribute whatever fix upstream as well

MeeskaMooska:
> I'll probably pull this in soon. I've taken a brief look at the code and it seems reasonable, but I want to do a bit further investigation and if possible it would be great to contribute whatever fix upstream as well

Thank you, I agree. This is a temporary downstream workaround, I just proposed the upstream fix for the Enigo team's review here: https://github.com/enigo-rs/enigo/pull/555. It refreshes the cached X11 mappings on MappingNotify and includes regression coverage for the stale-binding case.

cjpais:
Thank you. I will take a look at this when I get a chance. Hopefully, in the meantime, the folks who maintain Enigo will also take a look on their side

Just for heads up, I will eventually merge this after I get in some other slightly higher priority things, but actually, this is quite high priority issue for me since it's very, very easy to replicate

PR #969, opened 2026-03-05
fix: overlay not showing on non-primary monitors

jondeibel:
## Before Submitting This PR

**Please confirm you have done the following:**

- [x] I have searched [existing issues](https://github.com/cjpais/Handy/issues) and [pull requests](https://github.com/cjpais/Handy/pulls) (including closed ones) to ensure this isn't a duplicate
- [x] I have read [CONTRIBUTING.md](https://github.com/cjpais/Handy/blob/main/CONTRIBUTING.md)

## Human Written Description

The recording overlay wasn't showing on non-primary monitors. The fix replaces `work_area()` with `monitor.position()`/`monitor.size()` in `calculate_overlay_position`, since `work_area()` returns incorrect coordinates for monitors with negative positions. The per-platform `OVERLAY_TOP_OFFSET`/`OVERLAY_BOTTOM_OFFSET` constants already account for system chrome (menu bar, taskbar).

This is a single focused change to `calculate_overlay_position` only, based on CJ's fix in e0954ed. Rebased onto latest main (v0.7.10).

## Related Issues/Discussions

Fixes #811

## Community Feedback

Multiple users reported this in #811 across both macOS and Windows multi-monitor setups.

## Testing

- Verified overlay appears on the correct monitor based on cursor position at recording start
- Tested on macOS 26 with 4 monitors (2560x1440 @ 1x, 3024x1964 @ 2x, 1440x2560 @ 1x vertical, 2560x1440 @ 1x)
- Switching between monitors and starting new transcriptions works without flashing or disappearing
- Primary monitor behavior unchanged
- All 44 Rust tests pass, ESLint and formatting checks clean

## AI Assistance

- [x] AI was used (please describe below)

**If AI was used:**

- Tools used: Claude Code
- How extensively: Helped research Tauri coordinate spaces and draft the initial fix, then simplified to match CJ's upstream suggestion (e0954ed)

Issue #811, opened 2026-02-13
[BUG] Visual feedback overlay not displayed when using secondary monitors on Windows


parkghost:
## Before You Submit

Searched existing issues and discussions:
- Discussion #655 — Reports similar overlay behavior in multi-monitor setup on macOS
- Issue #764 — Reports display issue on external monitor on Windows (described as "tray icon", unclear if same issue)

## Bug Description

In a Windows multi-monitor setup, the visual feedback overlay works correctly when using the **primary monitor** — it appears as expected when pressing the transcribe shortcut during recording and transcription.

However, when operating on a **secondary monitor**, pressing the transcribe shortcut produces **no overlay at all** — not on the secondary monitor, and not on the primary monitor either. There is zero visual feedback.

**Steps to reproduce:**
1. Connect two or more monitors (primary + at least one secondary)
2. On the primary monitor, focus an input field and press the transcribe shortcut → overlay displays normally ✅
3. Switch to a secondary monitor and open any application with an input field
4. Focus the input field on the secondary monitor and press the transcribe shortcut

**Expected:** Overlay appears on the secondary monitor (the monitor currently in use)
**Actual:** No overlay is displayed on any monitor (no visual feedback whatsoever)

**Note:** Speech recognition and text insertion work correctly — transcribed text is properly inserted into the input field on the secondary monitor. Only the overlay fails to appear.

Discussion #655 reports a similar multi-monitor overlay issue on macOS.

## System Information

**App Version:**

v0.7.5

**Operating System:**

Windows 11

**CPU:**

AMD Ryzen 9 9955HX 16-Core Processor

**GPU:**

NVIDIA GeForce RTX 5070 Ti Laptop GPU / AMD Radeon Graphics

jsntn:
>     * Issue [[BUG] Tray icon not displayed when input cursor is on external monitor #764](https://github.com/cjpais/Handy/issues/764) — Reports display issue on external monitor on Windows (described as "tray icon", unclear if same issue)

Yes, it is the same issue.

zihaojng:
I'm using two monitors and it's kind of fifty fifty whether the overlay shows up.

willmero:
Seeing this issue as well. Looks to be that the main monitor it doesn't work on, but if I invoke it in a text field on any of the secondary monitors, it does show up along the bottom.

cjpais:
[...]

PR #175, opened 2025-10-04
feat: display overlay on active monitor

wsxiaoys:
## Summary

This change updates the recording and transcribing overlay to appear on the monitor that currently contains the mouse cursor, rather than always appearing on the primary monitor.

This improves the user experience for those with multi-monitor setups.

## Test plan

- [x] Run the app on a multi-monitor setup
- [x] Confirm the overlay appears on the same monitor as the mouse cursor

🤖 Generated with [Pochi](https://getpochi.com)

PR #1739, opened 2026-07-20
fix: refresh macOS cursor monitor after display changes

dudemeister23:
## Before Submitting This PR

- [x] I have searched existing issues and pull requests, including closed ones, to ensure this is not a duplicate

Not applicable. This is a bug fix following the merged multi-monitor overlay changes.

## Human Written Description

Fix for permanent UI location on fallback main monitor after monitors are disconnected and reconnected.

Regular reconnection is a common workflow for laptop owners, and this avoids inconsistent behavior between starting the app and using it after reconnection.

## Related Issues/Discussions

Follow-up to #175 and #969.

Related reports: #811 and Discussion #655.

## Community Feedback

Multi-monitor overlay behavior has previously been reported and confirmed in #811 and Discussion #655. This PR addresses the remaining macOS behavior after disconnecting or reconnecting displays.

## Testing

- Added three unit tests covering Retina cursor conversion, invalid scale factors, and negative-origin monitor matching.
- Focused overlay tests passed: 3 passed.
- Full Rust library test suite passed: 131 passed.
- `cargo fmt --manifest-path src-tauri/Cargo.toml -- --check` passed.
- `cargo clippy --manifest-path src-tauri/Cargo.toml --lib --tests` completed successfully, with only existing warnings in unrelated files.
- Built a production macOS app bundle, signed it locally, installed it, and launched it successfully.
- Manually verified overlay placement on the cursor's monitor before disconnecting an external display, while disconnected, and after reconnecting it.
- Debug logs confirmed successful overlay positioning with no cursor lookup, monitor enumeration, or primary-monitor fallback warnings.

Not included. The behavior was verified through a live monitor disconnect and reconnect cycle plus debug logs.

## AI Assistance

**If AI was used:**

- Tools used: OpenAI Codex
- How extensively: Codex helped diagnose the root cause, implement the focused patch and tests, and run validation. I manually reproduced the issue, validated the fix through monitor disconnect and reconnect testing, and made the final decisions on scope and behavior.

### DIFF
--- a/src-tauri/src/input.rs
+++ b/src-tauri/src/input.rs
@@ -1,5 +1,8 @@
-use enigo::{Enigo, Key, Keyboard, Mouse, Settings};
+#[cfg(not(target_os = "macos"))]
+use enigo::Mouse;
+use enigo::{Enigo, Key, Keyboard, Settings};
 use std::sync::Mutex;
+#[cfg(not(target_os = "macos"))]
 use tauri::{AppHandle, Manager};
 
 /// Wrapper for Enigo to store in Tauri's managed state.
@@ -16,6 +19,7 @@ impl EnigoState {
 
 /// Get the current mouse cursor position using the managed Enigo instance.
 /// Returns None if the state is not available or if getting the location fails.
+#[cfg(not(target_os = "macos"))]
 pub fn get_cursor_position(app_handle: &AppHandle) -> Option<(i32, i32)> {
     let enigo_state = app_handle.try_state::<EnigoState>()?;
     let enigo = enigo_state.0.lock().ok()?;
--- a/src-tauri/src/overlay.rs
+++ b/src-tauri/src/overlay.rs
@@ -1,3 +1,4 @@
+#[cfg(not(target_os = "macos"))]
 use crate::input;
 use crate::settings;
 use crate::settings::{OverlayPosition, OverlayStyle};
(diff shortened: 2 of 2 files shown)


cjpais:
@dudemeister23 can you write explicit before and after expectations, it's not exactly clear to me what the problem this is solving is. just looking so I can reproduce it with my external monitor and validate the fix after I can reproduce the bug itself

dudemeister23:
@cjpais Hi! The sequence for me is it works fine when I start the app while the MacBook is connected to external monitors. Then, obviously, it still works fine when I disconnect the monitors because the laptop then only has the one screen that it needs to anchor to. But then, once I reconnect to the external monitor setup, the anchor is always the main monitor that is selected in that setup, as that is the fallback. And that's the diagnosis 5.6 Sol gave me when inspecting the state as I described the issue. See if you can reproduce this on your end. There may, of course, be many confounding factors. One monitor is attached via USB C, the other via HDMI. One is in vertical orientation and one is in horizontal orientation. And I'm running the Dev Beta 3 of macOS 27. So, if you can't reproduce this issue, it could, of course, be any number of those circumstances. Thanks for looking into it.

cjpais:
At least I'm not able to replicate this with my external monitor, just wondering if you only experience this with all 3 monitors in use? or just two (built in and external I assume)? and if you are able to replicate with two please let me know how I can set my Mac in the same way

dudemeister23:
@cjpais It definitely makes sense to try and reproduce this yourself. What I can say is that I am experiencing it on my setup on an M1 with two external monitors. And as I mentioned, there may be many confounding factors. I also cannot trigger it reliably in the sense of connecting and reconnecting those monitors. It will often work immediately after. But for example, this morning after reconnecting for a long time, the issue surfaced again. 

The diagnosis that 5.6 Sol gave is illustrative in regards to that, and that's exactly what this fix constitutes. I'm giving you below its two sentence version of its description of the issue it found in the code. 

> On macOS, Handy obtains the cursor position from a long-lived Enigo instance that caches CGDisplay::main() when it is created, so after the display topology changes its Y-axis conversion can continue using stale main-display geometry. The resulting cursor coordinates no longer match Tauri’s freshly enumerated monitor bounds, causing get_monitor_with_cursor() to fall through to the primary-monitor fallback; this patch uses a fresh Tauri cursor query for each overlay placement so cursor and monitor data come from the same current display state.

PR #1763, opened 2026-07-22
feat: local post-processing on the loaded transcription model (Voxtral text-to-text)

### LINKED ISSUE #1761: Feature Request: reuse the loaded local model (e.g. Voxtral) for post-processing
## Summary

Models like Voxtral are full LLMs that happen to do speech — not just ASR. When one of them is already loaded for transcription, it could also serve local post-processing (cleanup, formatting, custom instructions), instead of requiring a separate external OpenAI-compatible endpoint as post-processing does today.

Context: this came out of the HN discussion — https://news.ycombinator.com/item?id=48963879

## What's being asked

1. **Reuse the loaded model for text-to-text post-processing.** The weights are already resident; a second inference pass over the transcript avoids running/configuring a separate server. Understood from the HN thread that this needs a header change in transcribe-cpp and isn't free — this issue is to track that design work.

2. **Clear semantics for "instructions" vs. Whisper's `initial_prompt`.** Today the prompt field means different things per model family: for Whisper it's a style/vocabulary biasing hint, for an instruction-following model it could be an actual system prompt. The UI/config should distinguish these so users know what a given model will actually do with the text.

3. **An interface that works across many models.** Since Handy supports many models, the post-processing abstraction should degrade gracefully: models that can't do text-to-text simply don't offer it (or fall back to the existing external endpoint), rather than a Voxtral-specific code path.

## Alternatives considered

- Current custom post-processing endpoint: works, but requires the user to run a second model/server, doubling memory for what could be one set of weights.

## Not in scope

- Any specific model beyond using capability detection (e.g. "supports text-to-text") to gate the feature.

ghost:
## Before Submitting This PR

## Related Issues/Discussions

Fixes #1761
Discussion: https://github.com/cjpais/Handy/discussions/168

This PR is the Handy half of a two-part change. The other half adds a
`transcribe_generate` text-to-text API to transcribe.cpp (branch:
https://github.com/w1nthinker/transcribe.cpp/tree/text-to-text — being
submitted separately by the author). This came out of the HN thread
(https://news.ycombinator.com/item?id=48963879) where @cjpais noted it
"needs a header change" in transcribe-cpp and asked for clear semantics
around instructions vs. whisper's initial prompt, plus interfaces that
work across many models — this PR follows that guidance:

- the header change is `transcribe_generate(_params)` + a
  `TRANSCRIBE_FEATURE_TEXT_TO_TEXT` capability bit in transcribe.cpp;
- instructions stay semantically distinct from whisper's
  `initial_prompt` (which only biases decoding and is documented as
  such on the feature enum);
- the interface is capability-gated per model, so audio-only families
  (whisper, parakeet, …) are untouched and simply don't offer it —
  no Voxtral-specific code path in Handy.

## Community Feedback

Discussion #168 (pinned: Post-processing / Editing Transcripts) has
long-standing interest in local post-processing without running a
separate server; this enables it with zero extra memory when the loaded
model is already an instruction-following LLM (Voxtral today).

## What this PR does

- Adds a **"Local Model (loaded)"** post-process provider. The selected
  post-processing prompt becomes the instruction; the transcription is
  the input; inference runs on the transcribe-cpp session already
  resident in memory (`TranscriptionManager::generate_text`, engine
  take/return + catch_unwind, mirroring `transcribe()`).
- Settings UI: the provider hides base-URL/API-key/model fields and
  shows an info note (i18n key added to `en`; other locales fall back).
- Models without the capability fail the pass gracefully and the
  unprocessed transcript is used, same as a failed HTTP provider.
- `Cargo.toml` TEMPORARILY pins transcribe-cpp to the proposal branch
  until `transcribe_generate` lands upstream and ships on crates.io.
  **This PR should stay in draft until that dependency is resolved.**

## Testing

- `cargo check` / `cargo clippy` clean (no new warnings), `eslint` and
  `tsc --noEmit` pass; `cargo fmt` + prettier applied.
- transcribe.cpp side: full ctest suite passes (33/33) including the
  voxtral smoke tests, after refactoring run()'s decode loop into the
  shared driver `generate` uses.
- End-to-end text-to-text verified against Voxtral-Mini-3B-2507
  Q4_K_M via a standalone harness calling the new
  `Session::generate` (punctuation/filler-word cleanup instruction on
  a filler-laden input) — output attached in a comment.
- Not yet tested: the full Handy UI flow (transcribe → local
  post-process → paste) on all platforms. Happy to get help testing
  Windows/Linux (vulkan/dynamic-backends builds of the pinned branch).

(none yet)

## AI Assistance

**If AI was used:**

- Tools used: Claude Code (Claude Fable 5)
- How extensively: the implementation (both this PR and the
  transcribe.cpp branch), including code, tests-run, and this PR
  body except the Human Written Description section, was written by the
  AI assistant under the direction of @w1nthinker, following the
  design constraints @cjpais outlined in the HN thread. The human
  author will review and own the change before it leaves draft.

### DIFF
--- a/src-tauri/Cargo.lock
+++ b/src-tauri/Cargo.lock
@@ -7167,9 +7167,8 @@ dependencies = [
 
 [[package]]
 name = "transcribe-cpp"
-version = "0.1.3"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "4c3c4d6136eeccf56cfe8a6669e2d63770d1ef051c7cafd2cb9226218c66cded"
+version = "0.2.0"
+source = "git+https://github.com/w1nthinker/transcribe.cpp?branch=text-to-text#69b393b0a89e94480a404048ff75a39c42fa6f9c"
 dependencies = [
  "log",
  "thiserror 2.0.18",
@@ -7178,9 +7177,8 @@ dependencies = [
 
 [[package]]
 name = "transcribe-cpp-sys"
-version = "0.1.3"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "278fd6a6da4d9d8d5f2716bd6761a76ea55c129fda6ba57856b80249a8570ed4"
+version = "0.2.0"
+source = "git+https://github.com/w1nthinker/transcribe.cpp?branch=text-to-text#69b393b0a89e94480a404048ff75a39c42fa6f9c"
 dependencies = [
  "cmake",
  "serde_json",
--- a/src-tauri/Cargo.toml
+++ b/src-tauri/Cargo.toml
@@ -73,7 +73,10 @@ sha2 = "0.10"
 # transcribe-rs is now ONNX-only (Parakeet, Moonshine, SenseVoice, GigaAM,
 # Canary, Cohere). Per-platform backend features are added in the target tables.
 transcribe-rs = { version = "0.3.8", features = ["onnx"] }
-transcribe-cpp = { version = "0.1.3", default-features = false }
+# TEMPORARY (draft): transcribe_generate (local text-to-text post-processing)
+# is not yet in a released transcribe-cpp; pinned to the proposal branch until
+# it lands upstream and ships on crates.io.
+transcribe-cpp = { git = "https://github.com/w1nthinker/transcribe.cpp", branch = "text-to-text", default-features = false }
 handy-keys = "0.3.1"
 ferrous-opencc = "0.2.3"
 clap = { version = "4", features = ["derive"] }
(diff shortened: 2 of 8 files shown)


ghost:
Verification evidence (macOS arm64, Metal, `Voxtral-Mini-3B-2507-Q4_K_M.gguf`), all against the transcribe.cpp `text-to-text` branch:

**Text-to-text (`Session::generate`, the path this PR uses):**

```
instructions: "You are a transcript post-processor. Rewrite the input with correct
               punctuation and capitalization and remove filler words. Reply with
               ONLY the corrected text."
input:        "um so basically the meeting is uh moved to friday three pm"

OUT: The meeting has been moved to Friday, 3 PM.
gen-test: OK
```

**Audio path regression** (run() was refactored to share its decode driver with generate()) — `transcribe-cli` on a `say`-generated clip:

```
run: ok
text: The quick brown fox jumps over the lazy dog.
  realtime:   3x (1130.3 ms for 2.9 s)
```

**Test suites:** transcribe.cpp `ctest` 33/33 passed (includes the api-smoke with new `transcribe_generate` malformed-call/ABI checks); Handy `cargo check`/`clippy` clean, `eslint` + `tsc --noEmit` pass.

cjpais:
I'm gonna close this one for now. I do appreciate you putting this up, especially so other people can play with it, but let's first get things into transcribe.cpp and then do the integration. I think it will be a bit easier because things are too much subject to change right now in regards to this. So I don't want to leave a draft PR hanging around because it's just too likely to change significantly enough

Issue #1101, opened 2026-03-19
[BUG] audio recording issue. Microphone initialize twice


kopach:
## Bug Description

When starting a new recording, the microphone appears to initialize twice in rapid succession. This is visibly noticeable in the UI as the microphone activation indicator blinks and then re-initiates. 

The main problem caused by this double initialization is a slight delay. If I speak immediately after hitting the Transcribe Shortcut, my very first word is often cut half in the recording.

**Steps to reproduce:**
1. Hit the "Transcribe" shortcut.
2. Immediately begin speaking.
3. Stop the recording and play it back.
4. Note that the first word is either missing or cut after around 1 sec

UI showed a double-initialization blink.

![Image](https://github.com/user-attachments/assets/9f6a6baf-e70b-4c73-a907-e298b212569c)

## System Information

**App Version:**
v0.7.11
**Operating System:**
macOS 26.3.1

cjpais:
Does this happen every time and is this using the push to talk mode? Just curious. Can you also provide a log?

What microphone is being used? I'm not really able to replicate on my own machine.

kopach:
Happens every time for me. In both push to talk and "toggle" mode. 
I have Mac Book built in microphone selected. but It doesn't really matter, I've tried to switch
Here are logs.

[handy.log](https://github.com/user-attachments/files/26124735/handy.log)

cjpais:
@VirenMohindra @tanshkoul

tanshkoul:
Couldn't reproduce on 0.7.11 or 0.7.12. Taking a look. Will get back.

VirenMohindra:
same ^ can't replicate on both `.11` or `.12`, tried a combination of short and long transcriptions with post processing enabled

VirenMohindra:
curious, what key have you set up to transcribe w Handy? @kopach

kopach:
> I have had it set to `Ctrl` + `Space`, but now reset to default - same result.

Anyway, I think I have found the issue. I have had option  "Mute While Recording" enabled. When disabled - no issue. 
On the other hand, when enabled + "Audio Feedback" enabled - the issue even more visible.

cjpais:
Thanks for the update @kopach 

Curious @VirenMohindra if you might be able to replicate with these exact settings? I will get around to it, it will just take me time

VirenMohindra:
[...]

Issue #828, opened 2026-02-15
[BUG] Subsequent recordings don't capture anything


qingxiang-jia:
## Before You Submit

Checked all existing bugs, this one doesn't seem to be reported yet.

## Bug Description

Only the first recording (sometimes also the second) can be recorded and the subsequent recording, don't capture any audio (and thus cannot have speech transcribed). This doesn't seem to be related to the hardware I am using. I have tried with Sony (h.ear MDB 100-ABN) and Bose (QuietComfort). I couldn't find a wired microphone, I will update the issue if I can find one.

When running with logs printed, the successful recording looks like this:

```log
[2026-02-15][04:01:09][handy_app_lib::managers::audio][INFO] Microphone stream initialized in 37.505494ms
[2026-02-15][04:01:09][handy_app_lib::audio_toolkit::audio::recorder][INFO] Using device: Ok("default")
Sample rate: 16000
Channels: 1
Format: F32
```

The subsequent (unsuccessful) recording looks like:

```log
[2026-02-15][04:01:26][handy_app_lib::managers::audio][INFO] Microphone stream initialized in 68.811µs
[2026-02-15][04:01:26][handy_app_lib::audio_toolkit::audio::recorder][INFO] Using device: Ok("default")
Sample rate: 16000
Channels: 1
Format: F32
```

As you can see, the unsuccessful one returns so much faster (68us vs 37ms). 

## System Information

**App Version:** 0.7.5 (but actually all versions have this problem)

**Operating System:**

CachyOS (KDE Plasma, both X11 and Wayland have this issue)

**CPU:** AMD Ryzen 7 5700X3D

**GPU:** AMD Radeon RX 6700 XT

cjpais:
Both of these devices are bluetooth yea? There are definitely issues with cpal and bluetooth

qingxiang-jia:
Yes, both are bluetooth device. One is Bluetooth LE, one is an older bluetooth.

qingxiang-jia:
I was able to get a USB camera with microphone (Live! Cam Chat HD VF0790). It works fine (although I was only able to get transcription for the first time, so subsequnt times, there's no recording show up in the history tab). But I guess that's a separate issue.

cjpais:
@DylanBricar would love a fix, and I'm sure the other >100k people using handy would too :) If you know it's actually fixed

DylanBricar:
Hello,

First of all, thank you very much for the original project and for making it open source!

[...]

PR #1774, opened 2026-07-23
feat(audio): native PipeWire capture backend on Linux

olosegres:
## Problem

On Linux, microphone capture is pinned to cpal's ALSA host (`get_cpal_host()` → `HostId::Alsa` in `audio_toolkit/audio/utils.rs`). Handy's capture stream is therefore an ALSA `"default"` client, and PipeWire never sees it as a first-class node. On the (now near-universal) PipeWire desktop this means:

- **The microphone can be far too quiet in Handy specifically.** Many built-in laptop mics are weak and are boosted at the PipeWire source (a WirePlumber/filter gain, or a per-source volume). That gain is applied by PipeWire *after* the node, so it reaches every PipeWire-native app — but **not** an ALSA `"default"` client. Handy therefore captures the raw, un-boosted signal and ends up dramatically quieter than every other app on the same mic (near-silent recordings, poor/empty transcriptions), with no in-app way to fix it: raising the source volume with `wpctl`/pavucontrol has no effect on Handy.
- **No per-application capture volume / routing.** More generally, the stream has no PipeWire node volume, so it can't be attenuated/boosted per-app or routed to a specific source via `wpctl` / pavucontrol / WirePlumber — Handy is invisible to all of that.
- **Device enumeration only ever shows `"default"`** on many PipeWire systems — the ALSA cpal host doesn't surface the graph's friendly per-source nodes — so the mic selector can't offer the real inputs.
- The ALSA path is also comparatively fragile on PipeWire (silent/empty captures, wedged streams); related symptoms: #806, #1349, #828, #1339, #1101.

## Change

Add a **native PipeWire capture backend** (`pipewire-rs` 0.10) that registers Handy's microphone as a real PipeWire node, so it shows up in `wpctl status` with its own settable volume and is routable per-app.

- New `pipewire_recorder.rs`: runs the PipeWire main loop on a dedicated thread (all non-`Send` pw objects stay on it), negotiates F32, downmixes to mono in the RT `process` callback, and feeds the **existing** resampler → VAD → consumer pipeline unchanged via the shared `Cmd`/`AudioChunk` protocol. No DSP is reimplemented.
- New `recorder_backend.rs` (`Recorder`): a thin seam that selects the backend. On Linux it **prefers PipeWire and falls back to the cpal/ALSA path** if PipeWire setup fails (e.g. no running session). Non-Linux targets compile cpal-only and are behaviourally unchanged; cpal stays fully compiled as the fallback everywhere.
- `pipewire` is added only under `[target.'cfg(target_os = "linux")'.dependencies]` (feature `v0_3_44`), so Windows/macOS builds are untouched.

## Testing

Built and run on Fedora (Asahi, aarch64) under GNOME/Wayland with PipeWire 1.6.8. Hotkey → record → transcribe works end-to-end through the native backend; the log shows `Microphone capture using native PipeWire backend`, and the capture node appears in `wpctl status` with a settable volume (`wpctl set-volume <id> …`) — which the old ALSA client did not support (`Node … does not support volume`).

## Scope / follow-ups

This is an MVP that captures the **default source** (reproducing today's `"default"` behaviour). PipeWire-native device **enumeration/selection** is intentionally deferred — there's a `TODO` in `Recorder::open`, where a user-selected mic would map to a `node.name` passed through `TARGET_OBJECT` (already threaded through `PipeWireRecorder::open`). Happy to iterate on device selection and feature-gating in review.

### DIFF
--- a/src-tauri/Cargo.lock
+++ b/src-tauri/Cargo.lock
@@ -98,6 +98,16 @@ dependencies = [
 ]
 
 [[package]]
+name = "annotate-snippets"
+version = "0.11.5"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "710e8eae58854cdc1790fcb56cca04d712a17be849eeb81da2a724bf4bae2bc4"
+dependencies = [
+ "anstyle",
+ "unicode-width",
+]
+
+[[package]]
 name = "anstream"
 version = "0.6.21"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -405,6 +415,25 @@ dependencies = [
 ]
 
 [[package]]
+name = "bindgen"
+version = "0.72.1"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "993776b509cfb49c750f11b8f07a46fa23e0a1386ffc01fb1e7d343efc387895"
+dependencies = [
+ "annotate-snippets",
+ "bitflags 2.11.0",
+ "cexpr",
+ "clang-sys",
+ "itertools",
+ "proc-macro2",
+ "quote",
+ "regex",
+ "rustc-hash",
+ "shlex",
+ "syn 2.0.117",
+]
+
+[[package]]
 name = "bit-set"
 version = "0.8.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -703,6 +732,15 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "6d43a04d8753f35258c91f8ec639f792891f748a1edbd759cf1dcea3382ad83c"
 
 [[package]]
+name = "cexpr"
+version = "0.6.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "6fac387a98bb7c37292057cffc56d62ecb629900026402633ae9160df93a8766"
+dependencies = [
+ "nom 7.1.3",
+]
+
+[[package]]
 name = "cfb"
 version = "0.7.3"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -757,6 +795,17 @@ dependencies = [
  "serde",
  "wasm-bindgen",
  "windows-link 0.2.1",
+]
+
+[[package]]
+name = "clang-sys"
+version = "1.8.1"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "0b023947811758c97c59bf9d1c188fd619ad4718dcaa767947df1cadb14f39f4"
+dependencies = [
+ "glob",
+ "libc",
+ "libloading 0.8.9",
 ]
 
 [[package]]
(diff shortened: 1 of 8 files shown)


cjpais:
Thank you I will take a look at this!

olosegres:
@cjpais any chance you've had a look at this? Been running it from my own fork build for about a month now — works great, zero complaints. Happy to rebase the conflicts if you're open to merging it 👍

cjpais:
Okay, I'm gonna say I really want to merge this. I just need time to review. I'm gonna put this on my priority, so I appreciate you bumping this. This is definitely very important, and we need to merge this. I just need to check the implementation. I've probably been deferring it because it's a thousand lines of new code, but I think that this is very important. So let's definitely get this in ASAP.

Hopefully, we'll have a chance to at least do some minor review this morning and rebase it

cjpais:
Hey @olosegres I have some thoughts. I think it's probably worth doing device enumeration here. And another thing I'm wondering about is whether it makes sense on Linux to add an advanced setting for 'Audio Backend' or similar. With 3 options "Auto", "Pipewire", "ALSA". Part of the reason I'm thinking about this is that I want to make device enumeration clean, and I feel like the only way to do that is to basically have the app always know which backend is actually selected, and that auto would effectively do resolution and prefer pipewire. But if it doesn't exist, fall back. I'm curious what your thoughts are on this and if this is a set of changes that you think you're able to make. Also, I believe that we might need to add some dependencies in the CI to fix it as well if you're able to take a look at that. I do want to merge this. I'm just curious your thoughts and opinions since you're a primary Linux user and I'm not

olosegres:
Hey @cjpais, sorry for the slow reply. Agreed on all three, pushed an update.

Audio Backend setting is in on Linux with your three options. Auto probes PipeWire and falls back to ALSA like before, just explicit now. Pinning PipeWire fails loudly instead of silently falling back, and the resolved backend is what the rest of the app reads, so enumeration stays clean.

Enumeration now goes through that resolved backend. On PipeWire it lists the real graph sources instead of just "default", and the picked device goes through TARGET_OBJECT. ALSA keeps the cpal path.

CI was just missing dev packages. The runners only had the runtime libpipewire, so libspa-sys couldn't find it via pkg-config. Added libpipewire-0.3-dev and libspa-0.2-dev, plus pipewire to the flake.

Also rebased onto current main.

rdslw:
One more data point for the priority of this PR. I traced a "some recordings return 0 samples" problem on Arch/PipeWire to the cpal 0.16 ALSA input worker: after a capture overrun it calls `prepare()` but never `start()`, so the PCM stays PREPARED and the worker blocks in `poll()` forever with no error callback. Handy then logs "Timed out waiting for the microphone callback to pause" and returns nothing. The overrun is easy to hit because the 100 ms ALSA buffer overflows whenever the model load runs at the same time as the recording start. 

Details in #2113

A native PipeWire capture node does not go through that worker at all, so this PR removes the whole failure class on PipeWire desktops, on top of the volume and enumeration benefits.

PR #1779, opened 2026-07-24
feat: add reset button to sliders when value differs from default

Cxkies:
## Human Written Description

While adjusting settings sliders like volume or paste delays, I noticed that if I changed a value and later wanted to restore the default, there was no easy way to do it without remembering the exact number. I added a "Reset" button that automatically appears when the slider value differs from its default. It's a small detail but makes configuration much more user-friendly.

## Related Issues/Discussions

No related issues or discussions found.

## Community Feedback

Small UX improvement. No community feedback was gathered.

## Testing

- Added `defaultValue` prop to Slider component
- When `defaultValue` is provided and `value !== defaultValue`, a "Reset" text button appears right-aligned below the slider
- Clicking Reset restores the slider to its default value via `onChange(defaultValue)`
- Button disappears when value matches default
- Tested with all 5 sliders: volume, word correction threshold, paste delay (x2), recording buffer
- TypeScript and lint pass without errors

## Screenshots/Videos

**Before:**
reset-original

**After:** 
reset-modificado

## AI Assistance

**If AI was used:**
- Tools used: opencode (AI coding assistant)
- How extensively: AI implemented the feature, user reviewed and adjusted UI

### DIFF
--- a/src/components/settings/debug/PasteDelay.tsx
+++ b/src/components/settings/debug/PasteDelay.tsx
@@ -21,7 +21,7 @@ export const PasteDelay: React.FC<PasteDelayProps> = ({
   descriptionKey = "settings.debug.pasteDelay.description",
 }) => {
   const { t } = useTranslation();
-  const { settings, updateSetting } = useSettings();
+  const { settings, updateSetting, resetSetting, isUpdating } = useSettings();
 
   const handleDelayChange = (value: number) => {
     updateSetting(settingKey, value);
@@ -31,6 +31,8 @@ export const PasteDelay: React.FC<PasteDelayProps> = ({
     <Slider
       value={settings?.[settingKey] ?? 60}
       onChange={handleDelayChange}
+      onReset={() => resetSetting(settingKey)}
+      isResetting={isUpdating(settingKey)}
       min={10}
       max={500}
       step={10}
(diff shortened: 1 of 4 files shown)


cjpais:
Not sure about this, it doesn't fit the existing UI very well...

Cxkies:
Thank you for the feedback. I understand the concern about the UI. Would you be open to this feature if the reset was shown as a circular arrow icon (↻) instead of text? That would make it more compact and consistent with standard UI patterns. The functionality is the same — it simply restores the slider to its default value when changed — but I think the icon approach would blend better with the existing design. Happy to update the PR if that works for you.

cjpais:
Yeah I would at least like to see it with that, but I'm still not sure about how it will look or not

Cxkies:
> Yeah I would at least like to see it with that, but I'm still not sure about how it will look or not

I've updated the implementation following your feedback. Instead of text, the reset now uses the same circular arrow icon (↻) that's already used throughout the app, like in general settings, post processing, shortcuts, and more. It sits inline next to the slider value, always visible. Personally, I think it would be great if these reset buttons were hidden by default and only appeared when the value differs from the default, but the current approach (always visible like the rest of the app) works well too, it's really just a preference. If this looks good to you, let me know and I'll update the PR accordingly.

image

cjpais:
I think it's fine, I think my only preference is for sliders outside of the debug menu we probably should leave them as is for now, I'll take a look when I have time but may take a bit for the merge

Cxkies:
Updated as suggested — reset button is now only on Debug sliders. Ready for review when you have time.

PR #1790, opened 2026-07-26
feat: upgrade Silero VAD from v4 to v6.2

NairoDorian:
## Human Written Description

Upgraded the Silero VAD (Voice Activity Detection) model from v4 to v6.2. The old silero_vad_v4.onnx (1.8 MB) has been replaced with the latest non-quantized silero_vad_v6.2.onnx (2.3 MB) from https://huggingface.co/BricksDisplay/silero-vad-6.2. Silero VAD 6.2 provides improved speech detection accuracy while maintaining the same ONNX interface, so no code API changes were needed.

The vad-rs crate wraps Silero VAD via ONNX Runtime and uses identical tensor I/O names across all Silero VAD versions, making this a drop-in model upgrade.

## Related Issues/Discussions

No related issues or discussions yet. This is a proactive upgrade.

## Testing

Build and run verified with bun run tauri dev:
- Model loads correctly from silero_vad_v6.2.onnx
- VAD logs confirm the new model path is used at runtime
- Transcription works end-to-end with the new model
- No fallback to VAD-less mode: recorder requires VAD preload to be available

## AI Assistance


cjpais:
Looks like you updated other files as well unrelated to the changes. Was this intentional?

cjpais:
Okay, I'm closing this for three reasons. Number one is you didn't respond to my former question. Number two is this upgrades all the dependencies, which was not asked for and effectively requires some other verification. I prefer not to do that. You can resubmit it without bumping all of the dependencies. And number three is I'm just genuinely not sure whether this is necessary or not. Like, it could be better for sure. But broadly speaking, I don't feel like the current version which is shipped is largely a problem. So yeah, I just want to keep the changes smaller. Like this introduces a lot of churn in the code base and like effectively it's not mergeable as is. And as a result of that, I don't want to spend my time looking through this. So if you're willing to resubmit this with a much smaller set of targeted changes, it's much more likely to be pulled in, but I can't waste my time right now

PR #1811, opened 2026-07-30
dont fire bindings when recording

cjpais:
This has been an annoying issue for me for some time. I don't know if it is a reported one, but certainly if you have, say, function set as a binding and then you want to record function plus one as a binding, it will fire the recording, which is kind of annoying and definitely needs to be fixed


cjpais:
I tested this and it fixes a long-standing bug for me. I'm just gonna merge and we will deal with any potential consequences in the future

PR #1813, opened 2026-07-30
fix(audio): recalibrate mic level meter so speech actually moves the bar

### LINKED ISSUE #1694: [BUG] Audio waveform / mic level meter not animating on macOS (Apple Silicon)
Dictation works perfectly — text transcribes fine — but the live audio waveform / mic level indicator never moves while I'm speaking.

Setup:

- MacBook Air, M4 (Apple Silicon)

- Handy 0.9.3 (aarch64)

- Model: Canary 180M Flash

Tried without any change:

- Switching the input device and back

- Full quit + relaunch

- Clean reinstall with config reset (~/Library/Application Support/com.pais.handy)

Purely cosmetic since transcription is unaffected, but wanted to flag it.

gauravsingh06:
The `db` value computed in AudioVisualiser::feed() is not true dBFS: it is a per-bin average magnitude divided by the FFT window size, which for speech-shaped input lands roughly 20 dB below the band's actual dBFS. DB_MIN/DB_MAX were calibrated as if it were dBFS, so the window sat about 20 dB too high:

- src-tauri/src/audio_toolkit/audio/visualizer.rs (level buckets)
- src/overlay/RecordingOverlay.tsx (draws 3 + v^0.7 * 15 px bars)

Normal dictation therefore mapped to ~0.1, which the overlay draws as a 4 px bar against a 3 px floor and an 18 px maximum — roughly one pixel of travel, which reads as a frozen waveform.

Recalibrate to -68/-30, measured against captured audio on a built-in mic at 48 kHz (dictation ~-32 dBFS median, room tone ~-48 dBFS) and scored on rendered pixels: speech gets ~5 px of travel while room tone stays pinned at the floor. Not lowered to -70, where a noisy room starts making the idle waveform twitch.

Fixes #1694

## Before Submitting This PR

## Human Written Description
I myself was facing the issue. So I gone through the issue tab, then I found the same issue, then I tried to fix it. I am contributing for the first time.

Before 
image

After
image

## AI Assistance

**If AI was used:**

- Tools used: Claude Mythos
- How extensively: Fully, but I tested it manually too

### DIFF
--- a/src-tauri/src/audio_toolkit/audio/visualizer.rs
+++ b/src-tauri/src/audio_toolkit/audio/visualizer.rs
@@ -1,8 +1,14 @@
 use rustfft::{num_complex::Complex32, Fft, FftPlanner};
 use std::sync::Arc;
 
-const DB_MIN: f32 = -55.0;
-const DB_MAX: f32 = -8.0;
+// `db` below is not true dBFS: it's a per-bin average divided by the FFT
+// window size, which lands ~20 dB low for speech. So this window is calibrated
+// against measured mic audio (dictation ~-32 dBFS, room tone ~-48 dBFS) rather
+// than absolute dBFS. The old -55/-8 left speech ~1 px above the overlay's
+// floor, which reads as a frozen waveform (#1694). Not lowered past -68: at
+// -70 a noisy room starts making the idle waveform twitch.
+const DB_MIN: f32 = -68.0;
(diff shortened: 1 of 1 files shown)


cjpais:
Thank you for this. I did test this on my machine and it is at minimum better on average, so I'm merging it. Thank you

PR #1815, opened 2026-07-30
feat: include streaming and translation filter buttons for easy access of models

vscg:
## Before Submitting This PR

## Human Written Description

Note: This description is not human written but human spoken and transcribed using Handy :D

When I want to pick a streaming model, I had to scroll up and down the entire big list to find what are all the streaming specific models that are available. I believe the list will only grow even more in the future as more models are released. So there needs to be a way to allow me to find the models with specific capabilities that I want. 

Currently there are two additional capabilities streaming and translation so I added two buttons before language filter which are used to filter the models for those specific capabilities. These two buttons can be used individually or can be combined together along with the existing language drop down.

## Related Issues/Discussions

Fixes #
Discussion: N/A 

Since this change is just a minor ui enhancement, I thought it would be overkill to create a discussion for this

## Community Feedback

N/A - See above.

## Testing

Verified that clicking the new buttons correctly filters the model list in real-time.

shows only streaming models (notice how big the scrollbar is as there are only a few models)
Screenshot 2026-07-31 010135

shows models that can translate and support a specific language simultaneously
Screenshot 2026-07-31 010917

## AI Assistance

**If AI was used:**

- Tools used: Antigravity
- How extensively: AI generated code with manual supervision and editing.


cjpais:
@vscg I think I'm pretty likely to pull this in. The one thing that I am wondering about is maybe can you make the rescan button actually not have text just so it's more in line with everything else. And then I don't know whether or not these buttons have a hover state on them. I haven't tested this build myself yet. But I do suspect that they do need some level of hover state just to make sure that people can hover over things and understand what these filters do

MikaKrul:
Is it me or are those buttons not the same size as the other's?

cjpais:
@MikaKrul you're right, that also needs to be fixed

MikaKrul:
Hey @vscg & @cjpais! I liked the idea of filtering models by capability, but since the filter buttons were taking up a bit of space and having sizing issues, I decided to build an alternative approach using a dropdown-menu. It addresses the UX/alignment feedback and includes the re-scan button change as well.

I've submitted it in #1864 as a fresh implementation if you'd like to check it out!

vscg:
@cjpais , the buttons do have tooltip on hover which use the same text property as the ones from model card.

@MikaKrul , honestly ur dropdown idea is the right way to handle filters, as it gives flexibility to add more filters in future without any code change and screen space.
I initially had the same idea, and started with dropdown, but when i tried to use it, it felt annoying to click multiple times, as currently we have only two options. 
So, If the app is expecting more options in near future i would go with dropdown, but if not then will want to use these buttons as they give quick convenience(for only two options) and move to dropdown later when we have at least 4 options.

I have to say, you got a sharp eye. The height difference might be due to buttons not having any text like others beside(as i used same classes), they might just need some padding/height. I will update them if we decide to go with buttons instead of dropdown.

cjpais:
I think for right now I prefer the buttons as I don't expect to really want to add many more filters to this at the moment. 

I think in the future a dropdown or rethought UI might make it into an "advanced" UI

vscg:
now all buttons have fixed height and tooltip shows description instead of just label

Screenshot 2026-08-07 075817

cjpais:
overall im happy enough with this change, thanks

MikaKrul:
Fair enough! Makes sense to keep it simple with buttons for now. @vscg and @cjpais appreciate the work!

PR #1832, opened 2026-08-01
feat(files): add offline batch transcription for audio files on disk

javi22020:
Handy could only transcribe what it recorded itself. Anyone with existing audio — a voice memo, a meeting recording, a podcast — had no way to run it through the model they already have set up locally, short of replaying it into the mic.

Adds a Files tab that queues user-picked audio files and transcribes them sequentially with the selected model:

- `audio_toolkit::decode_audio_file()` handles arbitrary user audio, where `read_wav_samples()` only ever worked for Handy's own 16 kHz mono recordings. It decodes via symphonia (wav/mp3/flac/ogg/m4a), downmixes by averaging channels to match `recorder.rs`, and streams through `FrameResampler` to 16 kHz rather than collecting the source first — an hour of 44.1 kHz stereo is over a gigabyte as interleaved f32.
- `commands/files.rs` runs the batch strictly one file at a time. `TranscriptionManager::transcribe` takes the engine out of its mutex for the duration of inference, so a concurrent caller would just fail with "Model is not loaded"; hence the up-front rejection when recording or streaming, and global shortcuts suspended for the life of the batch. Cancellation is honoured between files — inference itself is uninterruptible. Drop guards release the batch slot and restore shortcuts even on panic, so a crash mid-batch cannot wedge the tab as "running".
- Nothing is written next to the source audio. Transcripts stay in memory until the user picks a destination in a save dialog; the ZIP is written in Rust because the `fs` capability scope is `$APPDATA`-only and read-only, so JS can neither read `~/Music` nor write the archive. Archive entry names are sanitised (traversal, illegal chars, case-insensitive collisions) rather than trusted.
- Queue state lives in a Zustand store initialized from `main.tsx`, not in the tab: switching tabs unmounts the component, and a running batch has to survive that without dropping progress events.

The two new crates were picked to add nothing to the build: symphonia's feature list maps onto the sub-crates rodio already pulls in, and zip is already in Cargo.lock via tauri-plugin-updater (`deflate-flate2` reuses the miniz_oxide backend flate2 is compiled with).

Covered by tests for decode (stereo 44.1 kHz downmix/resample, 16 kHz mono passthrough, non-audio rejection) and for archive naming.

## Before Submitting This PR

## Related Issues/Discussions

Fixes #
Discussion: https://github.com/cjpais/Handy/discussions/299

## AI Assistance

**If AI was used:**

- Tools used:
- How extensively:

### DIFF
--- a/src-tauri/Cargo.lock
+++ b/src-tauri/Cargo.lock
@@ -2484,6 +2484,7 @@ dependencies = [
  "specta",
  "specta-typescript",
  "strsim",
+ "symphonia",
  "tar",
  "tauri",
  "tauri-build",
@@ -2509,6 +2510,7 @@ dependencies = [
  "vad-rs",
  "windows 0.61.3",
  "winreg 0.55.0",
+ "zip",
 ]
 
 [[package]]
@@ -8972,6 +8974,7 @@ checksum = "caa8cd6af31c3b31c6631b8f483848b91589021b28fffe50adada48d4f4d2ed1"
 dependencies = [
  "arbitrary",
  "crc32fast",
+ "flate2",
  "indexmap 2.13.0",
  "memchr",
 ]
--- a/src-tauri/Cargo.toml
+++ b/src-tauri/Cargo.toml
@@ -69,6 +69,28 @@ rusqlite = { version = "0.37", features = ["bundled"] }
 tar = "0.4.44"
 flate2 = "1.0"
 sha2 = "0.10"
+# Decoding for the Files tab (offline batch transcription of user-supplied
+# audio). The feature list maps exactly onto the symphonia sub-crates rodio
+# already pulls into Cargo.lock — adding adpcm/mkv/alac/caf here would mean new
+# downloads. Used directly rather than via `rodio::Decoder`, whose sample type
+# we cannot pin down: the cjpais rodio fork carries 0.21 APIs under a 0.20.1
+# version string.
+symphonia = { version = "0.5.5", default-features = false, features = [
+  "flac",
+  "mp3",
+  "aac",
+  "pcm",
+  "vorbis",
+  "isomp4",
+  "ogg",
+  "wav",
+] }
+# Results export for the Files tab. Already in Cargo.lock at 4.6.1 via
+# tauri-plugin-updater. `deflate-flate2` — not the broader `deflate` — is
+# deliberate: `deflate` also pulls zopfli and switches flate2 to the zlib-rs
+# backend, both new crates. `deflate-flate2` reuses the miniz_oxide backend
+# flate2 (already a direct dep) is compiled with, so this adds nothing.
+zip = { version = "4", default-features = false, features = ["deflate-flate2"] }
 # Whisper-family models run through transcribe-cpp (GGUF/ggml, native backends);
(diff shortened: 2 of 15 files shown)


cjpais:
Clearly, an AI wrote this with absolutely no human attempt to, so why should I care. No screenshots, nothing.

javi22020:
Hey, I had the screenshots ready but was still editing the PR description when it got closed. Here they are:

Captura de pantalla_20260801_110128
Captura de pantalla_20260801_110119

Happy to address any other feedback if you'd consider reopening.

cjpais:
I want to know what thought process and decisions you made as a human that went into this, and how it's a feature that is ready to ship to 10-100k people from a user experience perspective

How is it differentiated from the other attempts at this previously? Have you looked at them? How does it integrate with history? Etc etc.

I know this is a feature many people want but it needs to be thoughtfully considered. Of course anyone can just prompt and LLM to make some stuff, but i want to know what, and why the decisions were made. 

Surely you read new features are not being accepted, and largely this is true. And i need someone who's willing to bring the feature in and then continue to support it for at least 6 months after it's merged. 

I'm being direct because i have limited time and there are so many bugs to still fix, and features just add new bugs. And user experience is a top concern with any major feature that comes in and usually it's not something that LLMs consider very well

javi22020:
I understand your position, it's totally normal wanting to avoid AI slop.
And yes, I read that new features aren't accepted, but I really believe this would be a good addition; it's something I'd use myself (I tend to ramble ideas and descriptions for several minutes).

I believe this PR isn't slop because:
- I defined the scope and design of the feature myself (the Files tab, batch approach, accepted input formats, etc.), Claude didn't choose that.
- The diff is relatively small (1200 lines across 15 files), and probably could be trimmed further if really necessary.
- It wasn't a quick prompt; I used /plan, corrected some aspects, then delegated to /ultraplan on the cloud, then steered Claude through the implementation. I also used two of the best models currently (Opus and Fable 5).
- I ran the Linux build on my PC (I use Debian) and the program runs fine (it's where I took the screenshots).

Still, I don't pretend it's perfect yet; I wanna add VAD to the offline transcription too (so models don't hallucinate in silent segments), and translations should be added to the interface (it's visible in the screenshots that the files tab is still in English with the Spanish locale).

Let me know your opinion, or additional feedback 👍🏼

cjpais:
What you described is what you built, not what decisions you made. As far as I've read it, you delegated your thinking to plan and then ultraplan, but this does not address what I'm talking about in regards to user experience. And **why** decisions were made or accepted.

I think this will remain closed, and you're welcome to reopen a new PR at some point, but spend a week on this feature and get back to me if it's something that you really care about in getting into the application. This feature would take me at least two weeks, even with AI assistance, to accept and merge if I implemented it. So that's the bar that effectively needs to be met. It needs 10 to 50 hours of work ultimately of human decisions and thought process that goes into this. Handling all the corner cases and how it touches every part of the app

Like, even just looking at the screenshots, to me, there seems so many unnecessary features that I don't know if you prompted them if the model just added them because it thought it was a good idea. The UI is just like, to be honest, nothing special and not obviously user-friendly and like the app is already not as user-friendly as I want, and I don't want to add more complexity unless it has basically great user experience

PR #1862, opened 2026-08-05
fix: emit parseable names for compound shortcut keys

### LINKED ISSUE #1848: Cannot set Scroll Lock as shortcut: parses as "scroll lock" instead of "ScrollLock"
## Before You Submit

I searched existing issues and didn't find this reported.

## Bug Description

Setting the transcription hotkey to Scroll Lock fails with a parse error, even though Scroll Lock is a supported key in the underlying `muda`/`global-hotkey` crate that Handy uses for shortcut registration.

**Error message:**

Failed to register shortcut: Error: Failed to register shortcut: Failed to parse shortcut 'scroll lock': Couldn't recognize "scroll lock" as a valid key for hotkey, if you feel like it should be, please report this to https://github.com/tauri-apps/muda

**Root cause (checked in muda source):**

muda's accelerator parser only accepts the key name "SCROLLLOCK" (no space) - see accelerator/mod.rs, match arm "SCROLLLOCK" => ScrollLock. Handy appears to generate the string "scroll lock" (with a space) when capturing the key press for the shortcut recorder, which never matches the parser's expected format. This looks like a Handy-side bug in how the pressed key is converted to a string for muda, not a muda limitation - Scroll Lock is fully supported by muda on all three backends (Windows/GTK/macOS), just under a different string format ("ScrollLock" / "Scroll_Lock" depending on backend).

**Steps to Reproduce:**

Step A: Open Settings, then General.
Step B: Click to record a new "Transcription Hotkey".
Step C: Press Scroll Lock.
Step D: Error appears, shortcut is not saved.

**Expected behavior:** Scroll Lock should be accepted as a valid hotkey, same as any other key.

## System Information

**App Version:**
0.9.4

**Operating System:**
Debian (Linux), X11 session

**CPU:**
Not relevant to this bug (key-name parsing issue)

**GPU:**
Not relevant to this bug (key-name parsing issue)

## Logs

Not applicable - this is a shortcut-parsing error shown directly in the Settings UI, not something requiring log inspection.

BrianMcBrayer:
## Before Submitting This PR

## Human Written Description

I decided to see if I could fix #1848 where hotkeys with spaces in them failed to save as a hotkey. I compared the library that we use vs.
how we emit, and then I fixed our emitter. Added tests that proved the bug and then went green when the bug was fixed.

However, one hotkey cannot be fixed here, because "menu" is not something the upstream library recognizes.

This fix will fix all other valid hotkeys with spaces in them, however.

## Related Issues/Discussions

Fixes #1848
Discussion:

## Community Feedback

Bug confirmed by the reporter on Debian/X11 with Handy 0.9.4 (Tauri global shortcuts). Follow-up comment also notes Menu/ContextMenu fails; that key is **not** fixed here (see Testing / notes below) because `global-hotkey` 0.8.0's string parser does not accept `MENU`/`CONTEXTMENU` at all — that needs an upstream change.

## Testing

The frontend check runs in the code-quality workflow:

```bash
bun run test:keyboard
```

The Rust test verifies the emitted tokens against both shortcut parsers:

```bash
cd src-tauri && cargo test --lib shortcut::tests::compound_shortcut_keys_parse_on_both_backends
```

Coverage includes compact storage names and friendly display labels for `scrolllock`, `capslock`, `numlock`, `pageup`, `pagedown`, and `printscreen`, plus the generic `KeyboardEvent.code` fallback (`AudioVolumeUp` → `audiovolumeup`).

Also validated locally:

- `bun run lint`
- `bun run build`
- `bun run format:check`

**Not covered / out of scope:**

- **Menu / ContextMenu:** `global-hotkey` 0.8.0 does not parse it. HandyKeys 0.3.4 does, but there is no shared token accepted by the Tauri parser.
- **Numpad:** the backends use different names (`numpad*` versus `keypad*`).
- OS-level registration still requires manual testing in a packaged app with the relevant permissions.

**Manual verification still useful on a real build:**

1. Settings → General → record Transcription Hotkey → press Scroll Lock → should save (no parse error).
2. Same for Page Up / Caps Lock if available.
3. Confirm existing shortcuts (e.g. default / Space-based) still display and work.

N/A (parse/registration fix; no intentional UI redesign). Display still shows “Scroll Lock” etc. via label map.

## AI Assistance

**If AI was used:**

- Tools used: Grok (xAI) coding assistant and Pi coding agent
- How extensively: Co-written with AI assistance; Pi was also used for maintainer review, test cleanup, and validation.

### DIFF
--- a/src-tauri/src/shortcut/mod.rs
+++ b/src-tauri/src/shortcut/mod.rs
@@ -1340,3 +1340,52 @@ pub async fn get_available_accelerators() -> crate::managers::transcription::Ava
         .await
         .expect("get_available_accelerators panicked")
 }
+
+#[cfg(test)]
+mod tests {
+    use handy_keys::Hotkey;
+    use tauri_plugin_global_shortcut::Shortcut;
+
+    /// Compound key names that the Tauri shortcut recorder used to emit with
+    /// spaces (e.g. "scroll lock"). global-hotkey rejects spaced tokens, so
+    /// registration failed (#1848). The frontend now emits compact forms
+    /// (e.g. "scrolllock").
+    ///
+    /// Handy has two keyboard backends. This test locks the invariant that
+    /// the compact names we store are accepted by *both* parsers, so fixing
+    /// Tauri recording does not break HandyKeys (or switching between them).
+    ///
+    /// Only keys in the intersection of both parsers are listed here.
+    /// Excluded on purpose:
+    /// - Menu/ContextMenu: neither backend's string parser accepts it
+    /// - PrintScreen: Tauri yes, handy-keys has no Key variant
+    /// - Numpad keys: backends use different naming (numpad* vs keypad*)
+    #[test]
+    fn compound_shortcut_keys_parse_on_both_backends() {
+        // Compact forms produced by getKeyName after the #1848 fix
+        let keys = ["scrolllock", "capslock", "numlock", "pageup", "pagedown"];
+
+        for key in keys {
+            // Tauri path: tauri_plugin_global_shortcut → global-hotkey
+            assert!(
+                key.parse::<Shortcut>().is_ok(),
+                "tauri/global-hotkey should parse compact key '{key}'"
+            );
+            // HandyKeys path
+            assert!(
+                key.parse::<Hotkey>().is_ok(),
+                "handy-keys should parse compact key '{key}'"
+            );
+        }
+
+        // Spaced forms (pre-fix Tauri emission) must still fail — documents the bug
+        assert!(
+            "scroll lock".parse::<Shortcut>().is_err(),
+            "spaced 'scroll lock' must remain unparseable by Tauri backend"
+        );
+        assert!(
+            "scroll lock".parse::<Hotkey>().is_err(),
+            "spaced 'scroll lock' must remain unparseable by HandyKeys backend"
+        );
+    }
+}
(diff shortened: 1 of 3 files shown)


cjpais:
I'm pretty sure likely this is to break one of the two implementations because this is only happening as far as I can tell with the Tauri global shortcuts and not with handy keys. So, it will need special handling and testing on both. Make sure that you've tested both implementations

BrianMcBrayer:
That's a good point. I had checked those, but I didn't put a test in to verify that we wouldn't drift in the future.

From what I found in the code (correct me if I'm wrong):
The spaced-name bug only shows up on the Tauri recorder path (`GlobalShortcutInput` → `getKeyName`). The HandyKeys recorder never uses that helper; it gets `hotkey_string` from the backend. So HandyKeys was already emitting compact names like `scrolllock`.

The compact forms we now emit are the intersection both parsers accept (`scrolllock`, `capslock`, `numlock`, `pageup`, `pagedown`). I added a Rust unit test that actually parses those strings through both `tauri_plugin_global_shortcut::Shortcut` (global-hotkey) and `handy_keys::Hotkey` to prevent regression.

Menu / PrintScreen / numpad are intentionally excluded there (commented in the test) because they are not shared by both parsers.

BrianMcBrayer:
I also re-reviewed and realized that my legacy handling is cruft. I can't think of a situation where we would need to accept the old format in a new binary since we don't store these anywhere.

cjpais:
thx

PR #1865, opened 2026-08-07
fix js-yaml quadratic omap resolution

gwendall:
## Summary

- update the transitive js-yaml resolution from 4.1.1 to 4.3.1
- refresh the Bun lockfile integrity metadata
- remediate GHSA-5p4m-2wfm-xmqj

## Validation

- `bun install --frozen-lockfile --ignore-scripts`
- resolved-version scan: only js-yaml 4.3.1
- `git diff --check`

The change keeps the existing js-yaml 4.x major line and uses the minimum fixed release.


gwendall (commented):
Local Standards/Spec review: PASS. The change is limited to the Bun lockfile, stays within js-yaml 4.x, uses the minimum fixed release, and passes a frozen Bun install.

PR #1868, opened 2026-08-07
test build for reliable paste

cjpais:


cjpais:
Not convinced about this approach

Issue #147, opened 2025-09-29
Auto push to talk and command line options


jamaggs:
Hi, I just wanted to start by saying what a fantastic program. I have two minor suggestions to improve it:

1. What would be really good is if a brief press of the shortcut would bring up the dictation and then require another press of the shortcut to end it. However, a press and hold would bring up push to talk. That way, you wouldn't have to choose between push to talk or the other setting. Personally, I find both quite useful in different circumstances.
2. It would be really great if there were some simple command line options. For example, to start dictation either under push to talk mode or under the other mode (a different command line option for each). This way, it would be possible to assign shortcuts in my window manager, which would be more flexible than just the handy shortcut. For example, having a single modifier key to open the dictation.

Many thanks again. I have been looking for something like this for ages. For use in Linux, And handy is really great.

cjpais:
I like the feature you're describing. To me it sounds like a good default behavior. I don't the current code will support it well, but will consider it when the keyboard stuff gets rewritten at some point

Contemplating having a cmdline version entirely, which is a bit more bleeding edge than the UI version and more for power users. Allowing faster iterations without as much worry about cross platform stability as it should be much easier to package and distribute I expect. Right now you can do it all in the UI and I'm not interested in particular about adding cmdline options unless someone submits a PR for it.

retspen:
I think this is the good idea that could solve the problem with Wayland. If we will have this feature, we can use “Custom Shortcuts” in the Gnome settings for push-to-talk ON and OFF actions.

P.S.: Thanks for the great app.

codepunkt:
Either this or having the option of defining two shortcuts, one for push to talk and one for toggle would be really beneficial.

samleibowitz:
> Either this or having the option of defining two shortcuts, one for push to talk and one for toggle would be really beneficial.

+1 for this. The PTT is great, but the recognition models are getting so good that I find myself wanting to transcribe longer and longer segments.

yachal-upson:
+1 Came to find a feature request or make my own for being able to 'lock' transcription on for a time. 
The OP's description of how the UX might work would be ideal.

AlexPako:
+1 for the combined mode — this is the one thing I miss after switching to Handy.

[...]

PR #1877, opened 2026-08-08
feat: add opt-in always-toggle shortcut alongside push-to-talk

ljden:
## Before Submitting This PR

  - [X ] I have searched [existing issues](https://github.com/cjpais/Handy/issues) and [pull
  requests](https://github.com/cjpais/Handy/pulls) (including closed ones) to ensure this isn't a duplicate
  - [X ] I have read [CONTRIBUTING.md](https://github.com/cjpais/Handy/blob/main/CONTRIBUTING.md)

  ## Human Written Description

I just downloaded Handy today and after using it for a few hours it was the first feature I noticed that was missing.
There's definitely times where I want to be able to just quickly pop something in and other times where a longer bit of text is needed and therefore having the ability to do both is good.

The main thing that I didn't do here that I would have preferred to do is to change the push-to-talk pill control from the current `toggle | PTT` behaviour, to instead be `disabled | enabled`, with the separate control for the toggle feature.
The one other thing that I didn't do was make the default binding as unset - as it seems like that's not currently supported. That becomes more useful with the independent controls. 

I did see the [discussion early in the year](https://github.com/cjpais/Handy/discussions/211#discussioncomment-15656742) around the fact that you weren't sure about adding this at the time, so fair enough if you want to reject it! 

  ## Related Issues/Discussions

  Related to #147 — implements the two-shortcut variant proposed in the comments
  (a separate toggle shortcut alongside push-to-talk), rather than the
  press-vs-hold hybrid in the original post.

  **Design decisions worth flagging:**

  - Opt-in and off by default: a new `transcribe_toggle_enabled` setting gates
    registration exactly the way `post_process_enabled` gates the post-process
    shortcut, so existing users see zero change (the frozen v0.9 settings-store
    test still passes; the new field is `#[serde(default)]`, no migration).
  - With Push To Talk **off**, the main transcribe shortcut already toggles, so
    enabling this one is redundant (though harmless). It is deliberately *not*
    gated on `push_to_talk` — that would couple registration to two settings and
    create hidden-active-shortcut edge cases for a cosmetic gain. Happy to add
    that gating as a follow-up if preferred; per-binding modes seem better left
    to the keyboard rewrite mentioned in #147.
  - The Cancel shortcut row is now shown whenever any shortcut can start a
    toggle-mode recording (previously hidden whenever Push To Talk was on).

  ## Community Feedback

  In #147, @cjpais said "I like the feature you're describing", and @codepunkt /
  @samleibowitz specifically asked for "defining two shortcuts, one for push to
  talk and one for toggle" — which is what this PR adds.

  ## Testing

  - `cargo check`, `cargo clippy`, `cargo fmt --check` clean (no new warnings)
  - `cargo test`: 177 passed, including the frozen v0.9 settings-store
    compatibility test
  - `bun run lint`, `tsc`, `prettier --check`, `vite build` all pass
  - Manual on macOS: tested with new toggle both enabled and disables to confirm correct behaviour. Also tested both states of PTT with the new explicit toggle key combo.

image
image

  ## AI Assistance

  **If AI was used:**

  - Tools used: Claude Code (Sonnet 5 wrote the initial implementation; Fable 5 reviewed and reworked it)
  - How extensively: code and commit message AI-written under human direction; human reviewed the diff, ran all
  checks, and manually tested the feature

### DIFF
--- a/src-tauri/src/actions.rs
+++ b/src-tauri/src/actions.rs
@@ -915,6 +915,13 @@ pub static ACTION_MAP: Lazy<HashMap<String, Arc<dyn ShortcutAction>>> = Lazy::ne
             post_process: false,
         }) as Arc<dyn ShortcutAction>,
     );
+    // Same action as "transcribe"; the shortcut is forced to toggle mode in shortcut::handler.
+    map.insert(
+        "transcribe_toggle".to_string(),
+        Arc::new(TranscribeAction {
+            post_process: false,
+        }) as Arc<dyn ShortcutAction>,
+    );
     map.insert(
         "transcribe_with_post_process".to_string(),
         Arc::new(TranscribeAction { post_process: true }) as Arc<dyn ShortcutAction>,
--- a/src-tauri/src/lib.rs
+++ b/src-tauri/src/lib.rs
@@ -605,6 +605,7 @@ pub fn run(cli_args: CliArgs) {
             shortcut::change_binding,
             shortcut::reset_binding,
             shortcut::change_ptt_setting,
+            shortcut::change_transcribe_toggle_enabled_setting,
             shortcut::change_audio_feedback_setting,
             shortcut::change_audio_feedback_volume_setting,
             shortcut::change_sound_theme_setting,
--- a/src-tauri/src/secure_input.rs
+++ b/src-tauri/src/secure_input.rs
@@ -515,6 +515,9 @@ mod imp {
                 if id == "transcribe_with_post_process" && !settings.post_process_enabled {
                     continue;
                 }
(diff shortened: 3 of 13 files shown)


adamkdean:
Cross-posting for visibility: https://github.com/cjpais/Handy/pull/1878#issuecomment-5230812965

My preference would be that we have one hot key, and it's 200-300ms key up/down for toggle, or longer for PTT. This would be the best UX.

cjpais:
Yeah, I don't think that this is the right way forward. I do appreciate you submitting the PR, but adding more settings is honestly just confusing. When I see the screenshot, I don't really understand what this is, and we're generally not supporting new shortcuts right now, and this has been discussed at length

PR #1878, opened 2026-08-08
feat: hold-or-double-tap mode for the push-to-talk shortcut

ljden:

## Human Written Description

Optional extension for Push To Talk where a double tap of the Transcribe Shortcut functions as a toggle. Useful for sometimes recording a short transcript and sometimes for recording a longer session. This is based on input from other users such as on #147.
This is an alternative approach to the one I took in #1877 - happy to go with either (or neither if it's not the time). I think I like this one better, but the other one is ok.

## Related Issues/Discussions

Related to #147. This implements the behavior described there — a brief press
toggles, press-and-hold is push-to-talk — as an opt-in setting, refined to
require a **double**-tap to start a session so that an accidental key press
is a no-op instead of an open microphone.

With the "Double-Tap To Toggle" switch enabled (shown only while Push To Talk
is on, off by default):

- **Hold** the transcribe shortcut → push-to-talk, exactly as today
- **Double-tap** → hands-free session, ended by the next press
- **Lone tap** → the recording is discarded via the existing cancel path —
  nothing is transcribed or pasted, even if the tap caught real speech

Recording starts on key-down in all cases; classification happens at release
from the press duration, so no gesture adds latency. Every release still goes
through the existing `RELEASE_GRACE` deferral, so X11 auto-repeat can't
misclassify a held key as a tap (#1539's scenario). The coordinator's input
events now carry an `InputMode` enum (toggle / push-to-talk /
hold-or-double-tap) instead of a `push_to_talk` bool; signal/CLI triggers are
unchanged (always toggle).

Design notes:

- The setting persists when Push To Talk is turned off — it's inert without
  PTT (the handler derives the mode from `push_to_talk` first) and applies
  again when PTT returns.
- Timing constants live next to the existing `DEBOUNCE`/`RELEASE_GRACE`:
  `HOLD_THRESHOLD` 300 ms (hold vs tap) and `SECOND_TAP_WINDOW` 150 ms (lone
  vs double tap, measured from the tap's release). Deliberately constants,
  not settings — happy to expose them if there's demand. One consequence:
  a push-to-talk recording shorter than 300 ms is treated as a tap and
  discarded, which is the accidental-press protection working as intended.
- A sibling branch implements the two-shortcut variant from the #147 comments
  (separate always-toggle binding). The two are independent and composable —
  happy to go with either or both, whichever fits the project's direction.
see https://github.com/cjpais/Handy/pull/1877

## Community Feedback

In #147, @cjpais described this behavior as "a good default behavior" and
noted the code at the time wouldn't support it well — the coordinator added
since then (serialized input events with deadline-based deferral) turned out
to support it cleanly. @retspen, @codepunkt and @samleibowitz expressed
interest in a toggle mode coexisting with push-to-talk.

## Testing

- `cargo test`: 183 passed — includes 6 new deterministic tests for the
  hold-or-double-tap state machine (hold, lone-tap discard, double-tap lock,
  locked-session behavior, X11 auto-repeat absorption, hold duration measured
  from the initial press), driven by a synthetic-clock simulator mirroring the
  coordinator loop, in the same style as the existing PTT regression tests
- `cargo check`, `cargo clippy`, `cargo fmt --check` clean (no new warnings)
- `bun run lint`, `tsc`, `prettier --check`, `vite build` all pass
- Frozen v0.9 settings-store test still passes; the new setting is
  `#[serde(default)]`, no migration needed
- Manual on macOS (`bun run tauri dev`): hold/double-tap/lone-tap all behave
  as described; PTT off→on cycle keeps the switch state consistent

image
image
image

## AI Assistance

**If AI was used:**

- Tools used: Claude Code
- How extensively: implementation, tests, and commit message AI-written under
  human direction; design decisions (double-tap vs single-tap, opt-in scoping,
  timing values) made by the human, who reviewed the diff, ran all checks, and
  manually tested the feature

### DIFF
--- a/src-tauri/src/lib.rs
+++ b/src-tauri/src/lib.rs
@@ -605,6 +605,7 @@ pub fn run(cli_args: CliArgs) {
             shortcut::change_binding,
             shortcut::reset_binding,
             shortcut::change_ptt_setting,
+            shortcut::change_double_tap_toggle_setting,
             shortcut::change_audio_feedback_setting,
             shortcut::change_audio_feedback_volume_setting,
             shortcut::change_sound_theme_setting,
--- a/src-tauri/src/settings.rs
+++ b/src-tauri/src/settings.rs
@@ -349,6 +349,10 @@ pub struct AppSettings {
     pub bindings: HashMap<String, ShortcutBinding>,
     #[serde(default = "default_push_to_talk")]
     pub push_to_talk: bool,
+    /// Enables hold-or-double-tap behavior for the push-to-talk shortcut:
+    /// double-tap starts an ongoing session, a lone tap is discarded.
+    #[serde(default)]
+    pub double_tap_toggle: bool,
     #[serde(default)]
     pub audio_feedback: bool,
     #[serde(default = "default_audio_feedback_volume")]
@@ -849,6 +853,7 @@ pub fn get_default_settings() -> AppSettings {
         settings_schema_version: default_settings_schema_version(),
         bindings,
         push_to_talk: default_push_to_talk(),
+        double_tap_toggle: false,
         audio_feedback: false,
         audio_feedback_volume: default_audio_feedback_volume(),
         sound_theme: default_sound_theme(),
--- a/src-tauri/src/shortcut/handler.rs
+++ b/src-tauri/src/shortcut/handler.rs
@@ -10,7 +10,7 @@ use tauri::{AppHandle, Manager};
 use crate::actions::ACTION_MAP;
 use crate::managers::audio::AudioRecordingManager;
 use crate::settings::get_settings;
-use crate::transcription_coordinator::is_transcribe_binding;
+use crate::transcription_coordinator::{is_transcribe_binding, InputMode};
 use crate::TranscriptionCoordinator;
 
 /// Handle a shortcut event from either implementation.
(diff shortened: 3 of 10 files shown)


adamkdean:
My preference would be a short delay, to see if it's a key down/up or a key down to decide if it's toggle or PTT

ljden:
> My preference would be a short delay, to see if it's a key down/up or a key down to decide if it's toggle or PTT

@adamkdean I'm curious to hear why you prefer a short tap over a double tap?

My motivation was that I wanted to prevent accidentally triggering the toggle state. Particularly when you map a single modifier key like the right command/super/alt/option key, accidental triggers are likely.

adamkdean:
My preference has been through experience using this exact mechanism with
Wispr Flow and others over the last year or so.

It works so well that double tapping isn’t necessary imo. But: why not
allow that option?

cjpais:
I'm going to close this one as well. Again, I think maybe you misinterpreted what the original issue is describing. As far as I can tell, what the original issue is describing is a combined mode, which is probably the best default, which is you hold it and you push a talk, but if you don't hold it, you still get the behavior of like tap once, talk, tap again to stop, and probably push to talk would activate after 300 or 500 milliseconds of holding the key down just so you could disambiguate a short tap versus a long one. And this one just needs human testing. But largely, I do want to pull in something like this. It's just we don't want to add new settings, especially a menu of settings. It gets very confusing. Like this needs to be a drop-down menu. If anything, it cannot be toggles. A bunch of toggles is going to cause application logic to be a mess. It's going to cause user confusion. It's basically unsuitable for the application. So we potentially can do a drop-down menu and it would be much more clear than even what we have now. And I think that's the way we should move towards. But I need to close this just because it doesn't really represent the direction that I'm describing. So if you want to submit a new PR that moves to a drop-down menu, has this combined mode, as well as the legacy modes, that's totally fine. And the other mode that we should support is double tab. I don't really understand why people want this mode personally, but the fact is that people want it, so we probably should implement it

ljden:
@cjpais fair enough! I did this as a separate toggle because my goal was to not break backwards compatibility. I like the suggestion of a drop down though, that's a better approach. Not sure I'll get that up as a PR, but supportive of the idea.
Thanks for consideration and detailed thoughts

yachal-upson:
@ljden thanks for the exploration. Thanks @cjpais for your time.
+1 to @cjpais choice and response. I for one are looking for the UX CJ described.

PR #1892, opened 2026-08-11
fix(linux): clear stale modifier latch after xdotool type (#1817)

### LINKED ISSUE #1817: [BUG] Ctrl key can become virtually "stuck" if you keep the Ctrl key pressed while the typing tool streams transcription
### Before you submit

Searched existing issues and PRs. Closest matches, neither of which covers this:

- #707 "Modifier keys get stuck, requires logout to fix" (closed) — same symptom class, but macOS and a different code path.
- #1487 "fix(paste): release the modifier keys after pasting with wtype" (open PR) — the analogous fix for the **wtype** backend on Wayland. This report is about the **xdotool** backend on X11, which still re-presses modifiers.

### Bug Description

I have "push to talk" on and my transcribe shortcut set to the default Ctrl + Space.

If I press Ctrl + space, dictate, then release only the space bar (not the Ctrl key), the typing tool will begin streaming the transcription. If I keep the Ctrl key pressed while the typing tool works, and release only once it is done emitting, the Ctrl key becomes "virtually stuck". That is, the system acts as though the Ctrl key were still pressed.

Once it is in this state, everything I type is interpreted as a Ctrl chord until I press and release Ctrl once more, which clears it. It then happens again on the next dictation.

**Everything below the "Suspected cause" heading came out of an AI-assisted debugging session, not from me.** I have not verified any of it and I am not in a position to judge whether it is correct — please check it yourself rather than taking it at face value. The only thing I can personally vouch for is the behaviour described above and in the reproduction steps.

#### Steps to reproduce

1. Linux/X11. Push-to-talk enabled, transcribe shortcut set to the default `Ctrl + Space`.
2. Press and hold `Ctrl + Space`.
3. Dictate a sentence.
4. Release **only** the spacebar. Keep Ctrl held down.
5. Wait for the transcription to be typed into the focused window.
6. Release Ctrl **after** the text has finished being emitted.
7. Type anything. Keystrokes are now interpreted as `Ctrl` chords.
8. Press and release Ctrl once — the state clears. Repeat from step 2 and it happens again.

#### What I ruled out

- **A hardware fault or sticking keycap.** A single press-and-release of Ctrl clears it instantly, and it only ever happens immediately after a dictation — never otherwise.
- **Anything specific to one application.** The stuck state is system-wide, not confined to the window that received the text.
- **The transcription itself.** The model output is correct; the problem is purely in how the text is delivered.

#### Suspected cause

*(Reminder: not verified by me — see note above.)*

`type_text_via_xdotool` in `src-tauri/src/clipboard.rs` invokes:

```rust
Command::new("xdotool")
    .arg("type")
    .arg("--clearmodifiers")
    .arg("--")
    .arg(text)
```

`--clearmodifiers` snapshots the currently-held modifiers, releases them, types, and then **re-presses whatever it saw in the snapshot**. That restore step is the suspected problem: it assumes the snapshot is still valid, which holds for xdotool's normal synchronous use where the gap is microseconds, but this call happens after an async transcription lasting seconds. If the user has released Ctrl in the meantime, the synthetic `Ctrl-down` has no matching physical release and stays latched.

Supporting observation from the same session, which I also cannot verify:

```
$ xinput query-state 5      # "Virtual core XTEST keyboard"
key[37]=down                # keycode 37 = Control_L

$ xinput query-state 8      # the physical keyboard
(nothing held)
```

i.e. the latched Ctrl appeared on the synthetic XTEST device rather than the real keyboard, which is consistent with the restore step being the source rather than a hardware issue.

It was also suggested that this is self-sustaining: once Ctrl is latched, X reports it as logically held, so the next dictation snapshots it again and re-restores it — which would match the fact that it recurs every time until manually cleared.

Selecting `ydotool` as the typing tool avoids it (that path passes no `--clearmodifiers`), but ydotool is noticeably slower and, without modifier clearing, a held Ctrl turns the emitted text into shortcuts — in a browser it opened and closed tabs. So it is not a workaround I can use.

### System information

- Handy 0.9.4 (installed via the `handy-bin` AUR package, which repackages the official `.deb`)
- EndeavourOS, kernel 7.1.4-arch1-1
- X11, i3 4.25.1
- xdotool 4.20260303.1-1
- Settings: push-to-talk enabled, transcribe binding `ctrl+space`, typing tool `auto`, paste method `direct`, keyboard implementation `tauri`

### Logs

Relevant excerpt showing the path taken (timestamps are UTC):

```
[2026-07-31][03:36:37][handy_app_lib::clipboard][INFO] Using paste method: Direct, delay before: 60ms, delay after: 60ms
[2026-07-31][03:36:37][handy_app_lib::clipboard][INFO] Using xdotool for direct text input
[2026-07-31][03:36:40][handy_app_lib::actions][DEBUG] Text pasted successfully in 2.698175776s
```

Note that the paste is reported as successful — the stuck modifier leaves no error in the log.

xronocode:
## Before Submitting This PR

Not a duplicate: #1487 fixed the same class of bug (modifiers left latched) on the **wtype** (Wayland paste) path; this is the **xdotool** (X11 direct-typing) path, which #1487 didn't touch. One fix per PR — a single addition in one function (`type_text_via_xdotool`). Bug fix, not a feature → fits the feature-freeze "bug fixes are top priority" guidance.

## Human Written Description

@arturo-jc reported in #1817 that on X11 the Ctrl key can get "virtually stuck" after a push-to-talk dictation with Ctrl+Space — it acts held-down everywhere until you press and release it again. They traced it to `xdotool type --clearmodifiers`: it snapshots the held modifiers, releases them for the duration of typing, then re-presses the snapshot afterward. Since this paste runs after a multi-second transcription, the user has usually let go by then, so the synthetic re-press latches the key with nothing to release it. This change releases the common modifiers right after typing so a stale snapshot can't leave the key stuck.

I wrote this fix purely from reading the code and @arturo-jc's analysis — I'm on macOS and couldn't reproduce or test the X11 behavior myself. If someone on X11 can confirm it helps (or that it doesn't), that'd be welcome; I understand it's not formally mergeable without that verification, I'm leaving it here in case it's useful.

## Related Issues/Discussions

Fixes #1817

## Community Feedback

Single-maintainer project, one-file X11 typing-path fix. The reporter @arturo-jc did the root-cause legwork (xinput trace, narrowed it to `--clearmodifiers`) in #1817 — much appreciated, that's what made this a quick one.

## Testing

**Root cause (matches @arturo-jc's analysis in #1817):** `xdotool type --clearmodifiers` does snapshot → release → type → **restore (re-press)**. The restore re-presses whatever was held at snapshot time. This paste runs after a multi-second async transcription, so the snapshot is usually stale — the user released the push-to-talk modifier meanwhile. The synthetic key-down lands on the XTEST keyboard with no matching release, so the modifier reads as held system-wide (confirmed by the reporter via `xinput query-state`: Control_L down on the XTEST device, nothing held on the physical keyboard). Self-sustaining — the next dictation re-snapshots the latched key.

**What changed:** one addition in `type_text_via_xdotool` (`src-tauri/src/clipboard.rs`). After a successful `xdotool type`, explicitly release the push-style modifiers with `xdotool keyup ctrl shift alt super`. `xdotool keyup` injects through the XTEST extension (the same device the restore wrote to), so it clears the latch. Typing itself is unchanged.

**Intentional trade-offs (in the code comment):**
- Only the four push-modifiers (ctrl/shift/alt/super) are released. `--clearmodifiers` also tracks Meta / ISO_Level3_Shift (AltGr) / mouse buttons / lock keys; lock keys are skipped on purpose (they toggle on key events), the rest are out of scope for this Ctrl-latch bug.
- The release is unconditional, so it can also drop a modifier the user is physically holding across the paste. That's rare (the usual workflow is a push-to-talk binding, not a modifier held for subsequent typing); if it bites someone, the right move is to query actual modifier state (e.g. `xinput`) and release only latched keys.

**Verification I could do:** `cargo fmt --check` is clean. The function is `#[cfg(target_os = "linux")]`, so `cargo clippy` on macOS doesn't exercise it.

**Verification I need help with:** I'm on macOS and can't test X11. @arturo-jc — if you get a moment, a build of this branch + your original repro (hold Ctrl+Space, dictate, release Space, keep Ctrl held through the transcription) would be ace. The fix is tiny and this is volunteer-time, so any confirmation that Ctrl no longer latches — or a `xinput query-state` showing Control_L up afterward — would be hugely appreciated. Genuinely no rush.

## Screenshots/Videos

N/A — it's a key-state bug, not visual.

## AI Assistance

This PR was drafted with AI assistance (code + this description). The fix and the `--clearmodifiers`/XTEST reasoning were checked against the official xdotool man page, and the patch was reviewed by two independent local code models plus a manual pass over the call chain. The X11 runtime behavior was not machine-tested by me — see the verification request above.

### DIFF
--- a/src-tauri/src/clipboard.rs
+++ b/src-tauri/src/clipboard.rs
@@ -426,6 +426,37 @@ fn type_text_via_xdotool(text: &str) -> Result<(), String> {
         return Err(format!("xdotool failed: {}", stderr));
     }
 
+    // `--clearmodifiers` snapshots any held modifiers, releases them for the
+    // duration of typing, then re-presses the snapshot afterwards. This paste
+    // runs after a multi-second async transcription, so the snapshot can be
+    // stale: the user may have released the modifier (e.g. the Ctrl of a
+    // Ctrl+Space push-to-talk binding) in the meantime. The synthetic re-press
+    // then lands on the XTEST keyboard with no matching release, latching the
+    // modifier system-wide until it is pressed and released again (#1817).
+    //
+    // Release the push-style modifiers explicitly so a stale restore can never
+    // leave anything latched. `xdotool keyup` injects via the XTEST extension
+    // (the same device the restore wrote to), so this clears the latch.
+    //
+    // Two intentional trade-offs:
+    //  - Only ctrl/shift/alt/super are released. `--clearmodifiers` also tracks
+    //    Meta, ISO_Level3_Shift (AltGr), mouse buttons and caps/num/scroll lock;
+    //    the lock keys are skipped on purpose (they toggle on key events) and
+    //    the rest are out of scope for the reported Ctrl-latch bug.
+    //  - The release is unconditional. Because X modifier state is not
+    //    reference-counted per device, this can also drop a modifier the user is
+    //    physically holding across the paste. That case is rare (the common
+    //    workflow is a push-to-talk binding, not a modifier held for subsequent
+    //    typing); if it regresses someone we should switch to querying actual
+    //    state (e.g. xinput) and releasing only latched keys.
+    // Releasing a modifier that is not held is a no-op on X11, so this is safe
+    // when nothing was held. The result is ignored on purpose: the text was
+    // already typed successfully, and a failed cleanup must not fail the paste.
+    let _ = Command::new("xdotool")
+        .arg("keyup")
+        .args(["ctrl", "shift", "alt", "super"])
+        .output();
+
     Ok(())
 }
 
--- a/src-tauri/src/clipboard.rs
+++ b/src-tauri/src/clipboard.rs
@@ -421,41 +421,51 @@ fn type_text_via_xdotool(text: &str) -> Result<(), String> {
         .output()
         .map_err(|e| format!("Failed to execute xdotool: {}", e))?;
 
+    // `--clearmodifiers` restores the modifiers that were held when xdotool
+    // started. If the user releases one while xdotool is typing, that synthetic
+    // restore can leave the modifier latched on the XTEST keyboard (#1817).
+    // Release both sides of Handy's supported push-style modifiers to clear any
+    // stale restore. Lock keys are intentionally excluded because key events
+    // toggle them.
+    //
+    // The release is unconditional, so it can make a modifier that is still
+    // physically held appear released until the next physical event. This is
+    // preferable to leaving a synthetic modifier latched system-wide.
+    //
+    // Clean up before checking the typing status because xdotool may have
+    // changed modifier state before returning an error. Cleanup remains
+    // best-effort because the text may already have been partially or fully
+    // typed, but failures are logged so they can be diagnosed.
+    match Command::new("xdotool")
+        .arg("keyup")
+        .args([
+            "Control_L",
+            "Control_R",
+            "Shift_L",
+            "Shift_R",
+            "Alt_L",
+            "Alt_R",
+            "Super_L",
+            "Super_R",
+        ])
+        .output()
+    {
+        Ok(cleanup_output) if !cleanup_output.status.success() => {
+            let stderr = String::from_utf8_lossy(&cleanup_output.stderr);
+            log::warn!(
+                "xdotool modifier cleanup failed with status {:?}: {}",
+                cleanup_output.status.code(),
+                stderr.trim()
+            );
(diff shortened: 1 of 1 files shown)


arturo-jc:
Thanks @xronocode

I tested out the build.

Works great.

Issue #1580, opened 2026-07-02
[BUG] Visual problem. Not critical


sakhdevel:
## Bug Description

Not critical. Visual problem
<img width="313" height="73" alt="Image" src="https://github.com/user-attachments/assets/a1cd9a2c-43a7-4343-bef6-1477972a68b6" />

## System Information

**App Version:** 

0.9.0

**Operating System:**

ubuntu 24.04.4

other info will provided if needed

koloved:
x11?

sakhdevel:
# System Details Report
---

## Report details
- **Date generated:**                              2026-07-02 19:55:20

## Hardware Information:
- **Hardware Model:**                              XIAOMI Redmi Book Pro 16 2024
- **Memory:**                                      32,0 GiB
- **Processor:**                                   Intel® Core™ Ultra 5 125H × 18
- **Graphics:**                                    Intel® Arc™ Graphics (MTL)
- **Disk Capacity:**                               1,0 TB

## Software Information:
- **Firmware Version:**                            RMAMT6B0P0A0A
- **OS Name:**                                     Ubuntu 24.04.4 LTS
- **OS Build:**                                    (null)
- **OS Type:**                                     64-bit
- **GNOME Version:**                               46
- **Windowing System:**                            X11
- **Kernel Version:**                              Linux 6.17.0-35-generic

koloved:
i build flatpak 0.9 version https://github.com/cjpais/Handy/pull/1560, and new overlay is not looking good, If you look at the current screenshot presented in the head of the issue  and at my screenshots that were taken in Wayland, they are somewhat similar. Duplication and blurred overlay.

<img width="380" height="103" alt="Image" src="https://github.com/user-attachments/assets/a0e80600-9994-452b-bec0-c9bdd7636ada" />
<img width="741" height="170" alt="Image" src="https://github.com/user-attachments/assets/b1818b41-4adb-4df1-a980-f842091c0546" />
<img width="473" height="417" alt="Image" src="https://github.com/user-attachments/assets/7a9ac5da-b17b-4cca-b2d8-dc35def0dd49" />

cjpais:
@koloved do you mind tweaking the overlay to remove the animations and see what happens?

koloved:
@cjpais  If you remove all the animations, it results in without smooth duplication. And it keep phantom windows.  Perhaps it is still an NVIDIA wayland issue. Need to test on different hardware.

[...]

PR #1909, opened 2026-08-13
fix(overlay): stale pixels and show/hide races on Wayland layer shell

xilec:
## Before Submitting This PR

- [x] I have searched existing issues and pull requests, including closed ones, to ensure this is not a duplicate

## Human Written Description

Follow-up to [#1700](https://github.com/cjpais/Handy/pull/1700#issuecomment-5206874176).

Two cooperating problems on Wayland (GTK layer shell), both around the reused, cached overlay window:

1. **Stale pixels / blank overlay** WebKitGTK repaints only damaged regions; a layer surface unmapped and remapped with identical geometry keeps stale buffers (the ghost card from the previous session) or presents nothing. So to fix the rendering we repaint the overlay before hiding it, by resizing it back to the compact layout while it is still mapped.
orca-paste-1786029200011-1f4d7e75-60f9-4e1a-b13c-42fa2698b266

2. **Show/hide races between sessions.** Rapid re-triggers interleaved async show/hide calls, so a late hide from session N could unmap session N+1's overlay. Fix: an `OverlayController` state machine (Hidden / Visible / Hiding) serializes all visibility transitions, and every hide is tagged with the owning operation id (`begin_operation()` at transcription start), so a stale hide from a finished operation is dropped instead of unmapping a newer session's overlay.

   Out of scope here: the *coordinator* can itself desync on rapid external toggles (a press dropped while the pipeline is processing, or a sub-30 ms edge lost to debounce) and then legitimately show a recording overlay with the button already released. That is fixed separately in the follow-up PR: #1910.

Also removed the fade-out sleep in the hide path: the frontend returns null when the overlay is not visible, so no fade ever played — the delay just widened the race window.

All controller transitions live in pure methods on `OverlayController`; scenario tests drive a real controller instance and cannot drift from production behavior.

## Related Issues/Discussions

- Possibly related: #1580 (stuck "Transcribing..." overlay on Ubuntu 24.04, 0.9.0)

## Community Feedback

Bug fix; no prior discussion. Reproduced deterministically on niri (Wayland) — see Testing.

## Testing

- NixOS/niri (Wayland): manual dogfooding of rapid re-triggers, cancel mid-recording, bursts of 3-4 quick toggles — no ghost pixels, no stuck/missing overlay.
- A temporary integration harness (not included) reproduced the glitch on demand before the fix; clean after.
- `cargo test`: 214 passed (full transition matrix for the controller: show/hide × phases × fresh/stale op-id, mid-chain re-trigger, bursts, reuse after cycle).
- `cargo clippy`: clean for the touched files.
- Not tested: macOS / Windows (platform-specific parts are cfg-gated to Linux; the controller is platform-neutral).

### Known minor issue (for reviewers)

With the readiness indicator from 549cbde3, a show deferred to the hide-chain end (~210 ms window on Linux layer shell) can deliver `recording-ready` before `show-overlay`, so the restarted session briefly shows the "arming" state. Cosmetic; the ordering predates this PR.

## AI Assistance

**If AI was used:**

- Tools used: Kimi Code CLI (kimi k3, zcode (GLM5.3 Flash, reviewer GLM5.3)
- How extensively: AI did the debugging, built the reproduction harness, and drafted the fix under my direction; I reviewed and iterated on the design (park-before-hide, op-id-tagged hides, state machine) and ran the manual verification.


xilec:
Note: this push also fixes the overlay ghost-card regression that appeared after merging the latest upstream changes and resolving conflicts, and includes a small refactoring for more linear and clear code.

PR #1918, opened 2026-08-14
fix: recover gracefully when model files are deleted outside Handy

### LINKED ISSUE #1893: [BUG] Models deleted outside the app stay listed as Downloaded; Delete and Rescan don't recover, app hangs on "Loading <model>…"
## Before You Submit

Searched open and closed issues for `model`, `rescan`, `delete model`, `stuck loading`. Closest is #1656 (`.gguf model file not recognized in models/ directory`), which is about a model that *is* present not being picked up — the opposite direction. Didn't find this one.

## Bug Description

If the model files are removed from disk while Handy isn't looking (disk cleanup, a `~/.cache/huggingface` purge, migrating machines), Handy doesn't notice:

- **Models → Downloaded Models** still lists them as present, with sizes, and the previously selected one still shows **Active**.
- The status bar sticks on **"Loading Parakeet …"** with a yellow indicator and never resolves or errors.
- The hotkey still works and the recording indicator appears, but **no text is ever produced** — it fails silently, with no message anywhere in the UI.
- **Delete** on the stale entry does nothing.
- **Rescan** does not remove the stale entries either.

So the app ends up in a state it can't get out of from its own UI: it believes it has a model, can't load it, won't clear it, and won't say anything is wrong.

I'd argue the silent failure is the worst part. From the user's side the app looks like it's working — hotkey responds, indicator shows — and text simply never appears. The obvious first guess is that transcription is broken, not that the model file is missing.

## Steps to Reproduce

1. Have a model downloaded and Active (mine was `parakeet-unified-en-0.6b-Q8_0.gguf`).
2. Quit Handy.
3. Delete the model files outside the app: `rm -rf ~/.cache/huggingface`
4. Start Handy and press the transcription hotkey.

**Expected:** Handy detects the missing files, moves them to *Available to Download*, and says the model is missing — ideally offering to re-download.

**Actual:** Models still listed as Downloaded/Active, status stuck on "Loading …", hotkey records but produces no text, and neither Delete nor Rescan clears the stale entries.

## Suggested Fix

Two independent things, either of which would have avoided the dead end:

1. **Verify file existence when populating the model list** (and in Rescan), rather than trusting cached state. A stale entry pointing at a missing path should move to *Available to Download*.
2. **Make model-load failure visible and non-blocking.** A load that can't find its file should surface an error in the status bar instead of sitting on "Loading …" indefinitely. Delete should also work on an entry whose files are already gone — right now it appears to no-op, so there's no way to clear the entry.

## System Information

**App Version:** 0.9.4

**Operating System:** macOS 26.5.1 (25F80)

**CPU:** Apple Silicon (MacBook Air)

**GPU:** Apple integrated

**Model:** `handy-computer/parakeet-unified-en-0.6b-gguf/parakeet-unified-en-0.6b-Q8_0.gguf` (697 MB), also had Whisper Medium (793 MB) downloaded — both showed as present after deletion.

## Logs

Can attach on request — happy to reproduce with logging on if that helps narrow it down.

---

Happy to test a fix or put up a PR if that's useful — I noticed CONTRIBUTING.md says bug fixes are the priority during the feature freeze.

DhruvN102:
Fixes #1893 

## Problem
If a model's files are removed from disk outside Handy, Handy kept checking its cached state instead of checking reality, therefore, Delete and Rescan couldn't clear the stale entry, so there was no way out of the UI short of touching app data directly.

## Changes
- **Recover gracefully when model files are deleted outside Handy** — model
  listing (and Rescan) now verifies a model's files actually exist on disk
  instead of trusting the cached `is_downloaded` state. A stale entry pointing
  at a missing path moves back to "Available to Download" instead of staying
  listed as present.
- **Emit `loading_failed` for `get_model_path` errors too** — the
  `loading_failed` event emitter is now defined *before* the model path
  lookup instead of after, so a failure in `get_model_path` (e.g. the file no
  longer exists) also emits a terminal `loading_failed` event instead of
  leaving the frontend stuck on "Loading …" with no feedback.

### DIFF
--- a/src-tauri/src/managers/model.rs
+++ b/src-tauri/src/managers/model.rs
@@ -2399,8 +2399,14 @@ impl ModelManager {
                     deleted = true;
                 }
             }
+            // Files already missing (e.g. removed outside Handy) is not a failure —
+            // deleting is idempotent, so this still needs to fall through and clear
+            // the stale "Downloaded" entry rather than erroring out and leaving it stuck.
             if !deleted {
-                return Err(anyhow::anyhow!("No model files found to delete"));
+                debug!(
+                    "ModelManager: no HF cache/model files found on disk for {}; clearing stale entry",
+                    model_id
+                );
             }
             // Alternate-quant entries are discovery-created (the catalog only
             // seeds defaults), so deleting one un-discovers it rather than
@@ -2448,8 +2454,14 @@ impl ModelManager {
             deleted_something = true;
         }
 
+        // Files already missing (e.g. removed outside Handy) is not a failure —
+        // deleting is idempotent, so this still needs to fall through and clear
+        // the stale "Downloaded" entry rather than erroring out and leaving it stuck.
         if !deleted_something {
-            return Err(anyhow::anyhow!("No model files found to delete"));
+            debug!(
+                "ModelManager: no files found on disk for {}; clearing stale entry",
+                model_id
+            );
         }
 
         // Custom models should be removed from the list entirely since they
--- a/src-tauri/src/managers/model.rs
+++ b/src-tauri/src/managers/model.rs
@@ -1367,7 +1367,7 @@ impl ModelManager {
         let downloading_ids: HashSet<String> =
             self.cancel_flags.lock().unwrap().keys().cloned().collect();
         let mut models = self.available_models.lock().unwrap();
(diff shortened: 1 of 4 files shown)


cjpais:
Thanks I'll take a look soon!

DhruvN102 (commented):
Looks good, Thanks!

PR #1931, opened 2026-08-19
fix(ui): close settings window via Ctrl+W and sidebar button on decoration-less WMs (#1914)

### LINKED ISSUE #1914: [BUG] [i3wm] Parameter window must catch ctrl+w to be closed
## Bug Description

I'm using i3. When I open the parameter window (from the menu after a left click on the icon), I can't close it because i'm supposed to use the keyboard. With i3, there is not close, minimize, reduce at the top of the windows. 
It seems that ctrl+w is not caught and there is no close button.

Workaround for now: restart the software.

Thanks !

## System Information

**App Version:**

0.9.4

**Operating System:**

archlinux with i3

**CPU:**

NA

**GPU:**

NA
## Logs
NA

xronocode:
## Before Submitting This PR

Bug fix for #1914: on tiling WMs the settings window cannot be closed at all (no decorations, no shortcut). Not a feature; fits the "bug fixes are the top priority" guidance.

**One fix per PR:** one problem, one new close path (`hide_main_window_command`, mirroring the existing `show_main_window_command`), exposed two ways (keyboard + sidebar row) because the report names both the missing ctrl+w and the missing close button. Happy to split if preferred.

## Human Written Description

@sciunto reported in #1914 that on i3 and sway the settings window cannot be closed at all: no title bar, no close button, Ctrl+W not handled. The only way out was restarting the app. I can't reproduce this (I'm on macOS), so I went at it from the code side: in Handy closing a window already means hiding it to the tray, and the fix reuses exactly that path. I put that logic into a hide_main_window helper and wired it to the close chord (Cmd+W on macOS, Ctrl+W elsewhere, exact chord only, so configured shortcuts are not affected) and a Close button at the bottom of the sidebar. If someone on i3 or sway can test this, that'd be welcome.

## Related Issues/Discussions

Fixes #1914
Discussion:

## Community Feedback

Bug reported and described in detail by the issue author (i3, Arch, v0.9.4); no other comments on the issue yet.

## Testing

Not tested on a tiling WM (macOS host). The change is derived from code analysis: the main window is created with default decorations (server-side title bar on GNOME/KDE, nothing on tiling WMs), and `CloseRequested` → hide is the established close semantics in `lib.rs`, now shared via one helper.

Local static checks: `cargo clippy` passed with one pre-existing unrelated warning; `cargo fmt --check`, `tsc --noEmit`, and `prettier --check` clean.

Manual matrix (needs a machine with the relevant setup):
- sidebar "Close" button hides the window; tray reopens it
- exact platform chord only: Ctrl+W on Windows/Linux, Cmd+W on macOS; Ctrl+Shift+W / Ctrl+Alt+W must NOT close (user-bindable)
- shortcut recording in BOTH implementations (`tauri` and `handy_keys`): binding Ctrl+W as a custom shortcut must record normally, window must not hide
- macOS keyboard diagnostic: while it runs, the configured chord (e.g. Cmd+W) must be counted by the diagnostic without hiding the window; the close chord works again after it finishes
- known pre-existing limitation (surfaced by the new triggers, not introduced here): under `--no-tray` every close path — including the title bar — hides the window instead of quitting as the CLI table documents; recovery is relaunching Handy (single-instance shows the window). Filed separately for a maintainer decision
- reporter verification on i3 requested

Can add a macOS recording of the sidebar row hiding the window to tray on request (it behaves exactly like the red traffic-light button there).

## AI Assistance

**If AI was used:**

- Tools used: ZCode (agent) for code exploration, patch implementation, and running the local static checks
- How extensively: the patch was produced with the agent under my direction and went through an external review round before this PR. The "Close" translations for the 24 locales were AI-proposed; the Bulgarian label was corrected during review.


kvnloo:
Reviewed at exact head `5cc917357c692636b011ab5df8b55d535409c9ba` (base `cjpais:main` was `afbf44cd5f828bd48ca49588c0cf25dc6d0fc451`; live `main` is `00d255492d3aa3585c762142d54273447b6752bc`). GitHub reports mergeable. This is the right bug for #1914: on i3 the settings window has no decorations and Ctrl+W was a no-op, so the only way out was restarting. Mirroring `show_main_window_command` with `hide_main_window_command`, requiring an exact platform chord (Cmd+W / Ctrl+W, no extra modifiers), and skipping that chord while `[data-keyboard-capture]` is mounted are the correct shape.

Three things that should land before merge:

1. **The sidebar Close row is not Linux-only.** #1914 is decoration-less WMs. `src/components/Sidebar.tsx` always renders the X / “Close” control, including macOS and Windows where the native title-bar already hides to tray. Under the feature freeze that is extra chrome on platforms that do not have the bug. Gate it (`platform() === "linux"`, or a decorations check) so GNOME / macOS / Windows keep a single close affordance.

2. **The new control has no keyboard focus style.** It is a real `<button>` (good — the rest of the sidebar is clickable `<div>`s), but the class list is hover-only (`hover:bg-mid-gray/20`). Tailwind preflight removes `outline`, and `src/App.css` has no `focus-visible` fallback. Keyboard users — the people this PR is for — can Tab to Close and not see it. Add a `focus-visible:` ring/outline consistent with the active-section `bg-logo-primary/80`.

3. **`CloseRequested` does not actually call `hide_main_window`.** `src-tauri/src/lib.rs` documents one shared hide-to-tray path, then the title-bar handler still inlines `window.hide()` + `hide_dock_if_tray`. The new command always hides the `"main"` webview; CloseRequested hides whichever window emitted the event. If the overlay ever gets a close, Accessory would apply while settings stay up. Point CloseRequested at `hide_main_window(window.app_handle())` when `window.label() == "main"` so the three entry points cannot drift.

Minor, non-blocking: Ctrl+W is bound on `document` even while a textarea (post-process prompt, custom words) is focused, so it will hide the window mid-edit. Fine as Chromium-tab behavior; skip when `event.target` is `input` / `textarea` / `contenteditable` if you want to be conservative — the sidebar button still covers the mouse path.

I cannot test on i3. @sciunto confirming hide-to-tray via Ctrl+W and the sidebar row on Arch/i3 would close the loop the author asked for.

xronocode:
Thanks @kvnloo — all three addressed in 4e3548a, plus a merge with latest main in f20a2ac (it touched the same area of `src-tauri/src/lib.rs`):

1. **Linux-only sidebar Close.** Wrapped in `platform() === "linux"`, so macOS and Windows keep their native close affordance.
2. **Focus style.** Added `focus-visible:bg-logo-primary/80 focus-visible:opacity-100 focus-visible:outline-none` — the same visual as the active section, so Tab shows where you are.
3. **Shared hide path.** `CloseRequested` now routes through `hide_main_window(window.app_handle())` when `window.label() == "main"`; other windows keep the plain hide. The doc comment now describes the actual single path.

On the textarea point: left the Chromium-tab behavior as is — Ctrl+W closing the window mid-edit matches what a browser tab does, and the sidebar row covers the mouse path. Happy to gate it on `input`/`textarea`/`contenteditable` if you'd rather be conservative.

microsoft-ops (approved):

PR #1949, opened 2026-08-22
feat: pick OpenRouter upstream provider for post-processing

0x00-sys:
## Before Submitting This PR

## Human Written Description

I use OpenRouter for post-processing and noticed that the same model can get routed to different providers each time, so speed and cost are not consistent. I wanted to be able to say "for Gemini, always use Google AI Studio" or just pick the fastest provider, and also keep hitting the same provider so the prompt cache actually gets reused. OpenRouter already supports all of this, Handy just didn't expose it.

### What this adds

An "Upstream Provider" dropdown under Model, only visible for OpenRouter. Options:

- Automatic: same as today
- Fastest generation / lowest latency / cheapest: `provider.sort`, with `require_parameters` so only upstreams that accept our fields are considered
- A specific upstream (including flex/priority tiers): `provider.order` with `allow_fallbacks: false`, so every request hits the same one and reuses its prompt cache

The list comes from `/models/{id}/endpoints` and shows prices and cache read price per upstream, with a warning if one is degraded or lacks structured output.

The system prompt also gets a `cache_control` breakpoint on OpenRouter so Anthropic-style upstreams cache it. Most upstreams need a 1k+ token prefix before caching kicks in, so this mainly helps with long custom prompts.

### Implementation

- `supports_provider_routing` flag on `PostProcessProvider`, same pattern as `supports_structured_output`. No `"openrouter"` checks outside the provider defaults.
- New setting `post_process_upstream_providers` (provider -> model -> slug), defaults to empty. Old settings files load as-is.
- Shared `get_json` and `resolve_provider_and_key` helpers, used by both model fetching and the new endpoint fetch.
- Tests for the routing object, cache breakpoint and endpoint parsing. clippy/lint/format clean.

## Related Issues/Discussions

None

## Community Feedback

None gathered yet. This doesn't add a provider, it exposes routing options of the OpenRouter provider that's already built in, so I'm treating it as completing that integration rather than a new feature. Happy to open a discussion if you'd prefer that first.

## Testing

Manual, macOS Apple Silicon, OpenRouter key:

- gemini flash: ran Automatic, "Fastest generation", and pinned to google-ai-studio and vertex flex. Checked in the OpenRouter activity page that the right upstream served each one.
- Switching models loads the right list, refresh works, missing key doesn't loop.
- Other providers: dropdown hidden, requests unchanged.

Screenshot 2026-08-22 at 18 29 06

## AI Assistance

**If AI was used:**

- Tools used: Claude Code
- How extensively: wrote most of the code, I directed it, reviewed and tested.

### DIFF
--- a/src-tauri/src/actions.rs
+++ b/src-tauri/src/actions.rs
@@ -190,6 +190,13 @@ async fn post_process_transcription(settings: &AppSettings, transcription: &str)
     // field the endpoint understands and retries without it if rejected.
     let disable_reasoning = matches!(provider.id.as_str(), "custom" | "openrouter");
 
+    // Pin the gateway's upstream provider for this model (OpenRouter) so the
+    // request is reliable and always hits the same prompt cache.
+    let pinned_provider = settings.upstream_provider_for(&provider.id, &model);
+    if let Some(slug) = pinned_provider {
+        debug!("Pinning upstream provider '{}'", slug);
+    }
+
     if provider.supports_structured_output {
         debug!("Using structured outputs for provider '{}'", provider.id);
 
@@ -261,6 +268,7 @@ async fn post_process_transcription(settings: &AppSettings, transcription: &str)
             Some(system_prompt),
             Some(json_schema),
             disable_reasoning,
+            pinned_provider,
         )
         .await
         {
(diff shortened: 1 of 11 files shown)


cjpais:
Please start with community support first as mentioned in the PR template

PR #1955, opened 2026-08-23
fix: mark post-processing system prompt as a prompt cache breakpoint

0x00-sys:
## Before Submitting This PR

## Human Written Description

I noticed my post-processing requests were paying for the full system prompt every time, even though the prompt never changes. Turns out Anthropic models (and some others through OpenRouter) only cache a prompt when the request marks it with `cache_control`, which Handy never sent. With a long custom prompt that adds up fast, so this wraps the system message in a text part with the cache marker for OpenRouter and Anthropic. Everything else is unchanged.

## Related Issues/Discussions

Discussion: https://github.com/cjpais/Handy/discussions/1951

## Community Feedback

cjpais asked for this PR in the discussion above. Test numbers are posted there too.

## Testing

- `cargo test llm_client` (2 new tests: OpenRouter and Anthropic get the cache breakpoint, openai/groq/custom/cerebras/zai keep the plain string)
- `cargo fmt`, `cargo clippy` (no new warnings)
- Release build installed and used normally with OpenRouter:
  - `anthropic/claude-haiku-4.5`: request 1 $0.00587 (cache write), requests 2 and 3 $0.0006 with "Provider cached 4,530 prompt tokens", time to first token 1.81s to 1.2s
  - `google/gemma-3-27b-it` on DeepInfra (no cache support): works, normal price, no error
- Sent the same request shape through the OpenRouter API to a dozen providers that do not support caching (DeepInfra, Novita, SiliconFlow, Bedrock, Vertex, DigitalOcean, OpenAI, Google AI Studio). None rejected the field, OpenRouter drops it for them.
- Not tested: Anthropic direct (no key). Their OpenAI compatible endpoint documents `cache_control` on content parts.

OpenRouter activity, Claude Haiku 4.5 with the fix:

handy-pr2-claude-cached

Gemma 3 27B on DeepInfra, same build, no cache support, no error:

handy-pr2-gemma-nocache

## AI Assistance

**If AI was used:**

- Tools used: Claude Code
- How extensively: wrote the code change and tests, ran the API tests across providers. I tested the build in the app and reviewed the diff.


cjpais:
For now, I'm gonna close this. The main reason is oftentimes these caches are quite short-lived and not necessarily a good fit for Handy, and potentially will cost people more money by doing cache rights if they're not doing subsequent reads often enough. And in the future, we're going to be trying to move to local-only post-processing. So, this is kind of like a stopgap optimization that is eventually going to be removed anyhow

PR #1971, opened 2026-08-25
auto push to talk mode

### LINKED ISSUE #147: Auto push to talk and command line options

cjpais:
### New Shortcut Behavior

Handy now has an **Auto** shortcut mode, so the same shortcut can work as both push-to-talk and a toggle. It is the default for new installs.

- **Auto** — Hold to record and release to stop, or tap once to start and tap again to stop.
- **Hold** — Records only while you hold the shortcut. This is the old push-to-talk behavior.
- **Toggle** — Tap once to start recording and tap again to stop.

Existing users will keep their current shortcut behavior after updating. To try Auto, go to **General → Shortcut Behavior**.

## Related Issues/Discussions

Fixes #147

Shortcut Behavior setting with Auto, Hold, and Toggle options

## AI Assistance

**If AI was used:**

- Tools used: pi coding agent (Sol) & Claude Code (Fable)


eminos:
So I just ditched Wispr Flow for Handy, and it works great. Amazing work and thank you.

This is the one feature I'm kind of still missing a little bit, but I don't think it solves the problem in the best possible way. So I think if you start holding the shortcut, utilizing push to talk essentially, sometimes you want to switch over to hands-free mid-talking. You kind of realize that you have a lot to say and want to release the buttons and start doing something else, and that's possible with Wispr Flow, where you can hold your shortcut and then tap another button to switch over to hands-free mid-transcription. I think that's probably the only thing I'm kind of missing from Wispr Flow right now. (Used Handy with post-processing for this message.)

PR #1991, opened 2026-08-28
fix(macos): prevent silent AirPods recordings when also playing audio from iPhone

### LINKED ISSUE #1990: [BUG] macOS: Handy captures no audio while an iPhone is playing through the same AirPods
I often use my Mac while listening to a podcast from my iPhone through my AirPods. I found that when I triggered Handy in that situation, the recording overlay behaved normally but Handy captured nothing. There was no error, no transcription, and nothing in History. I would only discover after I had finished speaking that everything I had said was gone, which was very frustrating.

It took me a while to connect this to the AirPods. The problem occurs consistently when my iPhone is playing audio through them, even though the Mac still shows them as connected. Stopping playback on the iPhone makes Handy work again immediately.

### Steps to reproduce

1. Make the same AirPods available to both an iPhone and a Mac with automatic switching enabled.
2. Play audio from the iPhone through the AirPods.
3. On the Mac, start a Handy recording using the AirPods microphone.
4. Speak for several seconds, then stop the recording.

**Expected:** The Mac takes control of the AirPods microphone and Handy creates a transcription.

**Actual:** The overlay behaves normally, but Handy captures no audio, creates no transcription, and saves nothing to History.

### Investigation

I reproduced the problem with a small CoreAudio test client. The input stream appears healthy: callbacks fire, buffers arrive in real time, and no errors are reported. However, every audio sample is exactly zero. As a control, recording from the Mac's built-in microphone under the same conditions produces normal audio, which isolates the failure to the AirPods route.

The built-in Apple apps QuickTime Player and Voice Memos do not have the problem. Starting a recording in either app stops playback on the iPhone and allows the Mac to capture microphone audio from the AirPods.

Inspecting QuickTime's audio-only recording path showed that it keeps a muted output path open alongside the microphone input. Opening a silent output stream on the same AirPods for the lifetime of Handy's recording produces the same result and fixes the problem.

I have been using a patched build with this change during normal use. Starting a Handy recording now stops playback on the iPhone and captures the transcription.

I could not find documented guidance explaining why the concurrent output stream is required, or a documented, supported API for asking macOS to take control of the AirPods microphone from another Apple device.

### Related issues

#1886 discusses the Bluetooth playback tradeoff when using an AirPods microphone.

#1899 also involves an empty transcription, but its recording contains a non-zero number of samples. In this case the sample count is exactly zero and no History entry is created.

## System Information

**App Version:** 0.9.5

**Operating System:** macOS 26.6.2 (25G83)

**CPU:** Apple M1 Max

**GPU:** Apple M1 Max integrated GPU

**Other:** MacBook Pro (MacBookPro18,4), AirPods Pro, iPhone running iOS 26.6

## Logs

This recording contained fifteen seconds of speech:

```text
[08:48:06] TranscribeAction::start called for binding: transcribe
[08:48:21] TranscribeAction::stop called for binding: transcribe
[08:48:22] Recording stopped and samples retrieved in 39ms, sample count: 0
[08:48:22] Recording produced no audio samples; skipping persistence
```

I can provide complete logs from both failing and successful captures if needed.

mattdoran:
When an iPhone is playing through AirPods, macOS can open the AirPods microphone without reporting an error but deliver only zero samples. Handy then produces no transcription or History entry.

On macOS, look for an output device whose reported name exactly matches the selected input device. If one exists, open a silent output stream alongside the microphone stream. This restores microphone samples in the tested AirPods state, while inputs without an exact-name output match are unchanged.

This behaviour is gated behind a Mac only experimental flag "AirPods mode" to allow for community feedback.   Once positive feedback is confirmed it should become on by default.

Fixes #1990

Without this change Handy feels unreliable and can lose recordings.

## Testing

macOS 26.6.2 (MacBook Pro, Apple M1 Max), AirPods Pro, iPhone on iOS 26.6.

To reproduce without the patch: play audio from an iPhone to AirPods set to switch automatically, then record in Handy. The log ends with `sample count: 0` and nothing is saved. With the patch the same recording transcribes normally, and the iPhone's playback stops, the same way it does when you hit record in QuickTime Player or Voice Memos.

Recording from the built-in microphone is unaffected. The keep-alive stream is only opened when an output device exists under the same name as the input, which is not the case for the built-in mic (`MacBook Pro Microphone` against `MacBook Pro Speakers`). A USB headset that names its input and output identically would get one, which seems right but I have not tested it.

I also used a small standalone CoreAudio client to confirm the device really was returning all-zero samples at that moment rather than just quiet audio. Happy to share it if that is useful.

`cargo test --release --lib` passes (210 tests).

Not tested on Windows or Linux; the change is behind `#[cfg(target_os = "macos")]`. Only AirPods Pro were tested, no other Bluetooth headsets.

N/A

## AI Assistance

**If AI was used:**

- Tools used: Claude Code
- How extensively: Used for the investigation (writing CoreAudio probe clients, reading logs) and for drafting the patch and this description.


cjpais:
I'm quite hesitant to pull this in. Mainly, what I'm concerned about is that it will make us like stealing the control from the AirPods too liberally and possibly prevent other devices from switching how Apple intended. I need more testing on this, and I need to think about it further as well

mattdoran:
I know.  It's so weird.  I looked around for the "right" way to do it.   

The current behaviour is broken. The APIs act like they are working but return nothing. 

In using this for a few days it feels right.  If you consciously make a decision to transcribe you expect your AirPods to handover to this and record. 

You do not expect to talk for a bit and silently lose your transcription. 

It's a tricky one.

cjpais:
Yeah, I largely agree. The current behavior is also definitely broken with AirPods in particular

I am partially wondering whether or not this is something we can shove in advanced settings or debug settings so we can get kind of a broader range of feedback on without necessarily making it the default. For the most part I've been trying to avoid Bluetooth stuff generally speaking because it really needs a lot of time and focus to get correct, I think. So, yeah, curious your thoughts and feedback

I probably also need to try it with my AirPods in particular. I've just moved away from them for so long that I'm more or less used to just using the built-in MacBook microphone

mattdoran:
Yeah, I think that's probably a reasonable compromise.   Allowing people to opt in if they need it.  But I feel it is the sort of thing that should "just work" and be enabled by default once it's been validated.    

When the issue bites you and you get annoyed because you lose your transcription, it's not obvious what the cause is.  Nor you'd need to go hunting in settings to make it work.

The setting name is gonna be fun to name! :/    "Force AirPods to switch to Mac to Transcribe" ??

I wonder if this happens with other multi-link Bluetooth devices?  I don't have any to test.

cjpais:
I agree, I've not really had a time to test this myself yet. I am wondering right now if you want to get this in, and maybe it would be easier for me to test to see if it should be the default. Is can we throw it as an experimental setting and go from there? At least it'll make it easier for me to merge because we don't need to think about it being a default if there's a way to opt out of this kind of behavior. Either a debug setting or an experimental setting is fine. And we can probably just call it like "AirPods Mode" for now lol

And largely, I'm asking for this because if we merge it, I can turn it on as the default in my version of Handy, even as I do development, just to get longer-term testing on it, and I will more easily be able to run it with my AirPods day to day for a few weeks and see how it goes so we can decide the default

mattdoran:
Ok.  Sounds good.  Low commitment experiment that makes it easy to try (and
I don’t have to run a custom build :)

Want me to update the PR to have a “AirPods mode” experimental setting?

cjpais:
Yeah, when you get a chance, it would be great. I would love to pull it in

mattdoran:
@cjpais Done.

It passes in "airpods mode" into the audio recorder.  That keeps settings out of it at the cost of the parameter.

Note, this was done with Claude's help.  I'm an experienced developer, but do not have Rust experience.  I've been a curious observer and admirer for a long time.   So I don't have much input on the quality of the work :).

No offence if you'd like a different approach taken.

mattdoran:
I just replaced the first commit toremove the claude attribution in the log.

cjpais:
It's totally fine to use AI systems, and it's really encouraging that you're an actual developer and you're also speaking as a human in the comments because that is quite rare at this point. So I have a fair high degree of confidence in the PR, but regardless, I'll take a look and merge it fairly soon. It might take me till next week

If I don't get around to it, just ping me

mattdoran:
@cjpais I updated this PR to fix merge conflicts.  Any chance you can merge it in?

PR #1736, opened 2026-07-20
feat: add OverlayMode, mode-aware overlay functions, and AX paste verification

Caffa:
## Summary

Add overlay mode system and clipboard AX paste verification. **Depends on #1735.**

### Overlay Extensions
- Add `OverlayMode` enum (`Transcribe`, `TranscribeWithPostProcess`, `Router`)
- Add `show_recording_overlay_with_mode()`, `show_transcribing_overlay_with_mode()`, `show_processing_overlay_with_mode()` — mode-aware overlay functions
- Add `update_overlay_position_with_mode()` for dynamic overlay positioning
- Existing `show_recording_overlay()` etc. remain unchanged (default to `OverlayMode::Transcribe`)
- Integrate `focus::restore_frontmost_app()` into `hide_recording_overlay()` for focus recovery
- Add `OVERLAY_SESSION` counter for stale hide-event guard (prevents race conditions)

### Clipboard Extensions
- Add macOS Accessibility API (`AXUIElement`) bindings for paste verification
- Add `paste_via_accessibility()` — directly sets focused element's `AXValue` (bypasses Cmd+V)
- Add `verify_paste()` — confirms paste landed by reading the focused element's `AXValue`
- Add `paste_with_verification()` — 5-step fallback chain: Cmd+V → verify → AX paste → verify → clipboard-only

## Architecture
- Existing overlay functions are **unchanged** (backward compatible)
- New `*_with_mode()` functions are additive — callers opt in
- AX paste is a fallback for when Cmd+V fails (common in Electron apps)
- All new code is behind the existing overlay and clipboard modules

## Testing
- `cargo check` passes ✅

PR #1996, opened 2026-08-28
fix(macos): restore original app focus before pasting (#1995)

### LINKED ISSUE #1995: [BUG]  Dictation buffer lost when clicking outside the input field mid-session
What happens:

While dictating (Ctrl+Space held / active), if I click anywhere outside the target text field (e.g. in the Handy window itself, on the taskbar, another app), the dictated buffer is discarded. I have to click back into the original text field and start again.

Expected behaviour:

Handy should remember the original target field where the dictation started. When the user releases the hotkey, the transcribed text should paste back into that original field, regardless of where focus went in the meantime.

Why this matters:

In practice, people do a lot of clicking around while talking — scrolling, checking something, pausing to think. Losing the entire buffer on an accidental click makes the tool feel very fragile, especially for people who are using it as a replacement for typing (e.g. slower typers or accessibility users).

Suggestion:

A "sticky focus" mode — either by default or toggleable — where Handy tracks the originating input field and returns the paste to it on hotkey release.

Vaporjawn:
## Before Submitting This PR

## Related Issues/Discussions

Fixes #1995

While researching, I found closed PR #1736 included a `focus.rs` module (`save_frontmost_app`/`restore_frontmost_app`) tackling the same underlying problem with the same `NSWorkspace`/`NSRunningApplication` approach, bundled inside a much larger, unrelated multi-feature PR that was closed for scope reasons rather than because that specific mechanism didn't work. This PR extracts and ships just that piece, scoped to #1995.

## Community Feedback

None gathered yet beyond the original bug report. This is filed as a bug fix (not a new feature) restoring correct paste-target behavior, so per CONTRIBUTING.md this doesn't require a Discussion, but I'm glad to open one if the maintainer would prefer it reviewed that way.

## Testing

- `cargo fmt --check`, `cargo check`, and `cargo clippy` all clean on macOS (aarch64) — no warnings from the new code.
- `cargo test --lib` — all 242 tests pass (239 pre-existing + 3 new unit tests for the pure reactivation-decision logic in `focus_restore.rs`).
- Verified the core AppKit mechanism against the real, live desktop session with a small standalone harness using the same `objc2-app-kit` calls (`NSWorkspace::sharedWorkspace().frontmostApplication()`, `NSRunningApplication::runningApplicationWithProcessIdentifier(pid).activateWithOptions(...)`): confirmed it reads the real frontmost application (not sandboxed away from AppKit) and that `activateWithOptions` returns `true` without crashing.
- I was not able to drive a full end-to-end live dictation test (start recording → click away → release hotkey → verify paste lands in the original app) in this environment, since that needs a downloaded Whisper model plus real spoken microphone input that I can't provide programmatically. The change is scoped and low-risk (see below), but a manual click-away repro on a real machine would be a valuable final check before merging.
- Behavior when nothing changes: `capture()`/`restore()` are no-ops whenever focus never left the original application (the overwhelmingly common case), so existing dictation flows are unaffected. `restore()` is also skipped entirely when `PasteMethod::None` is selected, so users who disabled auto-paste never have their focus changed by this.
- Windows/Linux are intentionally left as no-op stubs for now rather than shipping unverified platform-specific code I have no way to build or test here — see the module doc in `focus_restore.rs`.

N/A — this is a background focus-tracking fix with no UI surface.

## AI Assistance

**If AI was used:**

- Tools used: Claude Code (Anthropic)
- How extensively: Claude Code investigated the root cause, designed and implemented the fix (`src-tauri/src/focus_restore.rs` plus the three call-site wiring changes), wrote and ran the unit tests, verified the AppKit mechanism against the live session, and wrote this PR description, under my direction.

### DIFF
--- a/src-tauri/src/actions.rs
+++ b/src-tauri/src/actions.rs
@@ -470,6 +470,15 @@ impl ShortcutAction for TranscribeAction {
         let start_time = Instant::now();
         debug!("TranscribeAction::start called for binding: {}", binding_id);
 
+        // Remember whichever application is frontmost right now, so the
+        // eventual paste can return to it even if the user clicks elsewhere
+        // (Handy's own window, another app, ...) while still dictating
+        // (#1995). Dispatched to the main thread since querying the
+        // frontmost application is an AppKit call.
+        if let Err(e) = app.run_on_main_thread(crate::focus_restore::capture) {
+            debug!("focus_restore: failed to dispatch capture to main thread: {e:?}");
+        }
+
         // Load model in the background
         let tm = app.state::<Arc<TranscriptionManager>>();
         let rm = app.state::<Arc<AudioRecordingManager>>();
--- a/src-tauri/src/clipboard.rs
+++ b/src-tauri/src/clipboard.rs
@@ -1,3 +1,4 @@
+use crate::focus_restore;
 use crate::input::{self, EnigoState};
 #[cfg(target_os = "linux")]
 use crate::settings::TypingTool;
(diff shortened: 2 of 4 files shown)


cjpais:
I don't take single platform fixes and as far as I can tell this can introduce more bugs than it fixes and the issue which was originally opened in my opinion is invalid so I'm closing this

PR #2035, opened 2026-09-04
Docker environment for Linux builds

vanviegen:
Building Handy on Linux is rather challenging. This PR provides a simple shells script that uses Docker to provide a working development environment. So with only Docker being installed you can:

```sh
docker-build-env/shell tauri build
```

## Testing

"Works on my machine!" TM

But the point of this is that Docker will hopefully make it work on your machine too.

## AI Assistance

Claude Fable 5.1 did most of the typing.


cjpais:
I mostly don't want to maintain this going forward

I would much prefer just improving documentation or writing scripts to aid in dx

PR #2054, opened 2026-09-09
feat: pause media playback while recording

cldt-fr:
## Before Submitting This PR

## Human Written Description

> **Maintainers:** this section still needs the author's own words and will be filled in before review. Please do not treat the placeholder below as the human description.

Placeholder summary of the change, for context only: dictating while music or a video is playing means either talking over it or hitting the media key first. This adds an opt-in setting that pauses what is playing when a recording starts and puts it back when the recording ends.

## Related Issues/Discussions

Fixes #
Discussion:

_No existing issue or discussion — see Community Feedback below._

## Community Feedback

None gathered yet. Handy is under a feature freeze, so this is opened as a concrete proposal rather than an expectation of a merge: happy to move it to a [Discussion](https://github.com/cjpais/Handy/discussions) first and let it sit there, or to close it, if that is preferred.

## What this does

New `pause_media_while_recording` setting (off by default), under **Settings → Sound**, next to *Mute While Recording*.

When it is on, `TranscribeAction::start` asks a new `media_control::MediaController` to pause whatever is playing, and `stop` / cancel / a failed start ask it to resume. The controller queues both on a dedicated thread, so nothing is added to the keypress → microphone path; ordering is preserved, and a recording that ends before the pause has run cancels it instead of producing an audible blip.

The rule every backend follows is that playback which was **not** already running is never started: each backend reports which players it actually paused, and the resume only touches those.

| Platform | Mechanism | Reaches |
| --- | --- | --- |
| Windows | `GlobalSystemMediaTransportControlsSessionManager` (the surface the media keys drive) | browsers, Spotify, anything publishing a media session |
| Linux | MPRIS over D-Bus — `playerctl` when installed, `dbus-send` otherwise (mirrors how `managers::audio` tries wpctl → pactl → amixer) | browsers, Spotify, VLC, mpv, … |
| macOS | AppleScript against Music, Spotify, TV, VLC and QuickTime Player, gated on a CoreAudio check that anything is playing at all; plus Chromium browsers (Brave, Chrome, Chromium, Edge, Vivaldi) via the JavaScript their dictionary can run | those players, and browsers **if** the user has enabled *Allow JavaScript from Apple Events* |

### macOS, stated plainly

macOS has no public API to pause another application's media session. `MediaRemote` is private and gated, and I verified on macOS 26 that synthetic `NSEventTypeSystemDefined` media-key events (`NSEvent otherEventWithType:` + `CGEventPost`) are no longer delivered — with Accessibility granted, on all three event taps, they moved neither the system volume nor playback in QuickTime, Music, or a Chromium browser.

So the macOS backend does two things. Dedicated players are scripted through their AppleScript dictionary. Browsers — the common case, and the one users will notice — are reached the only remaining way, by running JavaScript in their tabs through the Chromium scripting dictionary.

**That browser path depends on a setting Chromium keeps off by default**: *View > Developer > Allow JavaScript from Apple Events*. This is a real trade-off for the user, since it lets any app with automation access to the browser run JavaScript in their tabs, so nothing here turns it on or nags about it. When it is off, each `execute` is refused, the per-tab `try` swallows it, and macOS degrades to dedicated players only. I would completely understand if you would rather not carry this part; it is confined to `media_control/macos.rs` and can be dropped without touching anything else.

Two things that looked obvious did not survive contact with a real page, and are worth knowing if you review that file:

- **Chromium's tab `id` is not stable.** It changed under me between the pause and the resume of the same YouTube tab (`435388112` → `435388108`) as the SPA navigated. Nothing is keyed on it: the resume is offered to every `http(s)` tab and no-ops wherever it finds no flag of ours.
- **The flag lives on `window`, not on the element.** YouTube rebuilds its `<video>`, and a `data-` attribute goes with the old one — I watched a marker vanish mid-test.

Two things keep the macOS path quiet: CoreAudio's `kAudioDevicePropertyDeviceIsRunningSomewhere` is checked first, so a silent machine costs one property read instead of any `osascript` launch (and, on first use, instead of an automation permission prompt); and only applications whose process is actually running are scripted, so an installed-but-closed player is never launched. Each dedicated player gets its own script, because AppleScript resolves terminology at compile time and one unexpected dictionary would otherwise take the whole run down. Every `osascript` run is bounded by AppleScript's `with timeout` and killed after 5s.

Quitting the app mid-recording resumes synchronously on `RunEvent::Exit`, so the user's music can't be stranded paused.

## Testing

Environment: macOS 26 (Darwin 25.6), Apple Silicon.

- `cargo test --lib` — 262 passed (includes 5 new parser tests for the Linux MPRIS reply parsing).
- `cargo fmt --check`, `cargo clippy --lib --all-targets` — clean; no new warnings from `media_control`.
- `bun run check:translations`, ESLint, Prettier, `tsc --noEmit` — clean. The new strings are added to all 24 locales.
- **macOS, dedicated player, end to end**: QuickTime Player playing a real file, probed through its own dictionary — `playing=true` → `playing=false` while paused → `playing=true` after resume. Pause ≈350ms, resume ≈140ms.
- **macOS, browser, end to end**: a YouTube video in Brave, probed in-page — `paused=false t=19.8` → `paused=true t=20.5, flag set` → `paused=false t=21.6, flag cleared`. Pause ≈240ms, resume ≈200ms across all `http(s)` tabs.
- With nothing playing, `pause_playing_media` returns in well under a millisecond without spawning a single process — the CoreAudio gate answers it.
- **macOS, scripts**: every generated script compiled with `osacompile` against the real dictionaries. This is how Podcasts got dropped from the list — it ships no scripting dictionary at all.

Note that CoreAudio reports the output device as still running for a moment after a player pauses (the app keeps the device open), which is why that check is only ever used as an entry gate and never as proof that a pause worked.
- **Windows**: `cargo check --target x86_64-pc-windows-msvc` equivalent run as `cargo check --target x86_64-pc-windows-gnu` against `windows` 0.61.3 with the added `Media_Control` feature — compiles clean. **Not yet run on real hardware.**
- **Linux**: the module compiles and its parser tests pass, but the `playerctl` / `dbus-send` paths have **not been exercised on a real desktop session**.

Help testing on Windows and Linux would be very welcome — those are the two backends with the best coverage and the least runtime verification.

`Cargo.lock` needs no change: enabling the `windows` crate's `Media_Control` feature pulls in no new packages.

Not included; the UI change is a single toggle in **Settings → Sound**.

## AI Assistance

**If AI was used:**

- Tools used: an AI coding assistant (agentic CLI).
- How extensively: it wrote the implementation, the translations and this description, and ran the verification listed above under human direction. The Human Written Description section above is intentionally left for the author to fill in.


cjpais:
Did you look at all of the other open PRs which are effectively doing the same thing before submitting this? If not, then I have to close.

PR #2066, opened 2026-09-11
feat: add a Webhook paste target

viktor-ferenczi:
## Before Submitting This PR

Discussion: https://github.com/cjpais/Handy/discussions/2067

The closest prior art is #638, which added the External Script paste method this builds on, but that solution has shortcomings which the Webhook solves.

## Human Written Description

Allows selecting "Webhook" as a Paste Method and specifying a URL to sent a request to.

- Tested OK on Linux with my game plugin, it receives the text perfectly.
- Works the same way on Windows and Linux, no need to a platform dependent script to "relay" the request.
- If the game is not running, then it falls back to regular paste, so it still works into regular text fields and editors.

## Related Issues/Discussions

Closest is #638, but it has shortcomings.

## Community Feedback

Waiting on the discussion to occur.

## Testing

Automated (6 new tests in `clipboard.rs`, running a real listener on loopback):
- request shape: `POST`, `Content-Type: application/json; charset=utf-8`, body `{source, event, text, timestamp}`, and no `Expect: 100-continue`
- `{"handled": true}` claims the transcript — nothing is pasted
- `{"handled": false}`, non-2xx, unparseable body, refused port and wedged listener all fall back to a normal paste
- a refused port returns well inside the timeout budget; a wedged listener gives up *at* the budget

Also run: `cargo build`, `cargo test --lib` (274 passed), `cargo clippy --all-targets` (no new warnings), `cargo fmt`, `bun run build`, `bun run lint`, prettier.

Manual, on Linux (X11, Linux Mint 22.3) against a local listener, using the release `.deb`: `handled:true` pastes nothing, `handled:false` pastes normally, and a closed port pastes normally with no perceptible stall.

Not yet tested on Windows or macOS. The code path is platform-independent (one `reqwest` call plus the existing clipboard paste), but the Windows case is the one that motivates the feature, so independent confirmation there would be welcome.

Settings → Advanced → Paste Method gains a **Webhook** entry with a URL field beside it, in the same shape as the existing External Script path field.

## AI Assistance

**If AI was used:**

- Tools used: Claude Code (Opus 5 and Fable 5.1)
- How extensively: Wrote effectively all of the code, tests and this PR description, working from a detailed human-written specification that fixed the wire contract, the fallback semantics, the file-level landmarks and the design questions to settle. The human contributor supplied the reference bash implementation being replaced, chose the approach, reviewed the result and did the manual testing on Linux.

## What this changes

A new `PasteMethod::Webhook` that POSTs the transcript to a configured URL before pasting:

```
POST <configured url>
Content-Type: application/json; charset=utf-8

{"source":"handy","event":"pre-paste","text":"<transcript>","timestamp":"<ISO 8601>"}
```

A 2xx response of `{"handled": true}` means the receiver consumed the text, so Handy pastes nothing. **Everything else** — `{"handled": false}`, a non-2xx status, a connection failure, a timeout, an unparseable body — falls back to a normal clipboard paste, so a receiver that is down or wedged can never silently swallow dictation.

Design points, for reviewers:

- **Threading.** The request blocks on the main thread, exactly as the external script did. It is bounded by `webhook_timeout_ms` (default 400 ms), which covers connect *and* response, so a dead endpoint cannot stall the paste for longer than that.
- **Fallback** reuses the existing `paste_via_clipboard` Ctrl+V path rather than duplicating paste logic.
- **`auto_submit` / `clipboard_handling` are skipped when `handled: true`.** Sending Return would land in whatever window has focus — the receiving app — and there was no paste to justify overwriting the clipboard. This is the one behavioural subtlety in the change; it is commented in place.
- **Available on all platforms**, unlike External Script. Windows is the main beneficiary, since it has no such escape hatch today.
- The wire contract is fixed by an existing consumer, so the field names and the `Expect: 100-continue`-free request are pinned by tests rather than left to drift.

The External Script method is untouched and behaves exactly as before.

`"Webhook"` is added verbatim to all 24 locale files rather than machine-translated — it is IT jargon that stays "Webhook" in every language, and the placeholder is a URL.


cjpais:
Sorry new features are not being added without broad community support

PR #2082, opened 2026-09-13
fix: reset settings scroll position when switching sections

### LINKED ISSUE #2081: [BUG] Settings sections inherit scroll position from previously viewed section
**Bug Details**

Clear description of the bug: The settings sections (General, History, Models, etc.) all render inside one persistent scrollable container in `App.tsx` that never unmounts when switching sections. As a result, the scroll offset carries over from whichever section was scrolled last.

**Steps to reproduce:**
1. Open Settings, go to the History tab
2. Scroll down so History content is offset from the top
3. Click on the Models tab (or any other section)

**Expected behavior:** Models (or the newly selected section) opens scrolled to the top.

**Actual behavior:** Models opens already scrolled down, inheriting the scroll offset from History, even though its content is different and may be shorter, making parts of it inaccessible or hidden above the fold.

**Root cause:** The scroll container (`<div className="flex-1 overflow-y-auto">` around line 373 of `src/App.tsx`) wraps all settings section content via `renderSettingsContent()`, but since the container itself is never unmounted/remounted between sections, its `scrollTop` is preserved across switches.

I have a fix ready (adds a ref + `useLayoutEffect` to reset scroll to top on section change) and will open a PR referencing this issue.

wesleymatosdev:
## Before Submitting This PR

## Human Written Description

I noticed that in Settings, if you scroll down while viewing History and then switch to Models (or any other section), the new section opens already scrolled down instead of at the top. This is confusing since each section has different, unrelated content, and it can hide the top of a section behind the previous scroll offset. It's a small but noticeable UX papercut that's easy and low-risk to fix.

## Related Issues/Discussions

Fixes #2081

## Root Cause

All settings sections (General/History/Models/...) render inside one persistent scrollable container in `src/App.tsx` (`<div className="flex-1 overflow-y-auto">`, around line 373). Because `renderSettingsContent()` swaps only the child content and the scroll container itself never unmounts/remounts between section switches, the container's `scrollTop` carries over from whichever section was scrolled last.

## Fix

- Added a `settingsScrollRef` ref on the scroll container
- Added a `useLayoutEffect` keyed on `currentSection` that calls `settingsScrollRef.current?.scrollTo({ top: 0 })` — this runs synchronously before paint, so there's no visible flash of the old scroll position

This is a single, self-contained, regression-testable UI fix — 3 edits total in `src/App.tsx` (ref declaration, effect, ref attachment).

## Testing

- Manual repro: opened Settings → History → scrolled down → clicked Models → confirmed Models now opens at the top (previously retained History's scroll offset)
- Repeated across all settings sections to confirm no regressions in normal scroll behavior within a section
- `bunx tsc --noEmit` passes cleanly with no new errors

N/A — behavioral fix, visually the difference is scroll position on section switch.

## AI Assistance

**If AI was used:**

- Tools used: Claude Code (Anthropic)
- How extensively: Claude Code diagnosed the shared-scroll-container root cause and implemented the ref + useLayoutEffect fix; I reviewed and verified the fix by reproducing the bug and confirming it's resolved.


cjpais:
This is a really good catch. We will definitely pull this in ASAV as soon as I

PR #2083, opened 2026-09-13
feat: session-only Translate to English hotkey

at2me:
Implements https://github.com/cjpais/Handy/discussions/2076

Main Transcribe shortcut is unchanged.
New shortcut runs Whisper translate-to-English for that recording only and does not persist the checkbox.


cjpais:
See all the other closed pull requests for why this isn't getting pulled in. Also, you didn't fill out the PR template, so come on

PR #2108, opened 2026-09-18
fix: register settings store backend listeners only once

Lucenx9:
## Before Submitting This PR

## Human Written Description

I was reading through the frontend event handling and found that every component using the settings store calls initialize() when it mounts, and each call registers another pair of backend listeners that nothing ever removes. Dozens of components mount at once, so one settings change triggers dozens of duplicate refetches for the whole session. This PR makes initialize() set a flag before its first await, so only the first caller registers the listeners.

## Related Issues/Discussions

None found — searched issues and PRs (open and closed) for duplicate settings listeners / store initialization without an exact match. Found during a code audit of the frontend event-listener lifecycle.

Fixes #
Discussion:

## Community Feedback

No community feedback gathered yet — self-discovered bug from a code audit. Happy to open an issue or discussion first if maintainers prefer.

## Testing

- `bun run lint` — clean
- `bunx tsc --noEmit` — clean
- `bunx prettier --check src/stores/settingsStore.ts` — clean
- `cargo fmt` / `cargo clippy` / `cargo test --lib` (268 passed) — run on the audit tree that includes this change; this PR touches only a TypeScript file, so they are unaffected

Behavior verification (code-path trace):

- **Before:** `useSettings()` is used by ~53 components (App, Sidebar, all settings sections, model selector, update checker…). Each calls `store.initialize()` from its first effect when `isLoading` is still `true` — and every one of those effects can run in the same React commit, before any of the IPC round-trips inside `initialize()` resolve. So on the first render after the store starts loading, `initialize()` executed N times, and each execution registered a new `model-state-changed` + `settings-changed` listener pair that is never removed. Every backend settings event then triggered N parallel `refreshSettings()` IPC calls for the lifetime of the app (doubled again under React StrictMode in dev).
- **After:** `initialize()` sets an `initialized` flag synchronously before its first `await`. Concurrent first-mount callers see the flag and return immediately; only one listener pair is registered.

Not tested: a full `bun run tauri dev` session on this machine; happy to help with additional testing if needed.

Not applicable — no visual change.

## AI Assistance

**If AI was used:**

- Tools used: opencode (powered by z-ai/glm-5.3)
- How extensively: AI performed the code audit that found the duplicate-listener registration and drafted the guard. The human contributor reviewed the change, directed the work, and is filling in the human-written description above.

## Update: init-failure hardening

Follow-up commit `e6f9ba8` on top of the guard: both `listen()` calls are now awaited, so the listener pair is either established atomically or init fails cleanly. If init does fail, the synchronous `initialized` flag is reset in `catch` and the error rethrown, so the store is never left marked as initialized while having no listeners.


cjpais:
Thanks. I'll take a look at this

PR #1115, opened 2026-03-22
fix: show friendly error when no audio input device is available

MarceloPrado:
## Summary

- Adds a pre-flight check in `start_microphone_stream()` to detect when no audio input devices exist before hitting coreaudio
- Introduces a new `no_input_device` error type (following the existing `microphone_permission_denied` pattern)
- Shows a clear, actionable toast: **"No Microphone Found"** with a description guiding the user to connect a mic
- Adds translations for all 17 supported locales

Fixes #1114

⚠️ I have not validated the i18n keys besides en-US and pt-BR - they were all AI-gen

## Changes

**Backend (Rust):**
- `recorder.rs` — Added `is_no_input_device_error()` detection helper + 3 unit tests
- `managers/audio.rs` — Pre-flight device enumeration check before `rec.open()`
- `actions.rs` — New `"no_input_device"` error type classification
- `audio/mod.rs`, `audio_toolkit/mod.rs` — Re-exports

**Frontend (React/TS):**
- `App.tsx` — New `else if` branch for the `no_input_device` error type
- All 17 `translation.json` files — Added `noInputDeviceTitle` and `noInputDevice` i18n keys

## Before / After

**Before:** `Failed to start recording: Failed to open recorder: Failed to fetch preferred config: A backend-specific error has occurred: An unknown error unknown to the coreaudio-rs API occurred`

**After:** **No Microphone Found** — No audio input device was detected. Please connect a microphone or headset and try again.

## Test plan

- [x] `cargo test -- recorder::tests` — all 7 tests pass (including 3 new ones)
- [x] `bun run lint` — passes clean
- [ ] Manual test on a machine without a mic (e.g., Mac Studio with no external mic): trigger recording → verify the friendly toast appears

Issue #1114, opened 2026-03-22
Cryptic error when no audio input device is available


MarceloPrado:
## Description

When trying to record on a machine without an audio input device (e.g., Mac Studio with no external mic connected), the app shows a cryptic, unhelpful error message:

> Failed to start recording: Failed to open recorder: Failed to fetch preferred config: A backend-specific error has occurred: An unknown error unknown to the coreaudio-rs API occurred

## Expected Behavior

The app should show a clear, actionable error message telling the user that no microphone was found and suggesting they connect one.

## Steps to Reproduce

1. Use Handy on a machine without a built-in microphone (e.g., Mac Studio) with no external mic connected
2. Trigger a recording
3. Observe the cryptic error toast

## Environment

- Handy v0.7.12
- macOS (Mac Studio, no built-in mic)

## Root Cause

The error originates from cpal/coreaudio-rs when attempting to get the preferred audio config for a device that doesn't exist. The backend doesn't perform a pre-flight check for available input devices, and the error classification only distinguishes "microphone_permission_denied" vs "unknown" — there's no "no_input_device" category.

cjpais:
I pulled your PR in

In the future if you find a bug and immediately fix it, please feel free to just open the PR without an issue

Issue #984, opened 2026-03-09
Crash (SIGABRT) when no microphone is available


noelboss:
## Summary

Handy v0.7.9 crashes immediately with `SIGABRT` (Abort trap: 6) when invoking speech recognition on a Mac with no microphone input available (e.g., Mac Mini without built-in mic, Bluetooth headset not connected).

## Steps to Reproduce

1. Use a Mac without a built-in microphone (e.g., Mac Mini)
2. Ensure no external microphone is connected (Bluetooth headset off/disconnected)
3. Launch Handy
4. Trigger speech recognition (Push To Talk or Transcribe Shortcut)
5. App crashes instantly

## Expected Behavior

Handy should gracefully handle the absence of an audio input device — e.g., show a user-facing error message like "No microphone available" instead of crashing.

## Actual Behavior

The app calls `abort()` on a background thread and crashes with `EXC_CRASH / SIGABRT`.

## Crash Details

From the macOS crash report (`.ips`):

```
Exception: EXC_CRASH / SIGABRT
Termination: Abort trap: 6
ASI: libsystem_c.dylib — abort() called
Faulting Thread: #42 (background worker, unnamed)

Stack trace:
  0: libsystem_kernel.dylib :: __pthread_kill
  1: libsystem_pthread.dylib :: pthread_kill
  2: libsystem_c.dylib :: abort
  3–9: handy (internal offsets, no public symbols)
```

The crash is 100% reproducible — it occurred 6 times across two days (Mar 6 and Mar 9) with the same stack trace every time. A fresh reinstall did not fix it; only connecting a microphone resolves the issue.

## Environment

- **Handy**: v0.7.9
- **macOS**: 26.3 (25D125)
- **Hardware**: Mac16,10 (Mac Mini, ARM-64 native)
- **Audio**: No built-in microphone, no external mic connected at time of crash

cjpais:
This has already been reported. Please don't open duplicate issues.

https://github.com/cjpais/Handy/issues/475

PR #2110, opened 2026-09-19
fix(audio): don't abort startup when the always-on microphone can't open

nkootstra:
## Before Submitting This PR

## Human Written Description

While reviewing Handy's startup paths I found that a saved "Always-on microphone" setting crashes the app on every launch when no input device is available. The open error hits an `expect` in Tauri's setup before any window or tray exists, and the only way back in is editing settings_store.json by hand. The on-demand path already recovers from this same error, so startup should too.

## Related Issues/Discussions

Fixes # (no existing issue for the always-on startup case; the maintainer said in #1114 that a direct PR is fine for a bug found and fixed at once)

Related:

- #475 (closed): crash when starting a recording with no input device, fixed for 0.7.10 in #477 by replacing `expect`/`unwrap` in the recorder worker with a logged error. #984 was closed as a duplicate.
- #1114 / #1115: added `is_no_input_device_error` and the "no input device" toast on the on-demand (shortcut) path.
- This PR closes the remaining gap: the always-on path opens the stream inside `AudioRecordingManager::new`, which runs inside Tauri `setup`, so the same open failure aborted the whole process before any window or tray existed.

Discussion: none

## Community Feedback

Bug fix, not a feature. #475 and #984 show several users hitting the no-input-device open failure on the on-demand path. The always-on startup variant was not reported separately, probably because the app never gets far enough to show anything; the symptom is just "Handy quit unexpectedly" on every launch.

## Testing

**Root cause.** `AudioRecordingManager::new` did `manager.start_microphone_stream()?` when `always_on_microphone` is persisted. The error propagated to the `.expect("Failed to initialize recording manager")` in `setup`, which runs inside tao's `did_finish_launching`, so the panic cannot unwind and the process aborts (exit code 134). The setting never resets, so every subsequent launch aborts too.

**Fix.** Log the open failure and keep going, staying in always-on mode with the stream closed. `try_start_recording` already retries the open on every recording start, and the shortcut path already shows the "no input device" toast (#1115), so the existing on-demand recovery now covers always-on mode. Nothing else changes: with the stream closed, `update_mode(OnDemand)` early-returns in `stop_microphone_stream`, stop/cancel are unreachable from Idle, and mute is gated on `is_open`. A successful open takes exactly the same path as before.

**Why there is no new `cargo test`.** The defect sits in a constructor whose only inputs are a concrete `tauri::AppHandle`, the settings store, and cpal/CoreAudio. No test in the repo builds an `AppHandle` (the tauri `test` feature is not enabled), and adding a fake-recorder trait just to assert "does not return Err" would be a larger change than the fix. The reproduction below is deterministic and re-runnable instead; `cargo test` remains the regression guard.

**Reproduction (macOS, debug build, portable mode so the real install is untouched).** A `sandbox-exec` profile denies the process access to the CoreAudio daemon, which is what "no usable input device" looks like to cpal:

```
(version 1)
(allow default)
(deny mach-lookup (global-name "com.apple.audio.audiohald"))
(deny mach-lookup (global-name "com.apple.audio.coreaudiod"))
```

Seed `Data/settings_store.json` with `{"settings":{"always_on_microphone":true}}` next to the binary (plus a `portable` marker file and a `.cargo-lock` marker so Tauri resolves `resources/` from the exe dir), run `sandbox-exec -f no-audio.sb ./handy --debug`, and check whether the process is still alive after 10 s.

Before (`main`): the process aborts within about 2 s with exit code 134 and this in the log:

```
thread 'main' panicked at src/lib.rs:203:14:
Failed to initialize recording manager: Failed to open recorder: Failed to fetch preferred config: A backend-specific error has occurred: An unknown error unknown to the coreaudio-rs API occurred
thread caused non-unwinding panic. aborting.
```

After (this branch): the process is still running after 10 s (stopped with SIGTERM) and the log shows the new warning instead of the panic:

```
[WARN] Failed to open always-on microphone stream at startup (Failed to open recorder: Failed to fetch preferred config: ...); will retry on the next recording start
```

Unsandboxed, same seeded settings, real input device (AirPods Pro): `Microphone stream initialized in 567.807792ms`, i.e. the success path is unchanged.

Also run in `src-tauri`: `cargo fmt -- --check` (clean), `cargo clippy` (no warnings in the changed file), `cargo test` (274 passed, 0 failed).

Not tested on Windows/Linux; the change is platform-independent Rust and the open failure surfaces the same way through cpal there.

N/A. The before/after log lines above are the observable behaviour; the "before" state also shows macOS's "handy quit unexpectedly" crash reporter on every launch.

## AI Assistance

**If AI was used:**

- Tools used: Claude Code
- How extensively: it traced the crash path, wrote the one-hunk fix and the sandbox-exec reproduction, ran the red/green runs and cargo checks, and drafted this description, including a first pass of the Human Written Description, which I edited before posting. The reproduction and the fix were reviewed by me.


cjpais:
I gotta ask a question about these PRs, which is like, are these completely automated and are the findings effectively being driven by the AI? Because a lot of these do seem to be legitimate fixes, but a lot of weird corner cases that I suspect most people aren't going to hit mainly because they probably don't use these features or they're hidden behind debug flags. It just seems like this is kind of an automated pass picking them off. It's not a bad thing, I'm just kind of curious to know

nkootstra:
I ran into an issue on my Mac Mini and asked the AI to investigate possible reproduction paths, which resulted in this PR.

cjpais:
Cool appreciate the context

PR #2122, opened 2026-09-21
feat(history): search past transcriptions and highlight matches

arshit09:
## Before Submitting This PR

## Human Written Description

I needed a search feature and it was missing, so I thought I'd open this PR. Processing the already transcribed text won't break the app as mentioned in #590. I've also reviewed #974, #590 and #975 (which asked for a consistent UI and for making the search full width, so I made sure it's easy for you to review this; please see the screenshots below). I also made sure every case-insensitive match of the search query, including matches inside longer words, is wrapped in a `<mark>` using the app's primary color at 40% opacity (`bg-logo-primary/40`), so each occurrence is easy to spot. I also kept the active and normal states and keyboard accessibility in mind.

## Related Issues/Discussions

- Discussion: #974
- Previous PRs: #590, #975

## Implementation

- **Search runs in the backend.** `get_history_entries` takes an optional `search` string. When it is set, `HistoryManager` walks entries newest first with the same id cursor and returns only matches, so infinite scroll keeps working while searching.
- **Case-insensitive beyond ASCII.** Matching is done in Rust (`to_lowercase` + `contains`) instead of SQLite `LIKE`, which only folds ASCII case. "пятницу" finds "ПЯТНИЦУ", and `%` and `_` in a query are matched literally. It searches the transcript the list shows, not the post-processed text, so every result has a visible match.
- **Highlighting.** Matches are wrapped in `<mark>` with `bg-logo-primary/40`, the app's primary color at 40%, which keeps the text readable in light and dark. Offsets come from a Unicode case-insensitive regex (`search.ts`) instead of lowercasing, because lowercasing can change string length and shift the highlight.
- **Typing.** Search runs 250 ms after typing stops. Responses for an outdated query are dropped, so a slow result can't replace a newer one, and the previous results stay on screen while new ones load, so the list doesn't flicker. New recordings made during a search are only added if they match. The × button or Escape clears the search.
- **Empty state and i18n.** With no matches the list says `No transcriptions match "…"`. The three new strings (`settings.history.searchPlaceholder`, `clearSearch`, `noSearchResults`) are translated in all 26 locales.

Main files: `src-tauri/src/managers/history.rs`, `src/components/settings/history/HistorySettings.tsx`, `src/components/settings/history/search.ts` (new).

No breaking changes: the new command argument is optional, and without it `get_history_entries` behaves as before.

## Community Feedback

Discussion #974 has 1 upvote and no comments yet. On #590, its author wrote that a "search option would be really useful if implemented well". I haven't gathered further feedback.

## Testing

- [x] `bunx tsc --noEmit`, ESLint, Prettier, `bun run check:translations` and `vite build` pass
- [x] `cargo fmt --check` passes; the `managers::history` tests pass (7, including 4 new: non-ASCII case-insensitive matching, `%` and `_` matched literally, post-processed text not searched, pagination over matches only)
- [x] `bun src/components/settings/history/search.test.ts` (new) passes: matching, highlight segments, regex characters in the query, blank queries
- [x] Rendered the real `HistorySettings` in Edge with a mocked backend that uses the same search rules as the Rust code: typing "invoice" leaves the 2 matching entries and highlights all 3 occurrences, including "Invoice", in light and dark themes
- [x] Manual: search a real history in the app on Windows

| Before | After |
|---|---|
| History tab on main, no search bar | History tab searching "invoice": two matching entries with each occurrence highlighted in pink |

## AI Assistance

**If AI was used:**

- Tools used: Claude Code (Opus 5)
- How extensively: At my request, Claude Code wrote the implementation, tests and translations, rendered the screenshots and opened this PR. I've reviewed the PR multiple times and refined it.


cjpais:
You need to be a human and add your human written description. I'm not accepting just a bunch of bot text. I'm sorry, you can't spam this with a bunch of PRs even if they're valid. I need you to write something as a human

arshit09:
@cjpais I was working on this PR. Just updated the description. You can read the Human Written Description and screenshots to understand it clearly. Happy to help if there's any other problem.

cjpais:
Thank you. Just please next time do not submit PRs until everything has been written. Or leave it as a draft.

arshit09:
@cjpais Noted. My bad. Thanks.

